PR #2, opened 2019-01-14
Proposal for new scoped slots syntax

yyx990803:
[Rendered](https://github.com/vuejs/rfcs/blob/scoped-slots/active-rfcs/0000-new-slot-syntax.md)


znck:
I had doubt about the shorthand syntax being invalid attribute name but HTML spec is very flexible. Any printable charcter except (`=`, `/`, `'`, `"`, `<`, `>`) is valid attribut name (https://html.spec.whatwg.org/multipage/parsing.html#attribute-name-state).

znck:
Is this valid syntax?

``` html
<foo>
  <template :(dynamicSlotName)="{ name }">
    <div>Hello {{ name }}</div>
  </template>
</foo>
```

yyx990803:
@znck I'd avoid that. If you need that level of dynamism you probably should go render functions directly.

darrenjennings:
The `()="foo"` reminds me of a closure or anonymous function and could be confusing to scan in the case where you have code something like:

```vue
<foo ()="foo">
  <bar ()="bar">
    <baz 
      ()="baz" 
      @close="() => isVisible = false">
      {{ foo }} {{ bar }} {{ baz }}
    </baz>
  </bar>
</foo>
```

I like the `&` shorthand per @rellect's suggestion. Would it be possible to do something like `<foo &foo>` as another shorthand when you don't need to destructure? Then you could do `<foo &foo>`, `<foo &="foo">`, or `<foo &="{foo}">`. Pointers are a confusing concepts in languages like C, so I could see this turning people off from the syntax if it's "just like pointers". 

However, `v-scope` in conjunction with `&` provides nice symmetry with existing directives:

* `v-bind` => `:`
* `v-on` => `@` 
* `v-scope` => `&`

darrenjennings:
@znck you are right for events. Although it is not uncommon to see function properties that are not events which evaluate immediately, but still need to pass in properties from the template through an anon function. So a more common usage case would be something like:

```vue
<foo ()="foo">
  <bar ()="bar">
    <baz 
      ()="baz"
      :onBaz="() => someMethod(baz)">
      {{ foo }} {{ bar }} {{ baz }}
    </baz>
  </bar>
</foo>
```

However, it is not lost on me that the component author could implement `onBaz` as an event instead. I did this for vue-autosuggest ([related issue](https://github.com/darrenjennings/vue-autosuggest/issues/58)) as I had come from a React background and was used to function props, which can be helpful when you need to return a value and not just emit. You could also just make `someMethod` return a function to avoid putting this in the template, but it's something I've run across so thought I'd share.

>  so it's meant to appear like function.

Ah I see, I wonder if the fact that you cannot pass in any arguments is also what is trippy. It's a function underneath but `()` is just shorthand, not a function itself and not useful in the template for passing in arguments or a dynamic slot name as you questioned earlier.

dinhquochan:
I love this idea:

> However, `v-scope` in conjunction with `&` provides nice symmetry with existing directives:
>
> `v-bind` => `:`
> `v-on` => `@`
> `v-scope` => `&`

```js
<foo &="{ foo }">
  <bar &="{ bar }">
    <baz 
      &="baz"
      :onBaz="() => someMethod(baz)"
      @input="() => otherMethod(bar)"
      >
      {{ foo }} {{ bar }} {{ baz }}
    </baz>
  </bar>
</foo>
```

yyx990803:
I think having a scoped slot and an expression that contains an arrow function is quite rare (it's definitely possible, but also definitely not common). On the other hand, this can actually be solved at the syntax highlighting level (the parens will be colored differently).

CyberAP:
Square brackets could be used to avoid confusion with Angular props and arrow functions.

```html
<foo []="foo">
  <bar []="bar">
    <template [header]>
      123
    <template>
    <template [footer]="{ baz }">
      {{ foo }} {{ bar }} {{ footer }}
    </template>
  </bar>
</foo>
```

CyberAP:
This is not HTML compliant at all but I find this way of defining slot and scope props interesting:

```html
<foo="foo">
  <bar="bar">
    <template:header>
      123
    </template>
    <template:footer="{ baz }">
      {{ foo }} {{ bar }} {{ baz }}
    </template>
  </bar>
</foo>
```

dimensi:
v-scope says nothing about what is. how it's look with named slots? 
```vue
<bar>
  <template slot="header" v-scope="{ foo }">{{ foo }}</template>
</bar>
```
in the form of `slot="name" slot-scope="{foo}" `, I constantly spend time to understand what is which. I believe that the v-scope situation will also put mental pressure on the developer.
This is why slot-props seems clearer to me than v-scope.
Why I do not like brackets `()` because they do not work as they should work. 
In the react, the arguments are bracketed, and the arrow gives the result.
```js
(in) => <Out>{in}</Out>
```
But in our case, the brackets point to the correct slot, and then pass it inside.
```js
const { out } = (slot)
<tempalte>{{ out }}</template>
```
This is similar to react render props and not at the same time. In my opinion

chrisvfritz:
### Why a shorthand might not be appropriate for scoped slots

@yyx990803 I generally like the direction this proposal is going, except for the shorthand. I agree with @c01nd01r that at first glance, `()="foo"` can look like an arrow function _returning_ `"foo"` (e.g. `()=>"foo"`. But more broadly, I think we should reserve shorthands for features that are:

1. extremely common (any generic Vue demo is likely to include both `v-bind` and `v-on`) and B), and 
2. generally very intuitive and well-understood by the community, including recent adopters.

Both `v-bind` and `v-on` meet these criteria. Since scoped slots don't meet either of them, I worry that:

- it will take longer for people to become used to any shorthand, due to less frequent exposure to scoped slots, and  
- since scoped slots are already more difficult for many people to wrap their heads around, having to translate the shorthand while learning will only make it harder.

### Other limitations of this proposal

- **This proposal still allows and encourages patterns like `<slot v-bind="user" />`** to save a few characters with `slot-props="user"` instead of `slot-props="{ user }"`. As I mentioned in the other issue, I see this as an anti-pattern. The problem arises when the component wants to expose data other than the user. Then they have two choices: make a breaking change to their API or force this new property onto the `user` object, even though it may have very little to do with the user. Neither is a great option.

- The problem remains that **`slot-scope`/`slot-props` behaves like a `v-` prefixed directive rather than a special attribute**. All other special attributes accept any string and can be bound to expressions that evaluate to any string. With the special destructuring syntax that also adds new properties to the scope, `slot-scope`/`slot-props` are really more like a directive such as `v-for`. I think this is part of what creates confusion around scoped slots.

- The more I think about it, the more I realize **it feels strange having a separate attribute for defining the slot vs the props provided by that slot**. To get the full story of where some property is coming from, you have to look for `slot-props`, then a `slot` attribute, then the component name, which is one more step than it has to be.

### An alternative idea: `v-slot`

Just as `v-for` isn't split into multiple attributes like `v-iterate="items"` and `iteration-item="item"`, I wonder if we should combine the slot API into a single, `v-` prefixed directive, e.g. `v-slot`:

```vue
<!-- Default slot, without props -->
<foo>
  hello
</foo>

<!-- Named slot, without props -->
<foo>
  <template v-slot="one">
    hello
  </template>
</foo>

<!-- Default slot, with prop -->
<foo v-slot="{ foo }">
  {{ foo }}
</foo>

<!-- Default slot, with multiple props -->
<foo v-slot="{ foo, bar }">
  {{ foo }} {{ bar }}
</foo>

<!-- Default slot, with multiple, aliased props -->
<foo v-slot="{ foo as one, bar as two }">
  {{ one }} {{ two }}
</foo>

<!-- Named slot, with prop -->
<foo>
  <template v-slot="one with { foo }">
    hello
  </template>
</foo>

<!-- Dynamically named slot, without props -->
<foo>
  <template v-slot="[slotName]">
    hello
  </template>
</foo>

<!-- Dynamically named slot, with prop -->
<foo>
  <template v-slot="[slotName] with { foo }">
    hello
  </template>
</foo>

<!-- Nested default slots, with props -->
<foo v-slot="{ foo }">
  <bar v-slot="{ bar }">
    <baz v-slot="{ baz }">
      {{ foo }} {{ bar }} {{ baz }}
    </baz>
  </bar>
</foo>

<!-- Nested named/default slots, with props -->
<foo>
  <template v-slot="one with { baz }">
    <bar v-slot="{ bar }">
      {{ bar }} {{ baz }}
    </bar>
  </template>
  <template v-slot="two with { baz }">
    <bar v-slot="{ bar }">
      {{ bar }} {{ baz }}
    </bar>
  </template>
</foo>
```

#### A few notes:

- Just as Evan proposed, `v-scope` would only be usable on root elements for default slots and `<template>` for named slots. As a sidenote, I think it would be great if we could show a warning when someone tries to access a prop shared by the default slot inside a named slot.

- As for the exact syntax to join the slot name and props, I'm not sure whether `with` is the right choice. Maybe a colon would be better, e.g. `v-slot="one: { foo }"`/`v-slot="[slotName]: { foo }"`, or something else I'm not thinking of? I chose `with` because its use here semantically matches its role in JS: to add new properties to the scope.

- I'm not sure whether `v-slot` is exactly the right name or if it should be `v-scope` or something else. I'm personally leaning toward `v-slot`, since it would also replace the `slot` attribute. Seems like an easier transition.

#### Some advantages:

- With this proposal, it's _not_ possible to pull out all slot props as a single object, eliminating that less future-proof pattern in scoped slots.

- All slot information will now be visible on a single attribute, so there will never be any need to search for it, even with multiple properties.

- Adding new properties to the scope in a directive is more consistent with the rest of our API. I know this comes at the cost of diverging from the slot API for web components, but as long as we still use the word "slot", I think we get the best of both worlds, since the `v-` also makes it clear that this is a special implementation of the slot pattern that Vue provides.

- A legitimate argument for spreading an object onto a slot with `<slot v-bind="item" />` is that it allows users to call `item` whatever they want. This use case is addressed without the limitations of that pattern with the new `as` aliasing, e.g. `v-slot="{ item as product }"`. As a bonus, the use of `as` here should be familiar to many since `as` serves the same purpose in JS.

#### Some disadvantages/risks:

- If we introduce this in 2.6 with a soft deprecation of the old API, we essentially double the size of the slots API in the meantime, so that's more for users to learn.

- This proposal would require new limitations on slot names: unless used dynamically, they cannot contain whitespace nor include the characters `{` and `}`. That means there may be some existing libraries that would require an update to be used with the new syntax. I don't think this is a huge deal, because with the soft deprecation, users would still be able to use the old API in the meantime.

yyx990803:
@chrisvfritz 

I appreciate the input but I find the `v-slot` proposal quite flawed. I find it a big mental overhead to parse:

- Multiple different micro syntax can appear in the same position
- It's no longer straight up JavaScript destructuring (current `slot-scope` syntax has one simple rule: anything you can put in a JavaScript function's argument position)
- Semantics change completely from not-so-obvious syntax change (e.g. static named slot vs. default slot with destructuring)

Akryum:
Maybe the shorthand could be phased out of the proposal and we can make another RFC to discuss it?

yyx990803:
In response to concerns raised by @chrisvfritz regarding the original proposal:

### Usage of `v-bind` on `<slot>`:

I personally don't think this is something that should be forced at the syntax level. Preventing such usage would be overly opinionated from my perspective; a note in documentation on the potential drawbacks of such usage or even a soft warning from the compiler seems more reasonable.

### `slot` and `slot-props` not being directives.

This is something I also think is not fully resolved.

The reason `slot` was originally designed as a special attribute, was because it **is** a special attribute based on the Shadow DOM spec.

When introducing `slot-scope`, it was also a special attribute because it looks inconsistent to have a directive and a non-directive working together to denote the functionality of a single slot:

``` html
<foo>
  <template slot="foo" v-slot-scope="{ bar }>
  </template>
</foo>
```

Regarding the new proposal, `slot-props` is probably the term that best corresponds to the underlying concept: it's the props being passed to this slot. This can potentially help **a lot** in future education. I think `slot-scope` was confusing mostly because of such a lack of conceptual connection. Short directives like `v-slot` or `v-scope` also has the same problem.

We can try to make it a directive, like `v-slot-props`:

``` html
<foo>
  <template slot="foo" v-slot-props="{ bar }">
  </template>
</foo>
```

But the inconsistency still exists.

### `slot` and `slot-props` being separate attributes

This is again a carry-over from existing syntax so it can be introduced on top of simple `slot="foo"` usage.

It is indeed a viable option to ditch the syntax similarity to Shadow DOM slots altogether, since they are quite different internally (especially with scoped slots). This gives us the opportunity to unify the slot name and slot props designation in the same syntax. The `v-slot` proposal is a good exploration in this area, but as I mentioned I'm afraid the amount of micro syntax is a dealbreaker.

I actually had a similar draft for unifying slot usage in a single directive:

``` html
<foo>
  <template v-slot="{ msg }">
    Default slot {{ msg }}
  </template>
  <template v-slot:one="{ msg }">
    Slot one {{ msg }}
  </template>
</foo>
```

Here the slot name is denoted via a directive argument (like event names for `v-on`).

The major drawback is the default slot usage `v-slot="{ msg }"` isn't as conceptually accurate as `slot-props`.

Also, using arugments means the syntax do not support dynamic slot names.

But the bonus is that it makes the shorthand conversion rules similar to `v-on` and `v-bind`:

- `v-bind:id="id"` => `:id="id"`
- `v-on:click="foo"` => `@click="foo"`
- `v-slot:one="{ msg }"` => `(one)="{ msg }"` or `&one="{ msg }"`

yyx990803:
For reference: shorthand syntax comparison

## `()`

``` html
<foo ()="foo">
  <bar ()="bar">
    <baz ()="baz">
      {{ foo }} {{ bar }} {{ baz }}
    </baz>
  </bar>
</foo>

<foo>
  <template (header)="{ msg }">
    Header msg: {{ msg }}
  </template>

  <template (footer)="{ msg }">
    Footer msg: {{ msg }}
  </template>
</foo>
```

## `&`

``` html
<foo &="foo">
  <bar &="bar">
    <baz &="baz">
      {{ foo }} {{ bar }} {{ baz }}
    </baz>
  </bar>
</foo>

<foo>
  <template &header="{ msg }">
    Header msg: {{ msg }}
  </template>

  <template &footer="{ msg }">
    Footer msg: {{ msg }}
  </template>
</foo>
```

## `$`

``` html
<foo $="foo">
  <bar $="bar">
    <baz $="baz">
      {{ foo }} {{ bar }} {{ baz }}
    </baz>
  </bar>
</foo>

<foo>
  <template $header="{ msg }">
    Header msg: {{ msg }}
  </template>

  <template $footer="{ msg }">
    Footer msg: {{ msg }}
  </template>
</foo>
```

## `*`

``` html
<foo *="foo">
  <bar *="bar">
    <baz *="baz">
      {{ foo }} {{ bar }} {{ baz }}
    </baz>
  </bar>
</foo>

<foo>
  <template *header="{ msg }">
    Header msg: {{ msg }}
  </template>

  <template *footer="{ msg }">
    Footer msg: {{ msg }}
  </template>
</foo>
```

## `#`

``` html
<foo #="foo">
  <bar #="bar">
    <baz #="baz">
      {{ foo }} {{ bar }} {{ baz }}
    </baz>
  </bar>
</foo>

<foo>
  <template #header="{ msg }">
    Header msg: {{ msg }}
  </template>

  <template #footer="{ msg }">
    Footer msg: {{ msg }}
  </template>
</foo>
```

dima74:
Why not allow shorthand on normal elements? When using shorthand there is no ambiguity:

```
<parent>
  <foo
    (header)="parentSlotProps"
    ()="defaultSlotProps"
  >
    ...
  </foo>
</parent>
```

yyx990803:
@dima74 because then `<foo ()="defaultSlotProps">` would be valid syntax for passing `<foo>` as the default slot to `<parent>`.

yyx990803:
Would love to get a sentiment on the general reception of the shorthand here...

Please vote:

- 👍 for "Ok with shorthand and can live with any symbol"
- 😕 for "Ok with shorthand but have strong opinion about which symbol to use"
- 👎 for "Don't think shorthand should be added / won't use it"

dsonet:
Like @darrenjennings 's idea for use & as shorthand syntax.

zfeher:
@yyx990803 can you add an example where both the default and name slots are used?

I guess it would look like this:
```html
<foo>
  <template ()="one">
    {{ one }}
  </template>

  <template (two)="two">
    {{ two }}
  </template>

  <template (three)="three">
    {{ three }}
  </template>
</foo>
```

While trying to get to the above example I created this version too which probably a mistake and won't work: 
```html
<foo ()="one">
  <template (two)="two">
    {{ two }}
  </template>

  <template (three)="three">
    {{ three }}
  </template>
</foo>
```

murr4y:
I agree with @dccampbell, `()` looks too much like a function and it's especially confusing when compared to React (as @dimensi pointed out) because it doesn't behave like a function that receives a prop.

I voted against a shorthand because not a lot of keystrokes are saved and makes it potentially more confusing for beginners (and scoped slots/'slots with props' are quite confusing to begin with).
Even if you have a shorthand that makes sense (I personally like `[ ]` because well, it looks like a slot!) the result may not express what's happening.

```
<template (two)="{ two }">
    {{ two }}
  </template>
```
It kind of looks like you're assigning a value to a slot but actually you receive something to use for your slot.

Conceptually it's more like
```
<template slot="two" from-props="{ two }">
  </template>
```
or like 
```
<template { two }=slotname.props>...
```

`slot-props` makes this a little clearer I think and would help education like @yyx990803 said:
> the underlying concept: it's the props being passed to this slot.

rellect:
@murr4y A shorthand isn't useful just for saving a few strokes, for me it helps to easily scan the code and understand what's going at a glance.

It seems this discussion is focused on the shorthand more than the actual proposal, but I think it's important and giving it its own symbol will encourage using scoped slots.

jkzing:
I love the `v-slot` idea, but still have some questions. 

If I understand the `v-slot` idea correctly, it contains both functionality for "named slot" and "slot props". The question is how would it be with "named slot" only (without "slot props") ?

Would that be

``` vue
<foo>
  <template v-slot:one>
    Slot one {{ msg }}
  </template>
</foo>
```

or with shortland

``` vue
<foo>
  <template (one)>
    Slot one {{ msg }}
  </template>
</foo>
```

This would be another inconsistency with current `v-bind` and `v-on`: we does't allow `v-bind` and `v-on` with empty values in template.

Or just use `slot` for "named slot" only? But this could make the usage of `v-slot` not so intuitive:  you can use `v-slot` for "named slot" **and** "slot props", but cannot use `v-slot` for "named slot" only.

andreixk:
The `slot-props` is an excellent idea, but `()` shorthand seems like an odd ball solution: there are no other shorthands with other attributes employing brackets.

smolinari:
Maybe this is a totally whacky idea, as I realize it normally means "OR", but how about double pipes?

```html
<foo ||="foo">
  <bar ||="bar">
    <baz ||="baz">
      {{ foo }} {{ bar }} {{ baz }}
    </baz>
  </bar>
</foo>

<foo>
  <template |header|="{ msg }">
    Header msg: {{ msg }}
  </template>

  <template |footer|="{ msg }">
    Footer msg: {{ msg }}
  </template>
</foo>
``` 

And even a dynamic slot.
```html
<template :|slot.name|="{ msg }">
```

If anything could look like or symbolize a slot and avoid Angular confusion, to me, it's those double pipes. :grin:  

@yyx990803 - One other suggestion. Maybe you could put up a similar vote for the shorthand symbols suggested via one comment per suggested symbol and everyone upvote their favorites with a thumbsup or downvote the ones they dislike the most with a thumbsdown?

Scott

brophdawg11:
I like the new `slot/slot-props` idea and think it's fairly clear, even if it is a tad explicit.  I agree with the argument that it's likely used far less often then `v-bind`/`v-on`, and therefore the argument that it doesn't need a shorthand makes some sense to me.  But it seems as if in general, that's not a popular opinion ad shorthand are going to win out 😕.

I can't say I _love_ any of the current shorthand proposals, but I certainly like `()` the least, due to the ambiguity with lamda functions as mentioned by @dccampbell.  I think this is even more confusing with named slots.  This example reads to me like _"you now have a 'header' variable accessible to you"_, when all you have available is `msg`.  

```
<foo>
  <template (header)="{ msg }">
    Header msg: {{ msg }}
  </template>
</foo>
```

While those of us reading the proposal and this discussion may grasp this fairly quickly, I think this would be a very common mistake/issue amongst larger teams, teams with varying levels of experience, and maybe mostly, those that are relatively new to JS, ES6 and Vue.  One of Vue's greatest strengths in my opinion is the easy learning curve as compared to competing frameworks - and this somewhat double-meaning of `()` seems to work against that.

I also agree with @dccampbell on the non-full height characters (`*`, `^`) feeling weird.

So I think I'm also in favor of `#` and `&`, I suppose in that order.

yyx990803:
Proposal has gone through a major revision and is now based on the `v-slot` directive syntax. I'm also convinced about the drawbacks of `()` as a shorthand and now leaning towards `#`.

[Latest proposal, rendered](https://github.com/vuejs/rfcs/blob/scoped-slots/active-rfcs/0000-new-scoped-slots-syntax.md)

henriqemalheiros:
> The major drawback is the default slot usage `v-slot="{ msg }"` isn't as conceptually accurate as `slot-props`.

@yyx990803 for me it looks just like [`v-bind`](https://vuejs.org/v2/api/#v-bind) on reverse:

```vue
<div v-bind="{ id: someProp, 'other-attr': otherProp }"></div>
```

Overall, I think `v-slot` is conceptually easy to understand and the `&` makes sense as a shorthand since it reminds SCSS/Sass syntax for nesting things.

laander:
Fan of following the directive syntax in line with `v-bind` and `v-on` (and their respective shorthands):

```html
<!-- v-bind -->

<a v-bind:href="url">
<a :href="url">

<!-- v-on -->

<a v-on:click="doSomething">
<a @click="doSomething">

<!-- v-slot, named slot -->

<foo v-slot:bar>
<foo #bar>

<!-- v-slot, default slot with props -->

<foo v-slot="{ baz }">
<foo #="{ baz }">

<!-- v-slot, named slot with props -->

<foo v-slot:bar="{ baz }">
<foo #bar="{ baz }">
```

This would work equally well with any shorthand symbol, be it `*`, `^`, `&`, `#`

Biggest drawback is dynamic slot names, as that quickly become awkward. This just doesn't feel right:
```html
<foo v-slot:[myBar]="{ baz }">
<foo #[myBar]="{ baz }">
```

dimensi:
@yyx990803 what about dynamic slots? Where about it in proposal?

yyx990803:
@dimensi square brackets like @laander mentioned:

``` html
<foo v-slot:[myBar]="{ baz }">
<foo #[myBar]="{ baz }">
```

dimensi:
@yyx990803 if we add this feature, then we get dynamic directives too or this feature only for slots?

henriqemalheiros:
Doesn't dynamic slots add too much complexity, which is what this RFC is trying to mitigate? I think dynamic slots are a *really* tiny use case (dynamic bindings and dynamic events too). Also, [this was addressed 2 days ago](https://github.com/vuejs/rfcs/pull/2#issuecomment-454136062).

dimensi:
@henriqemalheiros i use dynamic slot for Tabs component or dynamic slots can used for animation.

yyx990803:
@henriqemalheiros it's rarely used for sure, but it would be a pain for those who wrote a component with templates and then had to refactor into render functions just because they need dynamic slot names. Having the syntax doesn't affect those who don't use it.

dccampbell:
A `[]` syntax for dynamic directives feels like it should probably be it's own RFC.

yyx990803:
@dccampbell yeah, that's why I didn't include it in this proposal

henriqemalheiros:
@yyx990803 the problem is that dynamic slots would open space to dynamic bindings and dynamic events, steeping the template syntax learning curve. I use dynamic slots too, but they are really tiny use case. They are not even in the documentation. If someone is using dynamic slots, they know their way through Vue and refactoring into render functions will not take too much time.

active-rfcs/0000-new-scoped-slots-syntax.md
```diff
@@ -259,10 +333,90 @@ The change is fully backwards compatible, so we can roll it out in a minor relea
...
+After the initial round of feedback I agree that `()` as the shorthand can be potentially confusing to beginners, and am now leaning towards `#`. Some examples of the usage in a real-world library that relies on scoped slots ([vue-promised](https://github.com/posva/vue-promised)):
+
+``` html
+<Promised :promise="usersPromise">
+  <template #pending>
+    <p>Loading...</p>
+  </template>
+
+  <template #="users">
+    <ul slot-scope="users">
```
bricksroo: the `slot-scope="users"` on the `ul` is no longer needed, correct?

yyx990803: thanks, fixed

yyx990803:
Since we seem to have a consensus for the updated `v-slot` based syntax and most of the disagreements are regarding the shorthand, I'm dropping it from this RFC so we can merge this one and discuss the shorthand and dynamic keys in separate RFCs.

Nandiin:
I'm concerning about the inconsistency between default slot and named slot introduced by this `v-slot` based syntax.

```vue
<foo v-slot="defaultSlotProp">
  Using {{ defaultSlotProp }} here
  <template v-slot:slotName="namedSlotProp">
    <bar>Using {{ namedSlotProp }} here</bar>
  </template>
</foo>
```

As we can easily see, `v-slot` directive for the default slot exists on the component node while that for the named slot exists on the `template` wrapper (one on the parent while another on the child)

IMO, this consistency is due to `v-slot` having two different effects/roles/responsibilities:
1.  Locates the content that should be plugged into the slot position.
2. Provides a place to declare slot scope.

While the default slot usecase only covers the second one. (The locating is achieved by **grabbing all nameless-nodes** -- a convention other than `v-slot`)

So I think providing two different directives for these two purposes would be better:
1. `v-scope` on the parent component which provides a centeral place to declare slot scope. All slot scopes, including named ones, would be declared on the parent component.
2. `v-slot` acts exactly the same as the original `slot` (or we could leave it unchanged)

### Advantages:
1. Default slot and named slot usages are consistent.
2. Get rid of `<template>` verbosity for named slots
3. All scope declarations is on the provider node which improves clarity.

### Examples:

1. default slot without scope (no changes introduced)
```vue
<foo>
    <div>I'm slot content</div>
    <div>I'm slot content, too</div>
</foo>
```

2. default slot with scope
```vue
<foo v-scope:default="foo">
    <div> {{ foo }} can be used here </div>
    <div> {{ foo }} can aslo be used here </div>
</foo>
```

3. default slot and named slot with scopes

```vue
<foo v-scope:default="foo" v-scope:header="headerProp">
    <div> Referencing {{ foo }} </div>
    <div v-slot="header"> Referencing {{ headerProp }}</div>
</foo>
```

We might elminate the necessity to explicitly write `:default` after `v-scope` when it's for default slots.

znck:
This would cause confusion as visually scoped variables are available in every slot, which is indeed wrong. 

```html
<foo v-scope:default="foo" v-scope:header="headerProp">
    <div> Referencing {{ headerProp }} </div>
    <div v-slot="header"> Referencing {{ headerProp }}</div>
</foo>
```

leopiccionia:
This ship has sailed, but, in my opinion, the `slot`+`slot-props` and `slot`+`v-scope` proposals were better (I don't know if it's viable to have `scope` in 3.x again, but I'm OK with it being a directive) and closer to faux-HTML syntax. The resulting code was slightly more verbose than this RFC, but easier to parse (possible bias here).

Then:
```html
<foo>
  <template slot="one" v-scope="{someProp}">
    <bar v-scope="bar">
      <div>{{ someProp }} {{ bar }}</div>
    </bar>
  </template>
</foo>
```

Now:
```html
<foo>
  <template v-slot:one="{someProp}">
    <bar v-slot="bar">
      <div>{{ someProp }} {{ bar }}</div>
    </bar>
  </template>
</foo>
```

As a nicer sidenote, it solves the problem of dynamic slot much more elegantly.

Nandiin:
@znck Yeah, I missed that point. While compared to intrinsic confusion caused by syntax inconsistency, this one can be easily catched by the compiler (at least at non production build) or cause a meaningful warning at runtime.

znck:
In my opinion, the API is quite consistent:

For example, the default slot usage should be:

``` html
<foo>
  <template v-slot="one>
    {{ one }}
  </template>
</foo>
```

But it's little verbose so the API allows putting the default slot scope on the component tag itself.

``` html
<foo v-slot="one">
  {{ one }}
</foo>
```

But if there is one more named slot then API forces to put the default slot as a child with a template wrapper:

``` html
<foo>
  <template v-slot="one">
    {{ one }}
  </template>

  <template v-slot:two="two">
    {{ two }}
  </template>

  <template v-slot:three="three">
    {{ three }}
  </template>
</foo>
```

Nandiin:
> But if there is one more named slot then API forces to put the default slot as a child with a template wrapper

If that constraint exists, then the consistency problem is solved. I'm sorry that I haven't preceived that idea from the proposal.

However, another motivation (quoted below) of the proposal is somehow compromised
> the placement of slot-scope doesn't always clearly reflect which component is actually providing the scope variable. 

The only case where the placement of new `v-slot` reflects which component is providing the scope variable is the simplified less-verbose default-slot-only case.

I understand that *not clearly reflecting which component is actual provider* is only a visual issue and does not lead to any compiler ambiguity, but the visual issue itself do be among the original motivations.

### Example (stolen from the proposal):
```vue
<!-- old nested usage -->
<foo>
  <bar slot-scope="foo">
    <baz slot-scope="bar">
      <div slot-scope="baz">
        {{ foo }} {{ bar }} {{ baz }}
      </div>
    </baz>
  </bar>
</foo>
<!-- The proposal considers above code as bad -->
<!-- and states that the new syntax improves the clarity as below  -->
<foo v-slot="foo">
  <bar v-slot="bar">
    <baz v-slot="baz">
      {{ foo }} {{ bar }} {{ baz }}
    </baz>
  </bar>
</foo>


<!-- But when we brings some named slots and the constraint -->

<!-- OLD -->
<foo>
  <bar slot-scope="foo">
    <baz slot-scope="bar">
      <div slot-scope="baz">
        {{ foo }} {{ bar }} {{ baz }}
      </div>
      <template v-slot:named-slot-for-baz>BAZ</template>
    </baz>
    <template v-slot:named-slot-for-bar>BAR</template>
  </bar>
  <template v-slot:named-slot-for-foo>FOO</template>
</foo>

<!-- NEW -->
<foo>
  <template v-slot="foo">
    <bar>
      <template v-slot="bar">
        <baz>
          <template v-slot="baz">
            <div>
              {{ foo }} {{ bar }} {{ baz }}
            </div>
          </template>
         <template v-slot:named-slot-for-baz>BAZ</template>
       </baz>
     </template>
     <template v-slot:named-slot-for-bar>BAR</template>
    </bar>
  </template>
  <template v-slot:named-slot-for-foo>FOO</template>
</foo>

<!-- This seems bringing nothing but verbosity to me -->


<!-- v-scope + v-slot approach -->
<foo v-scope:default="foo">
  <bar v-scope:default="bar">
    <baz v-scope:default="baz">
      <div v-slot:default>
        {{ foo }} {{ bar }} {{ baz }}
      </div>
      <template v-slot="named-slot-for-baz">BAZ</template>
    <baz>
    <template v-slot="named-slot-for-bar">BAR</template>
  <bar>
  <template v-slot="named-slot-for-foo">FOO</template>
</foo>
```

As I points out in my first comment, the root of the problem is `v-slot` doing 2 different things and these 2 things contradict each other under the condition of *placement of `v-slot` should reflect which component actually provides the scope variable*:

- To make the `v-slot` placement reflects the provider, it should be on the provider
- To make the `v-slot` placement determines the node(s) that should be plugged into slot, it should be on the child (template wrapper).

yyx990803:
@Nandiin your example of nested default slots vs. nested named slots is a considered tradeoff, because the former is in practice more commonly seen than the latter. I'd also argue the new syntax in this case, while being more verbose, is easier to read and understand.

smolinari:
I've taken some code from a tutorial on using dynamic named slots I was working on and would like to rebuild it with the new proposal to see if I understand it.

The idea of the code is to build different types of survey questions using the dynamic named slots feature. It is also using Quasar Framework btw. 

```html
<!-- old -->
<!-- this is the questions component code -->
<template id="questions-template">
  <div>
    <div v-for="question in questions" :key="question.id"> 
      <slot name="text" :text="question.text"></slot>
      <slot :name="question.id" :input="question.input"></slot>
    </div>
  </div>
</template>

<div id="q-app">
  <p class="caption"><strong>{{ title }}</strong></p>
  <questions-component :questions="questions">
    <div slot="text" slot-scope="{ text }">
      {{ text }}
    </div>    
    <template v-for="question in questions"
      :key="question.id"
      :slot="question.id" 
      slot-scope="{ input }"
    >
      <q-field :helper="input.help">
        <component v-if="input.type === 'input'"
          is="q-input"
          v-model="input.value">
        </component> 
        <component v-if="input.type === 'select'"
          is="q-select"
          v-model="input.selection"
          :options="input.options">
        </component> 
        <component v-if="input.type === 'multi-select'"
          is="q-select"
          v-model="input.selection"
          multiple
          :options="input.options">
        </component> 
      </q-field><br/>
    </template>        
  </questions-component>
</div>
```

```html
<!-- New -->
<!-- this is the questions component code -->
<template id="questions-template">
  <div>
    <div v-for="question in questions" :key="question.id"> 
      <slot name="text" :text="question.text"></slot>
      <slot :name="question.id" :input="question.input"></slot><br/>
    </div>
  </div>
</template>

<div id="q-app">
  <p class="caption"><strong>{{ title }}</strong></p>
  <questions-component :questions="questions">
    <div v-slot:text="{ text }">
      {{ text }}
    </div>    
    <template 
      v-for="question in questions"
      :key="question.id"
      v-slot:[question.id]="{ input }"
    >
      <q-field :helper="input.help">
        <component v-if="input.type === 'input'"
          is="q-input"
          v-model="input.value">
        </component> 
        <component v-if="input.type === 'select'"
          is="q-select"
          v-model="input.selection"
          :options="input.options">
        </component> 
        <component v-if="input.type === 'multi-select'"
          is="q-select"
          v-model="input.selection"
          multiple
          :options="input.options">
        </component> 
      </q-field><br/>
    </template>        
  </questions-component>
</div>
```

Is this correct? 

Dynamic slot names weren't mentioned in this RFP, but would that be how they might work? ( and please do keep them! 😃)

Scott

Akryum:
@smolinari See https://github.com/vuejs/rfcs/pull/4

smolinari:
@Akryum - Thanks. I missed that RFC completely. 😊 But, it seems I got it right. 😁 

Scott

yyx990803:
@smolinari the new syntax requires `<template>` wrapper for named slots, so for your first div it has to be `<template v-slot:text="{ text }"><div>{{ text }}</div></template>`

smolinari:
Aha! Ok. Thanks for the clarification Evan. 

Scott

guilhermeaiolfi:
Am I late to the party?

Can't the ```v-slot``` be used to represent the name? like: 

Named Slots: ```<template v-slot="namedSlot">``` 
Default slot: ```<template v-slot>``` 

And use attributes to get the variables into the slot scope:
```<template v-slot :someProp="someProp">{{someProp}}</template>```

For scoped slots (just like in ```<style scoped>```):
```<template v-slot :onlyProp="someVar" scoped>{{onlyProp}}</template>```

**Comment by donnysim** [comment]
Would it be possible to output the same slot to multiple named slots? For example the select component provides customization for dropdown item and displayed selected item via different slots, and the developer wants to use the same template for both?

**Comment by smolinari** [comment]
@JosephSilber - just remember, with this change, named slots must be within  a `<template>`. 😃 

Maybe it could work this way too???

```html
<beefy-select>
    <template #[beefySelectSlots]="item">
      <div></div>
    </template>
<beefy-select>
```
and in the code

```
export default {
  data () {
    return {
      beefySelectSlots: ['selection', 'items'],
....
```

In other words, if the slot's argument is just a string, there is only one slot to fill. If it's a dynamic named slot and the argument is an array, then there are multiple slots to fill. That would make this really flexible. Adds to the complexity though. 😄 

Scott

smolinari:
@donnysim

>  I don't really like having to declare a variable just to pass multiple slots, it just feels disconnected when viewing the template.

Well, I can't disagree with you and in fact, I was going to ask if this might be possible.

`#['selection', 'result']="item"`

But, I seemed to remember reading somewhere that spaces in the argument like that wouldn't be possible. Thus, why I suggested the variable/ array. I guess we'll have to wait until Evan can chime in on this.

Scott

donnysim:
@smolinari oh yeah, true, true, I totally forgot about spaces.

pxwee5:
The new syntax breaks when used with the special is attribute. Both the HTML below are server rendered. 
```html
<table>
  <tr is="my-tr">
    <template v-slot:default>
      <td>Test</td>
    </template>

    <template v-slot:named>
      <td>Test</td>
    </template>
  </tr>
</table>
```

The old syntax works flawlessly.
```html
<table>
  <tr is="my-tr">
    <template slot="default">
      <td>Test</td>
    </template>

    <template slot="named">
      <td>Test</td>
    </template>
  </tr>
</table>
```

Error message:
```
<template v-slot> can only appear at the root level inside the receiving the component

217|  <table>
218|    <tbody><tr is="my-tr">
219|      <template v-slot:default="">
   |      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
220|        <td>Test</td>
   |  ^^^^^^^^^^^^^^^^^^^
221|      </template>
   |  ^^^^^^^^^^^^^^^
```

yyx990803:
@pxwee5 this is an implementation bug, please open an issue in the core vue repo instead of commenting on the RFC.

pxwee5:
Was clarifying if it's a bug or not. 
Issue created by MyBeta https://github.com/vuejs/vue/issues/10152
Push requested by zrh122 https://github.com/vuejs/vue/pull/10167

Raiondesu:
@smolinari, your proposal, unfortunately, would break all compatibility with [pug](https://pugjs.org) (see [this part](https://pugjs.org/language/interpolation.html#tag-interpolation)), the loader for which ([pug-loader](https://github.com/pugjs/pug-loader)) already struggles with the current shorthand syntax.

tmorehouse:
It is too bad (that I am too late in bringing this up), but there are issues (most likely previously discussed elsewhere) with putting slot names/vars in the attribute name.

A few key  issues are (that really limit the flexibility of scoped slots):

1) case sensitivity of slot names is lost when using in-document templates: browsers will always lowercase the attribute name (but will leave the content between the double quotes as-is)

2) Slot names can no longer have spaces in them (unless passed as a variable), or start with a square bracket `[`

3) for dyanmic slot names (using the square bracket syntax), one cannot always use a `camelCase` variable name (if using in browser templates... the var name will be lower cased by the browser... see point `#1`)

A better approach (in my opinion), would have been to break the new syntax in two:

```html
<!-- pass a string literal as the slot name (needs quotes, just like regular directives) -->
<template v-slot="'slotName'" v-scope="{ some, props, here }"></template>

<!-- pass a variable containing the slot name, case sensitivity ov var name preserved -->
<template v-slot="someVar" v-scope="{ some, props, here }"></template>
```

`v-scope` would be optional if users don't need scoping, or don't need to use the scope properties (e.g.. if the scope is just passing convenience methods as properties, and are not needed by the user in all cases)

This gets around all the limitations with attribute names (spaces, case sensitivity not guaranteed, etc), while keeping the attributes as custom attributes (i.e. hyphenated).

A shortcut syntax could still be employed, but may not be so short... maybe the shortcut could be as is in 2.6+ (combining both slot name and scope into a single attribute/value pair), while the above suggested longhand syntax (`v-slot + v-scope`) provides the most flexibility to get around the limitations of attribute names.

viT-1:
Where I can find full specification about new v-slot naming convention (e.g. for naming with more than one word)?

PR #3, opened 2019-01-16
Proposal for v-slot syntax shorthand

yyx990803:
``` html
<foo #default="{ msg }">
  {{ msg }}
</foo>

<foo>
  <template #header="{ msg }">
    Message from header: {{ msg }}
  </template>

   <template #footer>
    A static footer
  </template>
</foo>
```

[Full rendered proposal](https://github.com/vuejs/rfcs/blob/slot-shorthand/active-rfcs/0000-slot-syntax-shorthand.md)

### DIFF
--- a/active-rfcs/0000-slot-syntax-shorthand.md
+++ b/active-rfcs/0000-slot-syntax-shorthand.md
@@ -35,8 +35,8 @@ In Vue we currently provide shorthands only for two directives: `v-bind` and `v-
 <div v-bind:id="id"></div>
 <div :id="id"></div>
 
-<button v-on:click="onClick"></div>
-<button @click="onClick"></div>
+<button v-on:click="onClick"></button>
+<button @click="onClick"></button>
 ```
 
 `v-bind` and `v-on` often get used repeatedly on the same element, and can get verbose when the information differing between each is **the directive argument** instead of the directive itself. **Shorthands thus help improve the signal/noise ratio by shortening the parts that are duplicated (`v-xxx`) and highlighting the parts that are different (the arguments).**
--- a/active-rfcs/0000-slot-syntax-shorthand.md
+++ b/active-rfcs/0000-slot-syntax-shorthand.md
@@ -1,0 +1,176 @@
+- Start Date: 01-16-2019
+- Target Major Version: (2.x / 3.x)
+- Reference Issues: https://github.com/vuejs/rfcs/pull/2
+- Implementation PR: (leave this empty)
+
+# Summary
+
+Add shorthand syntax for the `v-slot` syntax proposed in [rfc-0001](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0001-new-slot-syntax.md). Please make sure to read it first to get proper context for this proposal.
+
+# Basic example
+
+``` html
+<foo #="{ msg }">
+  {{ msg }}
+</foo>
+
+ <foo>
+  <template #header="{ msg }">
+    Message from header: {{ msg }}
+  </template>
+
+   <template #footer>
+    A static footer
+  </template>
+</foo>
+```
+
+# Motivation
+
+A shorthand, as the name entails, primarily aims to provide more succinct syntax.
+
+In Vue we currently provide shorthands only for two directives: `v-bind` and `v-on`:
+
+``` html
+<div v-bind:id="id"></div>
+<div :id="id"></div>
+
+<button v-on:click="onClick"></div>
+<button @click="onClick"></div>
+```
+
+`v-bind` and `v-on` often get used repeatedly on the same element, and can get verbose when the information differing between each is **the directive argument** instead of the directive itself. **Shorthands thus help improve the signal/noise ratio by shortening the parts that are duplicated (`v-xxx`) and highlighting the parts that are different (the arguments).**
+
+The new `v-slot` syntax can also get the same problem in components that expect multiple slots:
+
+``` html
+<TestComponent>
+  <template v-slot:one="{ name }">Hello {{ name }}</template>
+  <template v-slot:two="{ name }">Hello {{ name }}</template>
+  <template v-slot:three="{ name }">Hello {{name }}</template>
+</TestComponent>
+```
+
+Here `v-slot` is repeated multiple times where the actual differing part is the slot name (denoted by the argument).
+
+The shorthand helps slot names to be more easily scan-able:
+
+``` html
+<TestComponent>
+  <template #one="{ name }">Hello {{ name }}</template>
+  <template #two="{ name }">Hello {{ name }}</template>
+  <template #three="{ name }">Hello {{name }}</template>
+</TestComponent>
+```
+
+# Detailed design
+
+The shorthand follows very similar rules to the shorthand of `v-bind` and `v-on`: replace the directive name plus the colon with the shorthand symbol (`#`).
+
+For default slot on component:
+
+``` html
+<!-- full syntax -->
+<foo v-slot="{ msg }">
+  {{ msg }}
+</foo>
+
+<!-- shorthand -->
+<foo #="{ msg }">
+  {{ msg }}
+</foo>
+```
+
+For named slots:
+
+``` html
+<!-- full syntax -->
+<foo>
+  <template v-slot:header="{ msg }">
+    Message from header: {{ msg }}
+  </template>
+
+   <template v-slot:footer>
+    A static footer
+  </template>
+</foo>
+
+<!-- shorthand -->
+<foo>
+  <template #header="{ msg }">
+    Message from header: {{ msg }}
+  </template>
+
+   <template #footer>
+    A static footer
+  </template>
+</foo>
+```
+
+The choice of `#` is a selection based on the feedback collected in the previous RFC. It holds a resemblance to the id selector in CSS and translates quite well conceptually to stand for a slot's name.
+
+Some example usage in a real-world library that relies on scoped slots ([vue-promised](https://github.com/posva/vue-promised)):
+
+ ``` html
+<Promised :promise="usersPromise">
+  <template #pending>
+    <p>Loading...</p>
+  </template>
+
+   <template #="users">
+    <ul>
+      <li v-for="user in users">{{ user.name }}</li>
+    </ul>
+  </template>
+
+   <template #rejected="error">
+    <p>Error: {{ error.message }}</p>
+  </template>
+</Promised>
+```
+
+# Drawbacks
+
+- The `#` symbol doesn't translate that well conceptually when used on a component for default slot:
+
+  ``` html
+  <foo #="{ msg }">
+    {{ msg }}
+  </foo>
+  ```
+
+  Maybe we only support the shorthand for named slots? For default slot usage on component, full `v-slot` usage is required. It would also be consistent with `v-on` and `v-bind` where the shorthand does not apply if there is no argument.
+
+- Some may argue that slots are not that commonly used and therefore don't really need a shorthand, and it could result in extra learning curve for beginners. In response to that:
+
+  1. I believe scoped slot is an important mechanism for building 3rd party component suites that are highly customizable and composable. I think we will see more component libraries relying on slots for customization and composition in the future. For a user using such a component library, a shorthand will become quite valuable (as demonstrated in the examples).
+
+  2. The translation rules for the shorthand is straightforward and consistent with existing shorthands. If the user learns how the base syntax works, understanding the shorthand is a minimal extra step.
+
+# Alternatives
+
+Some alternative symbols have been presented and discussed in the previous RFC. The only other one with similarly positive interest is `&`:
+
+``` html
+<foo &="{ msg }">
+  {{ msg }}
+</foo>
+
+ <foo>
+  <template &header="{ msg }">
+    Message from header: {{ msg }}
+  </template>
+
+   <template &footer>
+    A static footer
+  </template>
+</foo>
+```
+
+# Adoption strategy
+
+This should be a natural extension on top of the new `v-slot` syntax. Ideally, we'd want introduce the base syntax and the shorthand at the same time so users learn both at the same time. Introducing the shorthand at a later time runs the risk of some users only aware of the `v-slot` syntax and get confused when seeing the shorthand in others' code.
+
+# Unresolved questions
+
+- See Drawbacks No.1.
--- a/active-rfcs/0000-slot-syntax-shorthand.md
+++ b/active-rfcs/0000-slot-syntax-shorthand.md
@@ -10,11 +10,7 @@ Add shorthand syntax for the `v-slot` syntax proposed in [rfc-0001](https://gith
 # Basic example
 
 ``` html
-<foo #="{ msg }">
-  {{ msg }}
-</foo>
-
- <foo>
+<foo>
   <template #header="{ msg }">
     Message from header: {{ msg }}
   </template>
@@ -67,22 +63,6 @@ The shorthand helps slot names to be more easily scan-able:
 
 The shorthand follows very similar rules to the shorthand of `v-bind` and `v-on`: replace the directive name plus the colon with the shorthand symbol (`#`).
 
-For default slot on component:
-
-``` html
-<!-- full syntax -->
-<foo v-slot="{ msg }">
-  {{ msg }}
-</foo>
-
-<!-- shorthand -->
-<foo #="{ msg }">
-  {{ msg }}
-</foo>
-```
-
-For named slots:
-
 ``` html
 <!-- full syntax -->
 <foo>
@@ -107,6 +87,18 @@ For named slots:
 </foo>
 ```
 
+**Similar to `v-bind` and `v-on`, the shorthand only works when an argument is present. This means `v-slot` without an argument cannot be simplified to `#=`.** For default slots, either the full syntax (`v-slot`) or an explicit name (`#default`) should be used.
+
+``` html
+<foo v-slot="{ msg }">
+  {{ msg }}
+</foo>
+
+<foo #default="{ msg }">
+  {{ msg }}
+</foo>
+```
+
 The choice of `#` is a selection based on the feedback collected in the previous RFC. It holds a resemblance to the id selector in CSS and translates quite well conceptually to stand for a slot's name.
 
 Some example usage in a real-world library that relies on scoped slots ([vue-promised](https://github.com/posva/vue-promised)):
@@ -117,7 +109,7 @@ Some example usage in a real-world library that relies on scoped slots ([vue-pro
     <p>Loading...</p>
   </template>
 
-   <template #="users">
+   <template #default="users">
     <ul>
       <li v-for="user in users">{{ user.name }}</li>
     </ul>
@@ -131,16 +123,6 @@ Some example usage in a real-world library that relies on scoped slots ([vue-pro
 
 # Drawbacks
 
-- The `#` symbol doesn't translate that well conceptually when used on a component for default slot:
-
-  ``` html
-  <foo #="{ msg }">
-    {{ msg }}
-  </foo>
-  ```
-
-  Maybe we only support the shorthand for named slots? For default slot usage on component, full `v-slot` usage is required. It would also be consistent with `v-on` and `v-bind` where the shorthand does not apply if there is no argument.
-
 - Some may argue that slots are not that commonly used and therefore don't really need a shorthand, and it could result in extra learning curve for beginners. In response to that:
 
   1. I believe scoped slot is an important mechanism for building 3rd party component suites that are highly customizable and composable. I think we will see more component libraries relying on slots for customization and composition in the future. For a user using such a component library, a shorthand will become quite valuable (as demonstrated in the examples).
@@ -152,11 +134,7 @@ Some example usage in a real-world library that relies on scoped slots ([vue-pro
 Some alternative symbols have been presented and discussed in the previous RFC. The only other one with similarly positive interest is `&`:
 
 ``` html
-<foo &="{ msg }">
-  {{ msg }}
-</foo>
-
- <foo>
+<foo>
   <template &header="{ msg }">
     Message from header: {{ msg }}
   </template>
@@ -173,4 +151,4 @@ This should be a natural extension on top of the new `v-slot` syntax. Ideally, w
 
 # Unresolved questions
 
-- See Drawbacks No.1.
+N/A


active-rfcs/0000-slot-syntax-shorthand.md
```diff
@@ -0,0 +1,176 @@
...
+
+A shorthand, as the name entails, primarily aims to provide more succinct syntax.
+
+In Vue we currently provide shorthands only for two directives: `v-bind` and `v-on`:
+
+``` html
+<div v-bind:id="id"></div>
+<div :id="id"></div>
+
+<button v-on:click="onClick"></div>
```
henriqemalheiros:
```suggestion
<button v-on:click="onClick"></button>
```

active-rfcs/0000-slot-syntax-shorthand.md
```diff
@@ -0,0 +1,176 @@
...
+<button @click="onClick"></div>
```
henriqemalheiros:
```suggestion
<button @click="onClick"></button>
```

henriqemalheiros:
I like it. I don't mind being `#`, maybe I like it even better than `&` now.

I don't see the use of the shorthand on a default slot as a drawback because it'll only be used when destructuring the slot props and I think people will naturally read it as "slot". If it becomes a problem, a possible solution would be

```vue
<foo #default="{ msg }">
  {{ msg }}
</foo>
```

This could be optional, though, allowing people to choose between `#` and `#default` for more readability.

Akryum:
If we go with a shorthand, I think we should not allow `#="..."`, it's a strange syntax and isn't consistent with `v-on` and `v-bind`.

darrenjennings:
Posting this also as it seemed that it seemed favorable by some:

```vue
<foo &default="{ msg }"> <!-- to @Akryum's point -->
  {{ msg }}
</foo>

<foo>
  <template &header="{ msg }">
    Message from header: {{ msg }}
  </template>

   <template &footer>
    A static footer
  </template>
</foo>
```

Benefits:
- It is communicated in languages like C to yield the address of a pointer. So it could be translated to mean "what's at the address of this slot?"
- Compared with `#`, the `&` operator is not used as much as in HTML templates, maybe to denote character codes (e.g. `&nbsp;`). `#` is the id selector in css as well as the hex code prefix, but both operators aren't common inside the templates themselves, in my experience.

Drawbacks:
- In javascript it is used as logical `or` operator (`&&`) and it is also used as a bitwise operator

As @yyx990803 mentioned before, most of the qualms with the aesthetic appeal can be fixed via syntax highlighting.

jacekkarczmarczyk:
May I propose one more shorthand - **?**. It's familiar symbol in web world, used to pass parameters, the semantic here is quite different but  personally i find it less weird looking that the other ones. Sorry if it was mentioned before and immeditely rejected

```vue
<foo ?="foo">
  <bar ?="bar">
    <baz ?="baz">
      {{ foo }} {{ bar }} {{ baz }}
    </baz>
  </bar>
</foo>

<foo>
  <template ?header="{ msg }">
    Header msg: {{ msg }}
  </template>

  <template ?footer="{ msg }">
    Footer msg: {{ msg }}
  </template>
</foo>
```

brophdawg11:
From the proposal:

> Maybe we only support the shorthand for named slots? For default slot usage on component, full v-slot usage is required. It would also be consistent with v-on and v-bind where the shorthand does not apply if there is no argument.

Echo-ing what @Akryum said above, I agree with this.  Disallowing the shorthand on default slots seems very consistent and clear to me.

Justineo:
For consistency, I think we can define a shorthand as a replacement for `v-<direction name>:`, which means it requires the directive have arguments:

* `v-bind:args` &rarr; `:args`
* `v-on:args` &rarr; `@args`
* `v-slot:args` &rarr; `#args`

For the argument-less case, we enforce the use of the full syntax. To declare a default slot scope, either of the following usage will be legitimate:

* `v-slot="{...}"`
* `#default="{...}"`

znck:
The shorthand syntax has quite grown on me, I like both full syntax and shorthand. But for a hypothetical component accepting multiple named slots, I'd prefer shorthand.

``` html
<TestComponent v-slot="{ name }">
  <template v-slot:one="props">Hello {{ props.name }}</template>
  <template v-slot:two="{ name }">Hello {{ name }}</template>
  <template v-slot:[dynamicSlot1]="props">Hello {{ props.name }}</template>
  <template>Bye {{ name }}</template>
</TestComponent>

<TestComponent #="{ name }">
  <template #one="props">Hello {{ props.name }}</template>
  <template #two="{ props }">Hello {{ props.name }}</template>
  <template #[dynamicSlot1]="props">Hello {{ props.name }}</template>
  <template>Bye {{ name }}</template>
</TestComponent>
```

Even the default slot shorthand feels quite okay, I would like to extend it to `v-bind`.

``` html
<div v-bind:foo="bar" v-bind:[foo]="bar" v-bind="someObject"></div>
<div :foo="bar" :[foo]="bar" :="someObject"></div>
```

Or even `v-on` can use shorthand too.

``` html
<div v-on:foo="bar" v-on:[foo]="bar" v-on="someObject"></div>
<div @foo="bar" @[foo]="bar" @="someObject"></div>
```

jameshulse:
When I read the initial proposal and saw the alternative shorthand `&` I really liked it - the similarity to passing an object by reference clicked with me.

Now that I've seen the `#` alternative, it actually feels clearer because the shorthand is linked more to the name of the slot and less to the props. The `#` is nicely reminiscent to how anchors work in a URL.

leopiccionia:
Like `()`, `#` is also used in Angular templates:
1. `#header` is similar to `ref="header"` (`this.header` is an [`ElementRef`](https://angular.io/api/core/ElementRef))
2. `#formName="ngForm"` is used to bind a "template-driven form" for validation, etc. (`this.formName` is a [`NgForm`](https://angular.io/api/forms/NgForm))

Newcomers with Angular experience may still be confused, but `#` is much rarer in Angular templates than `()`. The second form, particularly, may be less confusing (`ngForm` is the only valid literal I've ever seen).

chrisvfritz:
I've been thinking about this a lot today and the more I do, the more strongly I'm against a shorthand for `v-slot`. I've tried to summarize what I understand to be the advantages and disadvantages below.

### Undisputed Advantages

- Whenever using `v-slot`, users could save 5 keystrokes.

### Counters to Claimed Advantages

- For some people, `#header` stands out more visually than `v-slot:header`. I'm not one of these people, but **if the goal is to make named slots more visually distinct, that's a problem best solved by tooling, not new syntax**. We could just have Vetur and similar tooling use distinct highlighting for `v-slot:xxx`. An editor plugin could even treat `v-slot` as a ligature for `#` if people really want to see that symbol in place of `v-slot`.

- **As a less frequently used feature that's only ever used once per element, the potential benefit is miniscule compared to the benefit of the existing shorthands**. Here are some stats after taking a look through a few large projects I have access to:

  - On average, 17% of components used `slot` or `slot-scope`.
  - In components using `slot` or `slot-scope`, the shorthand could be applied 2 times, on average.

  Comparing that to `v-on`:

  - On average, 53% of components used `v-on`.
  - In components using `v-on`, the shorthand could be applied 3 times, on average.

  By that math, there's about 5 times as many uses for `v-on`'s shorthand. I didn't bother including stats for `v-bind`, because it's in a league of its own. 

- Through the poll in the other thread, most people were for the shorthand, but it was definitely not an overwhelming victory. A fair number of people - even on the core team like myself, @shentao, @phanan, and probably others - would actively _avoid_ the shorthand and possibly even ban it from our projects, because we feel the very small convenience isn't worth the lost explicitness and potential confusion. As a general rule, **when at least several core team members feel a feature would make Vue _worse_ and would actively avoid it, I think the advantages should have to be incredibly compelling**. In this case, the only inarguable benefit I've seen is being able to save keystrokes - and probably only a few tweets worth in a 500 component project.

- **No one is actually using `v-slot` in the wild yet, so by adding syntax sugar at this stage, we're solving a problem that doesn't exist.** It _might_ exist later. Maybe when `v-slot` is released, users will constantly complain about how verbose it is. In that case, we could always come back and investigate this further, but it's easier to add a feature later than to remove it later.

### Disadvantages

- **It would increase the API surface area, meaning there's more for everyone to learn in order to be able to read Vue code, without allowing any new possibilities.**

- **Shorthands are more difficult to google**, which is important for users who see the syntax for the first time in examples, blog posts, articles, etc. Our current shorthands for `v-bind` and `v-on` compensate for this by being introduced very early in the docs. Named and scoped slots aren't even in the Essentials. To give some context, every time I give a workshop, I do a poll to see how much of the docs participants have read. Looking only at my _advanced_ workshops, the number of people who've read all the Essentials is always between 40% and 55%. From our traffic analytics, it seems that out of the minority who do read all the Essentials, about 35% also read the guide on Component Slots, where named and scoped slots are introduced. This suggests that _optimistically_, **more than 80% of users will never read the docs on this shorthand syntax and are likely to encounter it for the first time when they see it in an example, blog post, tutorial, etc. Unless that content is specifically about slots and the shorthand, that means a potentially a terrible onboarding experience for named/scoped slots for _most_ users.**

- In a former career, I was a linguist and language teacher. From research, I found out that when learning a new word, adults needed to be introduced to it 7 times (on average) before they'd be able to integrate it into their own vocabulary. One exception was "false friends", which are words that look/sound similar to a word with a completely different meaning in a familiar language. Extremely common false friends are sometimes _easier_ to learn, strangely, but less common false friends typically require orders of magnitude more reinforcement. In the linguistic context of HTML/templating, we're introducing `#` as a new word that's also a false friend, because outside of Vue templates, it's commonly understood to mean referencing/adding an `id` attribute (and means something else in Angular). So **this syntax will not only be initially confusing for people unfamiliar with the shorthand, but worse - they'll often falsely assume they already know what it does _and_ because it's a relatively less common false friend, it will likely take most people orders of magnitude more reinforcement to learn.**

- As @Akryum and others have already mentioned, for the object syntax for `v-bind` and `v-on`, we decided not to allow shorthands, because `:=""` and `@=""` didn't provide enough context (and looked pretty strange). Given that, it seems inconsistent to allow `#=""` when the same problem exists here.

- **I fear this could set a precedent where power users (especially solo devs who don't have to think about team dynamics) will start asking for more shorthands, making Vue look more and more like Lisp.** It could be difficult to say no, too. Unlike `v-bind` and `v-on`, named/scoped slots are less commonly used feature that's not even in the Essentials of our guide, where only a single instance will ever exist on an element. I can easily imagine just as compelling arguments for adding shorthands for class bindings (they're used so frequently and the attribute's value is often really long), `v-model` (it already has modifiers and in Vue 3, will also likely gain an argument to replace `v-bind.sync`), etc. Then come requests to extend custom directives to allow custom shorthands to be defined...

yyx990803:
@chrisvfritz 

## Counters to Counters of Claimed Advantages

> If the goal is to make named slots more visually distinct, that's a problem best solved by tooling, not new syntax

Just because it can be helped by tooling doesn't mean we can ignore visual distinctiveness from syntax discussions. Also, we cannot assume that everyone will be using the same tooling.

> As a less frequently used feature that's only ever used once per element, the potential benefit is miniscule compared to the benefit of the existing shorthands.

The primary reasoning for including a shorthand is not just saving keystrokes - instead, it's about increasing the signal/noise ratio in cases where it gets used in high density. In that sense, the absolute number of times it occurs within the project is not the sole determinant of its overall benefits. On the other hand - I don't think 17% is an insignificant usage rate, and the ratio may get higher for projects that rely on component libraries with a primarily slot-based API.

> When at least several core team members feel a feature would make Vue worse and would actively avoid it, I think the advantages should have to be incredibly compelling.

If I'm not mistaken, I think @shentao and @znck have both changed their minds and are now ok with `#` being the shorthand. I'd encourage team members who were involved in earlier discussions to read the latest RFC and voice their opinions here directly.

> No one is actually using v-slot in the wild yet, so by adding syntax sugar at this stage, we're solving a problem that doesn't exist. It might exist later. Maybe when v-slot is released, users will constantly complain about how verbose it is. In that case, we could always come back and investigate this further, but it's easier to add a feature later than to remove it later.

The verbosity and lack of visual distinctiveness has been a problem since the `slot` + `slot-scope` time - `v-slot` was partially conceived to remedy that, and the shorthand is an extension on top of `v-slot` to take it a step further.

As mentioned in the adoption strategy - it actually creates a problem if we ship the base syntax and the shorthand separately. **Shipping them together ensures everyone reading about the `v-slot` syntax also know about the shorthand.** Shipping separately runs the risk of people only know about `v-slot` and miss the shorthand later.

## Counters to Disadvantages

> Increased API surface

The API surface increase is objectively minimal. If someone already knows the `v-slot` syntax (which **is** required to be able to read Vue code), the learning needed to know its shorthand is minimal since the shorthand rules *can* be 100% consistent with existing shorthands if we disallow `#=` usage.

> Shorthands are more difficult to google

This is true - but I doubt it is a problem significant enough to block this RFC, since Google isn't the only way to find answers to syntax issues. There are forums, chat rooms, alternative search terms like "Vue shorthands", and the fellow developer who wrote that code to ask from.

> I fear this could set a precedent where power users (especially solo devs who don't have to think about team dynamics) will start asking for more shorthands, making Vue look more and more like Lisp.

I think such a fear is unwarranted. The discussion we are having right now shows good arguments from both sides and actually helps us argue against unnecessary shorthands in the future. Landing it in now way indicates we are open to more shorthands without proper, case-by-case discussions.

yyx990803:
Updated: dropped `#=` usage and now enforces full syntax (`v-slot`) or explicit name (`#default`) for on-component default slot usage.

posva:
>  I'd encourage team members who were involved in earlier discussions to read the latest RFC and voice their opinions here directly.

I like that `#` was dropped in favor of `#default` because we cannot do `:="bindings"` nor `@="events"`. To me, I feel a lack of consistency between `v-slot` and `v-bind`/`v-on` because you can use multiple `v-bind`/`v-on` on a single element but you can only use a single `v-slot`. I think that as long as we clearly error this during compilation, it's fine though.

Differently, from bind and on, which are completely necessary to build a Vue application, a person may develop an application without ever developing components using slots. Only using them through 3rd party components. I think it's fine to use the classic syntax in these scenarios, leaving the shorthand syntax for more advanced and intensive scoped slot usages. Even mixing both syntaxes in a project sounds fine.

However, I'm also concerned about the syntax looking magical, especially if we add more like `v-slot:[name]="hey"`, making Vue hostile to beginners and not as intuitive as it was before. I will show this to other developers with little experience in Vue to collect some extra Feedback.

Overall I'm for the shorthand, one of the main points for me is [Drawbacks i](https://github.com/vuejs/rfcs/blob/slot-shorthand/active-rfcs/0000-slot-syntax-shorthand.md#drawbacks)

bencodezen:
From my experience, one of the major selling points of Vue has always been the fact that it is intuitive. It doesn't take much effort to read or comprehend what's going on regardless of your JS experience level because you all did that good of a job. All of the APIs and mental models required to understand Vue fit within the traditional frontend technology stack. 

That said, the shorthand syntax for v-slots is one of the first times where I had to stare and question what was going on. After all, as someone who writes a lot of CSS, seeing `#` immediately made me think of an ID. So from a concept mapping perspective, I'm already on the wrong path.

The reason the shorthand of `v-bind` and `v-on` work so well is that they used symbols which were unique and intuitive to what was happening. They are subtle but distinctive characters that were practically instantaneous to learn and acquire. I've never had a student tell me that the shorthand didn't make sense .

While I know that many of the advanced Vue.js developers may be okay with the shorthand, I want to make sure we are considering new and future new Vue.js developers who won't have a chance to weigh in on this. I would be remiss if we started to lose how approachable Vue is to developers of all walks of life over something like shorthand syntax.

Anyhow, just my two cents and I welcome any counter points if I missed something. Cheers!

yyx990803:
@bencodezen 

I think there's a fundamental flaw in this argument: you can't judge a shorthand by people's reaction when they are exposed to it without proper context - that's not how we teach the framework, and that's not how people learn.

We tell people what `v-bind` and `v-on` do before introducing shorthands for them. That's exactly what we do in the docs.

Similarly, this proposal has the pre-requisite of proper context from the new slot syntax proposed in #2. The shorthand should and will only be introduced AFTER the user has properly learned how `v-slot` works. Once they understand how `v-slot` works, the full-syntax to shorthand translation is a natural progression. In addition, slot is also definitely introduced AFTER `v-on` and `v-bind`, so the user is already aware of the shorthand conversion rules for `v-bind` and `v-on`:

- `v-bind:id` -> `:id`
- `v-on:click` -> `@click`
- `v-slot:header` -> `#header`

Another aspect of the analogy here is that `v-bind` and `v-on` themselves are very self-explanatory. In comparison, slots itself, especially including scoped slots, is an inherently complex concept in its own.

So when I see "stare and question what was going on", I think it's a combination of:

1. The inherent complexity of the slot mechanism
2. Unfamiliarity with the new `v-slot` base syntax
3. (Because of 2) confusion of the `#` symbol and its relation to CSS ids

I'd say it's inaccurate to attribute the confusion all to the shorthand itself, and I don't think users learning about the new syntax from the documentation or structured materials will have a similar problem. I honestly doubt this shorthand can be so disastrous as to cause Vue to lose its approachable-ness - that's a bit hyperbole to me.

bencodezen:
@yyx990803 

My apologies if it came across as a hyperbole. It was more an expression of concern more than anything else and a thought experiment at best.

I agree that with the appropriate context and background, shorthand syntax is merely a conversion rule for a person's mental model. The only point I wanted to raise was that the interesting phenomenon I found with `v-bind` and `v-on` is that people could almost reverse engineer the symbol to its purpose without any explanation because it was just that intuitive.

So my comment on the `#` shorthand is simply based on the notion that it's not as intuitive in comparison to the other shorthands. Other than that, love all the great work going on here for v-slot and for v3.0 in general!

yyx990803:
@bencodezen no worries. As I have explained, the fundamental cause is because `v-on` and `v-bind` themselves are simple concepts whereas slot is an advanced concept in itself. It's like you can guess simple words from a different language that describes everyday objects, but unlikely to be able to do that for words that describes abstract ideas.

chrisvfritz:
## Re-evaluating premises

@yyx990803 I think we're starting the get to the core of the disagreement here. I worry that some of the premises you're building on may not be accurate. Based on the data I've seen from analytics and workshops, and my background in educational psychology, I believe I have strong evidence that:

1. Most users will not learn about `v-slot` through the docs, so are likely to see this shorthand for the first time in a blog post, tutorial, etc that is likely to assume knowledge of the shorthand.

2. Learning a shorthand in addition to the base `v-slot` concept adds a higher cognitive cost to already more difficult-to-learn concepts.

If these are true, then then your two ✅ premises would be inaccurate and Vue's overall approachableness _would_ suffer for most users (not just some outliers) that would persist for as long as we keep getting new users. I'll explore these points now.

### 1. Most users will not learn about `v-slot` through the docs

> For future new users [the order of concepts in the docs] makes the shorthand a rather natural step in the learning process, so we are not making Vue harder to learn.

You're arguing the learning path is pretty nice _if_ people learn about slots for the first time from the docs, but it looks like that wouldn't happen. **Based on the data I mentioned earlier, it seems _more than 80%_ of users are not introduced to slots through the docs.** That means when they're first exposed to `v-slot` and its shorthand, it will likely be through an article/tutorial/example that will probably _assume_ knowledge of the shorthand unless specifically introducing `v-slot`.

> For existing users upgrading to 2.6: they will read the new documentation and learn about the new `v-slot` syntax along with the shorthand. Again, shorthand is only introduced AFTER the base syntax.

Again, based on the data I have access to, this seems extremely unlikely. Looking at the analytics for our docs, **we didn't see _any_ spike in page views when v2.5 released with new features**. This leads me to assume that once again, when v2.6 is released, the number of people revisiting the docs will also be minimal. This also matched my anecdotal experience, as I still meet _a lot_ of users who don't know about scoped slots and of those who do know, almost everyone says they learned about it from a blog post or course.

I've also witnessed this behavior first-hand in my consulting work. When people upgrade a dependency to a new minor version, they will _almost always_ just run tests (if they exist) and load up the app to manually check if things still seem to be working. **The vast majority of people aren't reading changelogs, release announcements, or coming back to the docs to read up on new/updated features.**

I agree we'd be in a better place if existing users did these things, but all my experience and all the data I have access to indicate that it's not the reality.

### 2. The shorthand adds a higher cognitive cost to already more difficult-to-learn concepts

This is where my ed psych background comes in. Learning is messier than we usually think of it. People won't read a section and immediately have it cemented and organized in their minds. So when learning about named/scoped slots for the first time, users will still be mentally digesting that information even after they feel they "get" it. It won't be until sleeping on it, being re-exposed to it a few more times, and using it a few times, that it'll really be a solid foundation they can easily build new knowledge on. So **learning the shorthand directly afterwards _will_ increase the time and effort it takes to learn named/scoped slots in general, by increasing the total number of discrete concepts that need to be digested at once**. 

The shorthand is worse than just one extra concept though, it's an extra layer of abstraction that users will have to mentally translate before they're "fluent", when reading examples in that shorthand.

## A deeper look at the tooling alternative

If I understand correctly, we both agree that all the benefits of this proposal could be accomplished through tooling - your worry is that not everyone will be using the same tooling. That means there would be duplicated effort in implementing the syntax highlighting (and potential ligature) and potentially a delay in when some users have access to it. 

First, **since this isn't a feature that opens any new possibilities, even from a development perspective, is it really important if some people don't have access to it right away? Plus, I don't get the impression that even those in favor of the proposal want it very badly.** Apart from you, the most impassioned arguments for the shorthand have had a tone of mild enthusiasm, bordering on indifference:

> it's beginning to grow on me

> I like both full syntax and shorthand

> I like it

Or the strongest endorsement I saw:

> The `#` is nicely reminiscent to how anchors work in a URL

Which I think should actually be an argument _against_ the shorthand, since slots have nothing to do with anchors.

Meanwhile, _most_ people who commented on the shorthand in the 3 issues (including those quoted above) have expressed concerns about the possible confusion, even after settling on `#`.

On the other hand, **making this shorthand opt-in through tooling would solve every disadvantage. We wouldn't have to cover it in documentation, it would never appear in community learning resources, and those who find it confusing because `#` means something completely different in HTML/JS and Angular would never have to see it.**

## Sidenotes

### GitHub polls

As a sidenote, out of those people who would enjoy it:

> we have about 80% of users who voted in favor of introducing a shorthand

That 80% is exclusively power users and others with enough free time to browse random Vue issues. **A poll on GitHub is so unrepresentative of our overall userbase that when the perspective of new/normal users is concerned, I don't think the results are relevant.**

### A similar precedent

**This actually reminds me of the rejected proposal to allow `:foo` to be shorthand for `:foo="foo"`.** It would save keystrokes and reduce syntax noise, but we decided against it, twice, because your intuition (and later, user testing) suggested it would confuse a lot of people - and resolving that confusion would be difficult through Google. We decided the mild convenience just wasn't worth it, even though that pattern is even more common than named/scoped slots. It didn't even matter that in user testing, most users eventually correctly guessed what `:foo` was shorthand for, suggesting that it was _somewhat_ intuitive.

While providing similar benefits, the argument for `v-slot`'s shorthand is even weaker, because:

- It's a less frequently used feature.

- In guessing what `#` means, most (if not all) users would be wrong, because `#foo` already means `id="foo"` in HTML/JS and `ref="foo"` in Angular.

Akryum:
I'm still not convinced yet another syntax is beneficial just to save 5 key strokes. We already have this new dynamic argument syntax which is adding some noise in the template but which actually opens new possibilities, so I like it, but this one is I'm really not sure... Maybe we should try to not end up closer to Angular with a absurd amount of special characters everywhere in the templates?

phanan:
I'd tend to agree with @chrisvfritz and @Akryum here. Slots (especially scoped) are a more advanced concept, mostly used by power users (I personally had a hard time wrapping my head around the idea, and am still not using it as frequently or fluently as I want to be). This essentially means several things:

* The vote we have on GitHub, and even this discussion, only involve users who are familiar with the concept and its practical usage enough to raise their opinions. Now, we can go the easy road by saying that this is OSS with every discussion open and those who don't contribute should just accept whatever decision made by those who do, but being notoriously known as an extremely beginner-friendly and approachable framework, I think Vue can do better than that.

* The shorthand, though with the undeniable advantages as stated by @yyx990803, would arguably benefit only those power users. Beginners and immediate-level users will instead be introduced with a weird, confusing syntax which is of no benefits for them. This is where the approachable-ness of Vue is prone to harm. 

* Even in the case of power users, the benefits of the shorthand would be minimal because, well, as power users, they must already have tools and tricks up their sleeves to achieve the same result without the need of support from the framework level – I'm talking about shortcuts, autocompletes, even editor extensions and whatnot. 

Last but not least, for the arguments saying this is a small upfront cost for a long run benefit, I am not so sure about that. This upfront cost might just be costly enough for me for example, as a team lead, to disallow the use of the syntax in our codebase entirely, to not raise the bar for newcomers unnecessarily. This again defeats the purpose of introducing a concept whose audience consists of mostly power users.

yyx990803:
> I'm still not convinced yet another syntax is beneficial just to save 5 key strokes

It's not just about saving 5 keystrokes. As stated in the proposal, it's also about increasing the signal/noise ratio when it is used in high density. "yet another syntax"

> Maybe we should try to not end up closer to Angular with a absurd amount of special characters everywhere in the templates

That is certainly something we want to avoid, which is exactly why we are having this lengthy discussion here on the introduction of one single extra special shorthand. I'm pretty sure we are not adding yet another shorthand any time soon (maybe never).

> Slots (especially scoped) are a more advanced concept, mostly used by power users... Even in the case of power users, the benefits of the shorthand would be minimal

This brings up an interesting topic on perspective here - the dynamics of power users vs. beginners. It seems some of the arguments here makes the assumption that DX for beginners automatically trumps that of power users. I don't think that is the case - DX for both groups are important, but it has to be measured in a case by case manner.

I think it's good that we have a part of the team that lean towards conservativeness whenever it comes to approachability, but we also have to realize that Vue is not **just** for beginners. We have been around for 4 years and we have people building sophisticated systems using Vue in all kinds of settings. We sure do have a lot of power users as well as beginners. Should we always lean towards a particular side? No. Our ultimate target is **all** Vue users. In this particular argument, I feel that the negative impact of the shorthand may have on beginners is being exaggerated, while the cognitive benefits for power users are being dismissed as just "saving keystrokes". That, in my opinion, is not a balanced take, and I'm willing to play the advocate for the power users here.

Overall, I'm still not convinced that the potential downside for beginners outweigh the potential upside for power users.

yyx990803:
> it looks like only about 35% of people who read all the Essentials also go on to read the slots page

**This doesn't mean they didn't learn slots from the docs.** It only means they didn't get to slots on their first reading. I'm afraid this number is fundamentally inaccurate.

> The same was true for introducing slot-scope, but I still meet a lot of users, even now, who don't know about scoped slots.

This actually means the shorthand will not affect them either.

> What's your evidence that people seeing #xxx for the first time, with no context, would assume it's short for v-slot:xxx?

Because it most likely only appears in the following form:

``` html
<SomeComponent>
  <template #header>
    <div>Header message</div>
  </template>

  <template #item="{ item }">
    <div class="item">{{ item.text }}</div>
  </template>

  <template #footer>
    <div>Footer message</div>
  </template>
</SomeComponent>
```

In this case, `<template>` is being passed to a component as children. **This only happens when you are using slots in Vue.** Put it another way, **it's almost impossible to see the `#` syntax without proper context.** So most of the arguments about that it could be confused for something else doesn't really hold.

chrisvfritz:
> This brings up an interesting topic on perspective here - the dynamics of power users vs. beginners. It seems some of the arguments here makes the assumption that DX for beginners automatically trumps that of power users. I don't think that is the case - DX for both groups are important, but it has to be measured in a case by case manner.

@yyx990803 I agree. That's why I, as a power user myself, suggested this as opt-in editor tooling. That way, beginners are unaffected and the power users who might find this helpful can enable it for their own code. Isn't that the best of both worlds?

I think this is more than power users vs beginners though. I think there's also a solo devs vs team dynamic. I think power users who also happen to be solo devs are most likely to gravitate toward this feature, because their knowledge is cemented enough that learning one more thing is easier. They never have to worry about the cost of other people learning it. Or, they're students and not even building real apps. (I think not many people besides students have the free time to browse random Vue issues. 😄)

On the other hand, those of us working in a team would rather not suffer productivity by having team members stumble over `#header` and not being sure what it means, or assuming wrongly, then having to explain it. We have more important things to focus on. That's why I and other team leads like @phanan would ban it from our projects, even though we're power users. We also want Vue to scale - and part of scaling is working well in team settings.

yyx990803:
> I think there's also a solo devs vs team dynamic. I think power users who also happen to be solo devs are most likely to gravitate toward this feature, because their knowledge is cemented enough that learning one more thing is easier. They never have to worry about the cost of other people learning it.

Valid point - but I would argue that there are two different problems here:

1. Some team members using a feature that the rest of the team are not even aware of: this can happen with or without shorthands.

2. Some team members using the shorthand, while other team members know about the feature but not aware of the shorthand: this seems to be a much less likely scenario, and blocking the shorthand based on that seems quite a stretch to me.

Banning certain features is a team choice. If Vue can work without any sort of enforced team opinion, then we wouldn't have the need for the ESLint plugin or the style guide.

On the other hand - it seems to me a better approach would be a proper onboarding document/process that primes any new member of the features the team is using to make sure everyone is on the same page, rather than hoping that they somehow will figure out things they didn't know themselves.

chriscalo:
[Private class fields][1] in JavaScript (partially implemented in [Babel 7.3.0][2]) use the `#` symbol:

``` js
class Counter extends HTMLElement {
  #x = 0;

  clicked() {
    this.#x++;
    window.requestAnimationFrame(this.render.bind(this));
  }

  constructor() {
    super();
    this.onclick = this.clicked.bind(this);
  }

  connectedCallback() { this.render(); }

  render() {
    this.textContent = this.#x.toString();
  }
}
```

I could see a future where people have a learned association that prefixing a property with a `#` means it's private. Does that cause any problems here?

[1]: https://github.com/tc39/proposal-class-fields
[2]: https://babeljs.io/blog/2019/01/21/7.3.0

yyx990803:
@chriscalo those are used in a completely different context so unlikely to cause any confusion.

yyx990803:
Did a [poll on Twitter](https://twitter.com/vuejs/status/1087401587012759552). As of this writing we have over 1,000 votes, where 55% guessed correctly, 16% guessed wrong, and 29% wasn't able to guess what it means. The percentage was very stable ever since it got over 100 votes, so I think it is a big enough sample size. I also think Twitter represents a diverse enough user base - probably including a decent percentage of casual or even non-users who are simply interested in participating in the design discussion.

The follow up poll also shows that over half (56%) of the voters who wasn't able to guess what it means do not know how slots work. I should have phrased it to include those who guessed wrong as well - but I think it's fair to assume a decent percentage of them also lacked prior slot knowledge. Let's assume it's a bit lower, maybe 40%.

So if we filter out users who don't even know slots, we get (55% / (55% + 29% * (1 - 56%) + 16% * (1 - 40%)) ~= 71%.

That is to say, for users with prior slots knowledge, **over 70% were able to guess what the syntax means just by looking at it point blank, without reading any documentation, or even learning about the new `v-slot` full syntax.** That means the syntax is intuitive - or, at the very least, not confusing. I believe with proper documentation or learning, the vast majority of users will be able to "get it" right on the spot.

victornpb:
Idk if that’s up to discussion but what about
‘‘‘html
<template $header>
’’‘

Like ’$’ standing for ’slot’

The only thing I was confused about slots when I first learned about it was the a lot tag vs slot attribute, my first intuition was that I had to wrap my code in a slot tag when I’m fact I had to add a slot attribute.

thelinuxlich:
I can't wait to use this with Vuetify!

chrisvfritz:
@yyx990803 I don't agree with @bencodezen that the userbase on Twitter isn't representative enough for this kind of research. I think it can be a reasonable proxy, as long as the poll is well-designed.

But the way you conducted that poll definitely does not follow user testing best practices as I understand them. First, the phrasing makes it seem like a contest ("Can you guess what this syntax does?"), where people are now invested in guessing "correctly". Second, by providing the "correct" answer at all, many users will read it before answering (even though you told them not to), biasing their answer toward guessing correctly. Third, the example you provided is probably the _most ideal_ scenario for guessing correctly, not the most typical. Fourth, while it makes counting up votes easier, we lose so much valuable nuance in the data by having people choose from limited options, rather than having them explain exactly what their guess is.

Then you assumed that the votes of people not familiar with slots don't matter - which I'm confused by. If new users can't even intuit that this is for content distribution, that's still important! So you can't just make up a percentage (I've been trying to keep all my numbers coming from real data) and decide to throw out the votes you don't like. 😄

But keeping all the data, and even knowing the design of the poll will skew results toward the answer you want, that _still_ left 46% of people who would be very confused by this syntax and having a more difficult time Googling it. That's only 6% below the number of people who were confused by the `:xxx`/`:xxx="xxx"` shorthand when we tested it, which we all agreed was unacceptable (despite the fact that `:xxx` had a higher impact than this shorthand).

I think **we can still make some conclusions based on the poll though. It tells us that even with the most ideal example, _at least_ 46% of people would be very confused when seeing this syntax.** Given that we've also agreed shorthand confusion is much more difficult to resolve, since `#header` will be difficult to Google, that seems like a pretty huge negative impact.

Earlier, **I asked you what percentage of confusion you would find acceptable, because I think it's important to have that discussion.** Even for `:xxx`/`:xxx="xxx"`, which _I_ initially proposed (and would have had a higher impact), I was unhappy with anything above 10%. So with 52% confusion, I immediately withdrew my support. I had to agree it just wasn't worth the small reduction in syntax noise - especially when tooling can do the same job without the disadvantages. This is an even smaller reduction in syntax noise, with a very similar initial confusion rate, while also adding complexity to an already more complex feature and is so controversial that some core team members would ban it from their projects, requiring yet more tooling. So **why is this proposal getting special treatment? Have I missed some critical difference?**

Also, **you never addressed [my full argument](https://github.com/vuejs/rfcs/pull/3#issuecomment-455917710) about moving this to opt-in tooling. If we can avoid all the disadvantages, while still giving power users all the benefits, isn't that a better solution?**

yyx990803:
> the phrasing makes it seem like a contest

I agree that can potentially skew the result, however - Twitter polls are anonymous. **People don't get any reward for voting as guessed right when they did not**, so the skewing is rather limited.

> assumed that the votes of people not familiar with slots don't matter

I'm not saying it doesn't matter. These users are indeed a relevant metrics for the percentage that would have a hard time Googling it. However it's definitely not relevant when evaluating the intuitiveness of the shorthand. If a user doesn't understand slots, **they don't even have the possibility of not being confused by the shorthand.** This happens even without the shorthand, so counting these confused users as victims of the shorthand would result in a very different narrative, which leads to:

> ...still left 46% of people who would be very confused by this syntax

1. As explained above, it's 29%, not 46%.

2. "very confused" is inaccurate. This percentage is the combination of all level of confusion, from mild to very. You can't just cast them all the one extreme.

3. They are confused because they were asked to **look at some new syntax, which is in turn a shorthand of another new syntax, which is in turn a syntax for a concept that is inherently complex in itself, point blank, without any other context.** Ignoring that context and phrasing as if that percentage of people will get confused by the syntax under any situation distorts the narrative quite a bit.

>  I asked you what percentage of confusion you would find acceptable, because I think it's important to have that discussion

Again, context is important. "29% of users don't get it even after proper explanation" is completely different from "71% of users get it even without proper explanation". I think the former is not acceptable while the latter is.

> why is this proposal getting special treatment? Have I missed some critical difference?

Because of what's explained above. Context is important.

> you never addressed my full argument about moving this to opt-in tooling

That doesn't seem to address the situation when an advanced team member opts-in without telling the rest of the team, or when beginners learn from open sourced code that happens to opt-in. They'd still have the exact same problems. In fact, this makes it even worse: because we actively try to hide the information, beginners that do run into it will have an even harder time finding resources about such usage, since it's not even in the docs.

---

I think we have conducted enough discussion here. I think what we can agree on is that *some* users want it, and *some* users do not. Among those that do not, some simply dislike it and prefer the full syntax; some actively want to prevent it from shipping due to concerns about increased learning barrier.

I've read all the feedback, I've looked at all the data. I have no problem admitting that I personally am strongly in favor of the syntax, from a user's perspective. I also admit that this will result in some learning churn, but my assessment is that the churn is acceptable. We still have disagreements, but the arguments have started to shift towards minute details in things that either cannot be empirically proven, or - in cases where data has been collected - how to interpret the data. The arguments can be continued indefinitely and at this point I believe the time investment from core team members is starting to hinder progress in other parts. This is probably the first time I have to say this, but it has to be said: as the BDFL, I've made my decision, and we are shipping it. Case closed.

laander:
@yyx990803 The majority of the discussion in this RFC seems to have been about whether to add a shorthand symbol or not, and I understand your position on wanting to move pass that point.

Given the Twitter thread, comments on the gist and comments in the previous RFC, I feel that the choice of symbol still remains undecided/undiscussed. It seems to me that not an insubstantial amount of concerns is regarding the `#` symbol.

There's been quite a few different alternatives suggested the past days, so I'd suggest we have a vote to gauge the preference sentiment in the community. It could be here on Github or on Twitter.

To keep it pragmatic, a maximum 5 different options should be presented. Here's all the different ones I've seen in the wild so far:
`#  $  &  ^  ||  @  %  ?  *`

What's your thoughts?

yyx990803:
@laander choice of symbol has already been extensively discussed in #2 - in fact it was changed to `#` based on feedback. At this point I believe there will be people dissatisfied with any symbol choice - however most of the opinions are purely subjective and is more or less bikeshedding. Function-wise none of these symbols are inherently superior to another - we have to pick one that works, because it's impossible to pick one that pleases everyone.

PR #4, opened 2019-01-16
Proposal for dynamic directive arguments

yyx990803:
``` html
<div v-bind:[key]="value"></div>
<div v-on:[event]="handler"></div>
```

[Full rendered proposal](https://github.com/vuejs/rfcs/blob/dynamic-directive-arguments/active-rfcs/0000-dynamic-directive-arguments.md)

### DIFF
--- a/active-rfcs/0000-dynamic-directive-arguments.md
+++ b/active-rfcs/0000-dynamic-directive-arguments.md
@@ -1,0 +1,112 @@
+- Start Date: 2019-01-16
+- Target Major Version: (2.x / 3.x)
+- Reference Issues: https://github.com/vuejs/rfcs/pull/2, https://github.com/vuejs/rfcs/pull/3
+- Implementation PR: (leave this empty)
+
+# Summary
+
+Support dynamic values in directive arguments.
+
+# Basic example
+
+``` html
+<div v-bind:[key]="value"></div>
+<div v-on:[event]="handler"></div>
+```
+
+# Motivation
+
+Due to directive arguments being static, currently users would have to resort to argument-less object bindings in order to leverage dynamic keys:
+
+``` html
+<div v-bind="{ [key]: value }"></div>
+<div v-on="{ [event]: handler }"></div>
+```
+
+However, this has a few issues:
+
+- It's a lesser known technique that relies on knowledge of the object-based syntax for `v-bind`/`v-on` and the existence of computed property keys in JavaScript.
+
+- It generates less efficient code: an ad-hoc object is allocated and if there are other static bindings on the same element, it has to be dynamically iterated and mixed into the existing data object. The code looks roughly like this:
+
+  ``` js
+  return h('div', {
+    on: Object.assign({
+      click: onClick
+    }, {
+      [event]: handler
+    })
+  })
+  ```
+
+  Where as with dynamic arguments we can directly generate:
+
+  ``` js
+  return h('div', {
+    on: {
+      click: onClick,
+      [event]: handler
+    }
+  })
+  ```
+
+In addition, `v-slot` doesn't have an equivalent object syntax, since it's value is used for declaring the slot scope variable. So without the dynamic argument, the `v-slot` syntax will not be able to support dynamic slot names. Although this is probably a very rare use case, it would be a pain having to rewrite an entire template into render function just because of this single limitation.
+
+# Detailed design
+
+``` html
+<!-- v-bind with dynamic key -->
+<div v-bind:[key]="value"></div>
+
+<!-- v-bind shorthand with dynamic key -->
+<div :[key]="value"></div>
+
+<!-- v-on with dynamic event -->
+<div v-on:[event]="handler"></div>
+
+<!-- v-on shorthand with dynamic event -->
+<div @[event]="handler"></div>
+
+<!-- v-slot with dynamic name -->
+<foo>
+  <template v-slot:[name]>
+    Hello
+  </template>
+</foo>
+
+<!-- v-slot shorthand with dynamic name -->
+<!-- pending #3 -->
+<foo>
+  <template #[name]>
+    Default slot
+  </template>
+</foo>
+```
+
+# Drawbacks
+
+- Theoretically this opens up the directive argument to arbitrarily complex JavaScript expressions, but html attribute names cannot contain spaces and quotes, so in some cases the user may get tripped up with something like:
+
+  ``` html
+  <div :[key + 'foo']="value"></div>
+  ```
+
+  Which does not work as expected. A workaround would be:
+
+  ``` html
+  <div :[`key${foo}`]="value"></div>
+  ```
+
+  That said, complex dynamic key bindings should probably be pre-transformed in JavaScript via a computed property.
+
+# Alternatives
+
+N/A
+
+# Adoption strategy
+
+This is non-breaking and should be straightforward to introduce with appropriate documentation updates.
+
+# Unresolved questions
+
+What kind of expressions should be allowed


henriqemalheiros:
How it would look to custom directives? Would binding.oldArg be necessary?

yyx990803:
@henriqemalheiros custom directive would receive the evaluated argument I think - which requires no change in implementation. Is there a valid use case where the directive would want access to the raw argument expression?

henriqemalheiros:
@yyx990803 not the raw argument expression. I meant, if the argument is updated, how will the custom directive know about it?

Take for instance this overly simple directive that shows an alert:

```js
Vue.directive('alert', {
  bind(el, { arg, value }) {
    if (![ 'click', 'mouseover' ].includes(arg)) {
      throw new Error('incorrect directive argument')
    }

    el.__alertValue = value

    el.__alertCallback = () => {
      window.alert(el.__alertValue)
    };

    el.addEventListener(arg, el.__alertCallback)
  },
  update(el, { value, oldValue }) {
    if (value !== oldValue) {
      el.__alertValue = value
    }
  }
})
```

If the user changes the argument from `click` to `mouseover`, the directive should remove the old event listener and add a new one, hence my question about the `binding.oldArg`.

znck:
> Is there a valid use case where the directive would want access to the raw argument expression?

I think above is a use case to access raw argument expression, that way a directive can choose to throw an error if a dynamic argument is used. Or `dynamic: boolean` should be added to disable dynamic arguments.

posva:
Wouldn't dynamic ones unbind and bind again?

yyx990803:
@henriqemalheiros I see what you mean now, yes, dynamic argument does introduce an extra change point to watch for in custom directives. I'll update the RFC to reflect that.

laander:
In general, I like it. It's quite concise and applies to different directives with the same syntax.

One thing worth mentioning is that people coming from Angular might be thrown off by this, as `<div [foo]="value"> ... </div>` in Angular is equivalent to `<div :foo="value">` in Vue.

One alternative could be using `{ }` instead, like so:
```
<div v-bind:{key}="value"></div>
<div v-on:{event}="handler"></div>
```

The biggest benefit is that single `{ foo }` in JSX denotes javascript expressions i.e. dynamic evaluation. This also aligns nicely with the double `{{ foo }}` that you usually see in most other templating languages for expressions, including Vue.

With that said, do we know how to achieve dynamic attribute names in other framework that could be used as a mental reference? I can't seem to find anything for neither Angular, Aurelia nor Ember.

chrisvfritz:
I like the proposal. 🙂 I don't have any strong opinions on what kinds of expressions should be allowed, but it might be useful to only _officially_ support simple property expressions (like what `v-model`'s value allows). That would simplify documentation and also encourage simpler templates.

@laander I agree it's good to consider people coming from other frameworks, but I think it's better to be consistent with JavaScript in general, where `[xxx]` is used for dynamic keys in an object (e.g. `{ [dynamicKey]: value }`).

Justineo:
One potential downside is that this feature may lead to the assumption that any JavaScript expression can be used inside the `[]`. People may try to write `v-slot:[item + index]` which will break the whole attribute. On the other hand, the old `:slot="item + index" solved this nicely. But anyway this usage may be not so significant to be considered.

laander:
@chrisvfritz That's the winning argument for `[ ]`, agreed 👍

@Justineo Good point, but perhaps that's a saviour in disguise: inline interpolation of slot name should as a minimum be discouraged (use computed properties instead) and even better if the compiler won't accept it as valid. IMO opinionated decisions on fringe use cases like this are where Vue shines.

active-rfcs/0000-dynamic-directive-arguments.md
```diff
@@ -0,0 +1,112 @@
...
+- Theoretically this opens up the directive argument to arbitrarily complex JavaScript expressions, but html attribute names cannot contain spaces and quotes, so in some cases the user may get tripped up with something like:
+
+  ``` html
+  <div :[key + 'foo']="value"></div>
+  ```
+
+  Which does not work as expected. A workaround would be:
+
+  ``` html
+  <div :[`key${foo}`]="value"></div>
```
smolinari:
```suggestion
  <div :[`${key}foo`]="value"></div>
```
Scott

smolinari:
Just so I am certain I do understand, is this valid?

```html
<!-- v-slot with dynamic name -->
<foo>
  <template v-slot:[object.propName]>
    Hello
  </template>
</foo>
```

Scott

jacekkarczmarczyk:
What would be the behaviour of
```vue
<div v-on:[event]="handler"></div>
```
with null/undefined value of `event`? Maybe it should ignore the handler to make conditional binding of events simpler than it's today:

```vue
<template>
  <div v-on:[conditionalClick?'click':null]="handler"></div>
</template>
```

or 

```vue
<template>
  <div v-on:[conditionalClick]="hander"></div>
</template>

<script>
export default {
  computed: {
    conditionalClick () { 
      return isMonday() ? 'click' : null
    }
  }
}
</script>
```

Of course that could apply also to `v-bind` directive

yyx990803:
@smolinari yes, that's valid. As long as the expression does not contain characters that causes the HTML parser to end parsing of an attribute name - i.e. space and quotes.

yyx990803:
@jacekkarczmarczyk that's an interesting point. I'm inclined to allow falsy values to indicate that the binding should be ignored/removed. The concern would be it silently lead to undesired behavior when the falsy value is passed mistakenly.

yyx990803:
@jacekkarczmarczyk on a second thought, unfortunately if we generate the code like the following:

``` js
on: {
  [null]: handler
}
```

`null` would have already been converted to the string `"null"` key in the resulting object, so Vue would not be able to tell whether the user intentionally provided a `null` value or a `"null"` string.

Generating extra code to check the dynamic value type will resulting in less efficient code, but maybe it's worth it since dynamic argument is probably not a very commonly used feature.

yyx990803:
We pretty much have 3 options regarding falsy/non-string values:

1. Silently convert to strings (vote with 🚀)
2. Throw warnings for anything not strings (vote with 🎉)
3. Treat `false` and `null` as special values (remove binding), and warn the rest. (vote with ❤️)

jacekkarczmarczyk:
@yyx990803 Re: 3 - Wouldn't `undefined` be more reasonable, or did you mean null/undefined?

yyx990803:
@jacekkarczmarczyk `undefined` is much more likely to be a mistake.

Justineo:
I just tend to put DSL syntax into the attribute name while leaving JavaScript expressions with full expressive power to the attribute value. In that way we won't have to write space-less JavaScript code and It's also easier for tooling.

Is there any chance that we move the dynamic part into the attribute values? I've been thinking about this but haven't come up with something neat enough yet.

yyx990803:
@Justineo I am in general against adding artificial restrictions when we don't have technical reasons to do so. Watch expressions are limited because we don't want to implement a full expression parser - that's a tradeoff due to technical constraint. Dynamic arguments are just placed into generated code and handled by the JS engine. The only technical restriction here is attribute name characters.

yyx990803:
Updated: added section regarding [handling of `null` as a special value](https://github.com/vuejs/rfcs/blob/dynamic-directive-arguments/active-rfcs/0000-dynamic-directive-arguments.md#handling-of-null-as-special-value).

This RFC is now in final comment period.

jacekkarczmarczyk:
@yyx990803 "Unresolved questions" could be removed i guess

Justineo:
@smolinari It will be `<template v-slot:[name]>...</template>` as slot syntax will be updated as well.

smolinari:
@Justineo - Ok. But, `:slot` is possible now in oder to make the name dynamic. I'm just asking if that can/ will stay with the new proposal? 

Scott

Justineo:
As I understand, this proposal is exactly originated from the problem that dynamic slot names doesn't fit into our current directive syntax after we merged `slot` and `slot-scope` into `v-slot` ([ref](https://github.com/vuejs/rfcs/pull/2#issuecomment-454523404)).

smolinari:
True, but that has nothing to do with leaving `:slot` a possibility with this change. So, this:

<template :slot:namedSlot="someProps">

reads

Bind as namedSlot with someProps.

And....

`<template :slot[slotName]="someProps">`

reads

Bind as dynamic slotName with someProps.

I'll leave it at that. I don't want to muddy the discussion further. Just a yea or nay from Evan will be fine, maybe with a short explanation why it's a bad idea or not possible.   

Scott

yyx990803:
@smolinari the main problem is consistency: `:slot` is technically already the shorthand for `v-bind:slot`, so `slot` is the argument for `v-bind`. It doesn't seem to make sense to add an argument after the argument.

smolinari:
@yyx990803 - Thanks. Makes sense, Is there somewhere I can DM you an idea along these lines? 

Scott

yyx990803:
@smolinari why not make an RFC? The process is designed to help you think through rough ideas.

smolinari:
Nevermind. I did what you said Even and was writing down my thoughts in an issue here and then realized I couldn't come up with a decent example to make the suggestion worth while. 😁 

Scott

PR #6, opened 2019-01-21
Update filters only when inputs change

Demivan:
Instead of running filters on component re-render, only run them when inputs change.
[Full rendered proposal](https://github.com/Demivan/rfcs/blob/cached-filters/active-rfcs/0000-make-filters-cached.md)

### DIFF
--- a/active-rfcs/0000-make-filters-cached.md
+++ b/active-rfcs/0000-make-filters-cached.md
@@ -1,5 +1,5 @@
 - Start Date: 2019-01-21
-- Target Major Version: 3.x
+- Target Major Version: 2.x & 3.x
 - Reference Issues: https://github.com/vuejs/vue/issues/8880, https://github.com/vuejs/vue/issues/8497, https://github.com/vuejs/vue/issues/7556, https://github.com/vuejs/vue/issues/5682
 - Implementation PR: (leave this empty)
 
--- a/active-rfcs/0000-make-filters-cached.md
+++ b/active-rfcs/0000-make-filters-cached.md
@@ -1,0 +1,70 @@
+- Start Date: 2019-01-21
+- Target Major Version: 3.x
+- Reference Issues: https://github.com/vuejs/vue/issues/8880, https://github.com/vuejs/vue/issues/8497, https://github.com/vuejs/vue/issues/7556, https://github.com/vuejs/vue/issues/5682
+- Implementation PR: (leave this empty)
+
+# Summary
+
+Make filters work like computed properties - only update them when inputs change.
+
+# Motivation
+
+Copied from https://github.com/vuejs/vue/issues/8880:
+
+There are a lot of issues created about filters: https://github.com/vuejs/vue/issues/8497, https://github.com/vuejs/vue/issues/7304, https://github.com/vuejs/vue/issues/5920, https://github.com/vuejs/vue/issues/5682, https://github.com/vuejs/vue/issues/5109 etc.  
+Most of the issues are created because filters are executed multiple times.  
+For all of these issues answer is the same: for vdom comparison vue needs to run filter and then compares resulting dom tree - use computed properties instead.
+
+Because filters are evaluated on every re-render and not only when inputs change, right now there is no reason to use filters in mustache interpolations - they are the same as methods.  
+Angular filters for example are only evaluated when inputs change and this makes them much more usable.
+Filters are supposed to be pure functions anyway and I see no reason to not cache them (evaluate only when inputs change).
+
+
+Making filters cachable opens up possibilities for using them in mustache interpolations.
+Examples where cachable filters would be beneficial:
+* vue-i18n by @kazupon works around this issue by having `v-t` directive to cache result of the translation. 
+https://kazupon.github.io/vue-i18n/guide/directive.html#t-vs-v-t But is less flexible then mustache interpolation:
+It would be more readable and flexible instead of:
+```js
+<p v-t="'hello'"></p>
+```
+ to write
+```js
+<p>{{ "hello" | $t }}</p>
+```
+and it will be possible to do this:
+```js
+<p>{{ "hello" | $t }} {{ "world" | $t }}</p>
+```
+* Use Moment.js for date formatting. Moment.js is slow and because filters are not cached it is not really usable, especially when rendering lists.
+* All other places for formatting text, dates etc. 
+* Creating computed property for each formatted value (text, date etc) is just a lot of work and unnecessary code to write.
+
+# Detailed design
+
+Idea how it can be implemented in Vue: Vue compiller can create computed property when it encounters filter 
+
+# Drawbacks
+
+Why should we *not* do this? Please consider:
+
+- not sure how to implement this funtionality and how long it would take
+- it is possible to implement memoization for filters in user space
+- docs need to be extended to tell people to not write statefull filters
+- it is already discouraged to write 
+
+There are tradeoffs to choosing any path. Attempt to identify them here.
+
+# Alternatives
+
+Alternative solution is to use different symbol (instead of pipe) to denote filters that are cachable (pure).
+
+# Adoption strategy
+
+If we implement this proposal, how will existing Vue developers adopt it? Is
+this a breaking change? Can we write a codemod? Can we provide a runtime adapter library for the original API it replaces? How will this affect other projects in the Vue ecosystem?
+
+# Unresolved questions
+
+Optional, but suggested for first drafts. What parts of the design are still
+TBD?
--- a/active-rfcs/0000-make-filters-cached.md
+++ b/active-rfcs/0000-make-filters-cached.md
@@ -61,7 +61,7 @@ This way it will not be a breaking change, but this will create another syntax f
 
 # Adoption strategy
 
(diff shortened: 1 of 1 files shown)


posva:
Filters can be cached with memoization (https://lodash.com/docs#memoize):

```js
cost translate = memoize((key, ...args) => translate(key, args))
```

yyx990803:
This requires built-in memoization for every single filter use - and it may very well not be worth it. See https://twitter.com/dan_abramov/status/1083897065263034368

Demivan:
Can't Vue compiler generate computed property (or something) similar when compiling filter in template?

Right now if you use filter for formatting you are just losing performance for no reason. You better off just creating computed properties for formatting, and this is just a lot of code:
```html
<script>
  import moment from 'moment'

  export default {
    props: {
      date: { type: Date, required: true },
    },
    computed: {
      dateFormatted () {
        return moment(this.date).format('lll')
      }
    }
  }
</script>

<template>
  <div>{{ dateFormatted }}</div>
</template>
```
I know that in most applications difference in performance is not that big.  
But why would anyone use filters inside mustache interpolation in their current state? For me it is just complication for Vue without any benefits.

yyx990803:
It's true, we are actually thinking of removing filters altogether in v3. (We already attempted to in v2, but some users were strongly attached to it)

JosephSilber:
> we are actually thinking of removing filters altogether in v3.

Please don't 😢 

We've already shed a bunch of tears on their partial removal in Vue 2.

---

> This requires built-in memoization for every single filter use - and it may very well not be worth it. See https://twitter.com/dan_abramov/status/1083897065263034368

1. I don't think Dan's reasoning there holds up in this context (see point 2 here). The reason you don't put `memo` around every one of your functions is because you're not constantly calling them over and over again with the same parameters, whereas with filters in a vdom you always do.

2. I'm not sure if Dan meant it this way, but another reason you don't add `memo` all over is because function memoization usually only memoizes based on shallow equality; so if nested items change, your memoized value wouldn't change, which is wrong.

    That applies to React, but not to Vue. If Vue would use memoization, it should do it by creating computed properties for them all, instead of just memoizing the computed value as is. Since the computed _will_ update when its dependencies change, this makes much more sense for Vue.

---

While on the topic of memoization/auto-generated computed functions: you know what else could really use this? Inline function calls.

If you currently make a function call in your template, that function is being called every time the vdom renders. This can quickly become very expensive.

The usual solution is to use a computed property, but that doesn't work if you have to pass stuff to it (such as when you're looping through a list). The only way to not incur the cost of recalculation is to actually create a separate component for each item in the list. This is both wasteful, and mainly: tedious.

If inline function calls could be wrapped in a memoized computed, that would solve this issue beautifully.

BTW, Knockout does this out of the box. Since they don't use a vdom, they must do that (otherwise the view wouldn't update).

Akryum:
I don't see the benefits over a standard computed property (which can even be reused).

JosephSilber:
@Akryum if you have a dozen values that you want to transform, they add up pretty quickly. Having to always create computed properties manually is quite tedious.

For example, I have a translation filter that translates a piece of text in the UI to the local language:

```html
<h1>{{ 'Flights' | t }}</h1>

<ol v-for="flight of flights">
    {{ 'Departs' | t }}: {{ flight.departs_at | datetime }}<br>
    {{ 'Arrives' | t }}: {{ flight.arrives_at | datetime }}<br>
</ol>

<button>{{ 'Edit' | t }}</button>
```

You can imagine a component having dozens of these throughout its HTML. Having to manually create a computed property for each one of those is both tedious and clutters up the component.

To me, an explicit computed property is only useful for when you use the same computed value in multiple places.

JosephSilber:
@Akryum if you have a dozen values that you want to transform, they add up pretty quickly. Having to always create computed properties manually is quite tedious.

For example, I have a translation filter that translates a piece of text in the UI to the local language:

```html
<h1>{{ 'Flights' | t }}</h1>

<ol>
    <li v-for="flight of flights">
        {{ 'Departs' | t }}: {{ flight.departs_at | datetime }}<br>
        {{ 'Arrives' | t }}: {{ flight.arrives_at | datetime }}<br>
    </li>
</ol>

<button>{{ 'Edit' | t }}</button>
```

You can imagine a component having dozens of these throughout its HTML. Having to manually create a computed property for each one of those is both tedious and clutters up the component.

To me, an explicit computed property is only useful for when you use the same computed value in multiple places.

Akryum:
```html
<h1>{{ t('Flights') }}</h1>

<ol>
    <li v-for="flight of flights">
        {{ t('Departs') }}: {{ datetime(flight.departs_at) }}<br>
        {{ t('Arrives') }}: {{ datetime(flight.arrives_at) }}
    </li>
</ol>

<button>{{ t('Edit') }}</button>
```

znck:
I think with pipeline operator (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Pipeline_operator) there’s no need for filters. If we add a constraint that filters (or expressions with pipeline operators) should be pure then they can be memoized easily.

yyx990803:
My major objection is listed in the drawbacks section:

- This can already be done in userland
- Not all filters will actually benefit from the caching (some filters are simply cheaper to re-compute than to cache and compare arguments) - this makes selective caching a better option.
- Caching all filters by default breaks existing usage of stateful filters
- Caching with only shallow compare is not enough to cover deep object mutations - doing deep comparisons makes the caching even more expensive.

I just don't see this being worth it.

yyx990803:
@JosephSilber please stay on course for what is originally proposed in the RFC. If you have relevant but different ideas, open a separate RFC.

victornpb:
What is a good use case for the current filter implementation (called on every render)?

When I first learned about it I automatically assumed it would be great for doing exactly this, a `t_` filter for translation. But on the first attempt it feel short, I couldn’t use it on directives like `<button v-tooltip=“‘hint’ | t_”>OK</button>` and I quickly realized the tons of recomputations so I just started to use a method instead and cached in the view life cycle.

Demivan:
> some filters are simply cheaper to re-compute than to cache and compare arguments
doing deep comparisons makes the caching even more expensive.

@yyx990803 I agree with you that memoization for all filters is a bad solution for this.
Maybe `caching` was a bad word usage on my part. I'm thinking more on making filters reactive like computed properties instead.

Akryum:
@Demivan It's already reactively re-calculated since the render itself is re-executed is a reactive data source is updated. In this case, computed properties whose data dependencies didn't change won't be re-computed precisely because the last result was cached.

Demivan:
@Akryum But `capitalize` here this will be re-executed even if `name` did not change and only `age` have changed.
```html
<p>{{ name | capitalize }} {{ age | formatAge }}</p>
```
I think filters should collect dependencies like computed properties and update only when those dependencies change.

Akryum:
Which means their result would then need to be cached so they are reused to next renders.

Demivan:
That is true but if you add memoization to `capitalize` you need to store all parameter combinations and results for them. Because filters can (and should) be reused you can not just store last value.
```html
<p>{{ name | capitalize }} {{ age | capitalize }}</p>
```
With computed properties only last value need to be cached.

Justineo:
I think I got @Demivan's point. Actually it's *caching* for each occurrence of the filter which only need to store the last output, instead of the *memoization* of a filter function which stores all of its input/output.

daniele-orlando:
@yyx990803 Just for curiosity:

> It's true, we are actually thinking of removing filters altogether in v3. (We already attempted to in v2, but some users were strongly attached to it)

why are you thinking of removing the filters? What are the advantages for the vue-compiler?
If you have any link to the discussion about it would be appreciated.

JosephSilber:
> At best, you could create a computed property for each key in v-for. However, what if a new item is added to array? If an item is removed, wouldn't it require avoiding memory leak?

This can easily be solved with a nested structure of `WeakMap`s. Here's some pseudo code:

```js
const memoizedFilterBank = {
    map: new WeakMap,
    getComputedFor(vm, filter, params) {
        const map = this.getMapFor(params);

        if (! map.has(filter)) {
            map.set(filter, Vue.createComputed(
                () => vm.$options.filters[filter](...params)
            ));
        }

        return map.get(filter)
    },
    getMapFor(params) {
        return params.reduce((map, param) => {
            if (!map.has(param)) {
                map.set(param, new WeakMap)
            }
            return map.get(param)
        }, this.map)
    }
};
```

leopiccionia:
> This can easily be solved with a nested structure of `WeakMap`s.

Correct me if I'm wrong, but aren't weak references non-polyfillable in ES5 and, therefore, a no-go in Vue 2.x (Vue 2.x-next apart)?

JosephSilber:
> aren't weak references non-polyfillable in ES5 

@leopiccionia people usually polyfill `WeakMap` by storing the related value on the actual object itself, under a custom random key.

Here's some pseudo code:

```js
class WeakMap {
    constructor() {
        this.key = '__WeakMapPolyfill' + someRandomString() + '__'
    }
    set(key, value) {
        Object.defineProperty(key, this.key, {
            enumerable: false,
            value
        })
    }
    get(key) {
        return key[this.key]
    }
}
```

It's obviously not exactly the same as a `WeakMap`, but it's close enough for most use-cases.

yyx990803:

PR #7, opened 2019-01-26
Fix variable in template string to be consistent with js expression example

petr001:
There is little inconsistency in examples in `Constraints on expressions` section


smolinari:
I made the same suggestion here: https://github.com/vuejs/rfcs/pull/4#pullrequestreview-194347207

Scott

PR #10, opened 2019-02-01
Different external and internal names for props

JosephSilber:
**[View fully rendered RFC](https://github.com/JosephSilber/rfcs/blob/prop-aliases/active-rfcs/0000-prop-aliases.md)**

---

Allow aliasing props, to have different external and internal names (some languages call these "argument labels"):

```js
props: {
    externalName: {
        as: 'internalName'
    }
}
```

### DIFF
--- a/active-rfcs/0000-prop-aliases.md
+++ b/active-rfcs/0000-prop-aliases.md
@@ -1,0 +1,195 @@
+- Start Date: 2019-02-01
+- Target Major Version: 2.x
+- Implementation PR: (leave this empty)
+
+# Summary
+
+Allow aliasing props, to have different external and internal names (some languages call these "argument labels"):
+
+```js
+props: {
+    externalName: {
+        as: 'internalName'
+    }
+}
+```
+
+The `as` nomenclature is of course open for bikeshedding.
+
+# Basic example
+
+In this example (which is a modified version of the one [in Vue's docs](https://vuejs.org/v2/guide/components-props.html)), the component has a `counter` prop as part of its external API. It's aliased as `initialCounter` within the component itself, to prevent clashing with its own `counter` data property:
+
+```js
+export default {
+    props: {
+        counter: {
+            as: 'initialCounter'
+        }
+    },
+    data () {
+        return {
+            counter: this.initialCounter
+        }
+    }
+}
+```
+
+Here's how the component would be consumed:
+
+```html
+<the-component :counter="5" />
+```
+
+# Motivation
+
+Since props are not to be mutated, it is recommended to define a local data property that uses the prop as its initial value.
+
+1. Here is one example from Vue's docs:
+
+    ```js
+    props: ['initialCounter'],
+    data() {
+        return {
+            counter: this.initialCounter
+        }
+    }
+    ```
+
+    This makes sense within the component itself, but it feels wrong for it to affect the external API:
+
+    ```html
+    <the-component :initialCounter="5" />
+    ```
+
+    Aliasing the prop locally (as outlined above) would allow the consumer to use the simpler `counter` name for the prop:
+
+    ```html
+    <the-component :counter="5" />
+    ```
+
+1. Here's another example from the docs:
+
+    ```js
+    props: ['size'],
+    computed: {
+        normalizedSize() {
+            return this.size.trim().toLowerCase()
+        }
+    }
+    ```
+
+    This works, but now you have to always refer to the sanitized size as `normalizedSize` within your component.
+
+    Using an alias, we can change it to:
+
+    ```js
+    props: {
+        size: {
+            as: 'rawSize'
+        }
+    },
+    computed: {
+        size() {
+            return this.rawSize.trim().toLowerCase()
+        }
+    }
+    ```
+
+    which would allow us to refer to the normalized size as just `size` within our component, without affecting the component's external API.
+
+## Prior art
+
+External argument labels is something that is supported in many different languages.
+
+Here's how they work [in Swift](https://docs.swift.org/swift-book/LanguageGuide/Functions.html):
+
+```swift
+func greet(person: String, from hometown: String) -> String {
+    return "Hello \(person)! Glad you could visit from \(hometown)."
+}
+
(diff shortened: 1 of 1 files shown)


posva:
Also at https://github.com/vuejs/vue/issues/7943

LinusBorg:
I like the idea, nice work!

One thing I'm concerned about (but not qualified to answer) is how this complicates typings for Typescript.

laander:
Nice. I think `alias` get's my vote, reads better without further explanation:
```
props: {
    ageOfPerson: {
        type: Number,
        alias: 'age'
    }
}
```

jonaskuske:
Hmm, if we call it `alias` it'd have to work like an actual alias in my opinion, meaning both names would work – the same way aliases in Vue Router don't replace the path of a route but, well, alias it.

So for a component `Greet.vue` with
```javascript
{
  props: {
    fullName: {
      type: String,
      alias: 'name',
  }
}
```
these would have to be identical:
`<Greet name="Eddy Example" />`
`<Greet full-name="Eddy Example" />`

If we don't want an alias but *"to have different external and internal names"* then `as` is more accurate.

<hr>

**TL;DR:** If it's an actual alias and both names are valid when passing props, call it `alias`. Otherwise `as` (or maybe something completely else) is more accurate.

chriscalo:
Might it be able to accept multiple aliases by providing an array?

``` js
{
  props: {
    fullName: {
      type: String,
      alias: ['name', 'fullname'],
  }
}
```

garygreen:
> Might it be able to accept multiple aliases by providing an array

I would say that's overcomplicating it, if you want to provide an extra alias, you can do that in your own data method.

gruhn:
For real aliases you can already utilise computed properties

```js
props: ['external'],

computed: {
  internal () {
    return this.external
  }
}
```
Sure this is a little bit verbose but it elegantly arises from Vues natural capabilities.

znck:
There is an API available for provide/inject feature to have different local name, I think we should use similar API for props. 

```
props: {
  localName: {
    from: ‘externalName’
  }
}
```

https://vuejs.org/v2/api/#provide-inject

skyrpex:
I like the idea... I usually find myself writing the code blow. Anyway, my concern is how such a feature would play along with the planned typing support of Vue 3.

```js
export default {
    props: {
        user: { type: Object },
    },
    data() {
        // Need to keep a local copy to allow changes.
        return { localUser: this.user };
    },
};
```

thedamon:
@JosephSilber 
I really like this. A lot. I think you could expand the RFC text to mention that when/if it can combined with computed instead of data, it also essentially enables the ability to do prop coercion and set prop defaults based on other props:

## Stateful prop defaults

```
{
  props: {
   link: String,
   linkType: {
     type: String,
     as: 'rawLinkType'
   }
  }
  computed: {
    linkType(){ return this.rawLinkType || this.link.startsWith('mailto:') ? 'email' : 'default' }
  }
}
```

## Prop coercion

{
  props: {
   title: {
     type: String,
     as: 'rawTitle'
   }
  }
  computed: {
    title(){ return titleCase(this.title) }
  }
}

KorHsien:
I think the examples list in the Motivation section of the RFC (as well as some use cases commented) are not good arguments for this proposal.
1. This example expresses its intent quite accurately, the prop name indicates that it's the initial value, which is different from the internal mutable value. I think making this distinction is preferable.
```js
props: ['initialCounter'],
data() {
  return {
    counter: this.initialCounter
  }
}
```
2. This example is a bit representative for a lot of arguments around the previously commented use cases of props coercion, sanitization or preprocessing in general.
```js
props: ['size'],
computed: {
  normalizedSize() {
    return this.size.trim().toLowerCase()
  }
}
```
With the introduction of Composition API, those use cases could be handled elegantly:
```js
props: ['size'],
setup(props) {
  const size = computed(() => {
    return props.size.trim().toLowerCase()
  })

  return {
    size
  }
}
```

<hr/>

That being said, I'm **not** entirely against this. The Swift example in the RFC regarding argument labels is what in the similar vein with this proposal.
Take that example into the context of Vue:
```js
props: {
  person: String,
  from: {
    as: 'hometown',
    type: String
  }
}
```
And take the example of the new `<teleport>` component (though it's not implemented with normal component options). Its target prop name is `to`, but it's not ideal to refer it internally as `to`.
For the sole purpose of renaming (or aliasing), we could do it in setup function:
```js
props: {
  to: {
    type: [String, Element]
  }
},
setup(props) {
  const target = toRef(props, 'to')

  // setting up

  return {
    target
  }
}
```
But it's not as expressive and cohesive as following IMO:
```js
props: {
  to: {
    as: 'target',
    type: [String, Element]
  }
},
setup(props) {
  // setting up

  return {
    //
  }
}
```

So, in summary:
1. IMHO it's a bit of a detour to use this feature to solve the problems of props preprocessing of sort.
2. This proposal can be useful for providing the feature of argument labels.

CyberAP:
Just to clarify: it's not possible to solve that with Composition, you'll get an error for using a property with the same name as in props.
----
I think this change complicates things a lot more than it actually helps a developer. TypeScript support would be poor for this and probably require manually typing your props. It would also be unclear what value is being used in context without looking at props declaration first. Additionally since you can't mutate props but probably are going to use that value as an initial value in data that alone introduces confusion: component takes a prop foo then internally mutates foo context value. It's much more clear when we have foo as a prop and currentFoo as an internal state. Lastly, it can be mitigated by wrapping internal state into an object, which I think has a well balanced developer experience.

yogeshgalav:
https://github.com/vuejs/rfcs/discussions/513
It might resolve this issue.

JosephSilber:
Now that we have [reactive props destructuring](https://vuejs.org/api/sfc-script-setup.html#reactive-props-destructure) with `defineProps`, [this](https://github.com/JosephSilber/rfcs/blob/prop-aliases/active-rfcs/0000-prop-aliases.md#basic-example) is a solved use-case 👍 

```js
const { initialCounter: counter } = defineProps(['initialCounter']);
```

PR #11, opened 2019-02-04
Event Forwarding

Aferz:
[Fully rendered proposal](https://github.com/Aferz/rfcs/blob/master/active-rfcs/0000-forwarding-events.md)

---

Syntactic sugar for emitting events through a deeply nested component tree without repeating too much boilerplate.

```html
<!-- Children -->
<div>
  <button @click="$emit('click', $event)"></button>
</div>

<!-- Parent -->
<div>
  <children-component @click.emit />
</div>

<!-- Grand Parent -->
<div>
  <parent-component @click="handleClick" />
</div>
```

### DIFF
--- a/active-rfcs/0000-forwarding-events.md
+++ b/active-rfcs/0000-forwarding-events.md
@@ -1,0 +1,163 @@
+- Start Date: 2019-02-03
+- Target Major Version: 2.x
+- Reference Issues: N/A
+- Implementation PR:
+
+# Summary
+
+Allow *re-emitting* (forwarding) an event that is not intended to be handled in the current component but we want to forward the children event behaviour to a grandparent component.
+
+This RFC helps reducing boilerplate for a common vue.js pattern.
+
+Introducing the new `.emit` event modifier. (Name opened to change/discussion)
+
+# Basic example
+
+This would be a good case where we could apply this new modifier:
+
+```html
+<!-- Children -->
+<div>
+  <button @click="$emit('click', $event)"></button>
+</div>
+
+<!-- Parent -->
+<div>
+  <children-component @click.emit />
+</div>
+
+<!-- Grand Parent -->
+<div>
+  <parent-component @click="handleClick" />
+</div>
+```
+
+In above example, `parent-component` does not want/need to handle or add extra behaviour to _click_ event, but it wants to _re-emit_ the event upwards to allow parent components to handle it.
+
+# More realistic example
+
+Let's imagine we are building a custom text input. If we want to allow to react to common input events (focus, blur, etc...) we have to handle all the desired events and just _re-emit_ all the events (For the sake of the example, we won't add any custom behaviour to them):
+
+```html
+<!-- MyInput.vue -->
+<div>
+  <input 
+    type="text"
+    @focus="$emit('focus', $event)"
+    @blur="$emit('blur', $event)"
+    ...
+    />
+</div>
+```
+
+With the `.emit` modifier this could be simplified as:
+
+```html
+<!-- MyInput.vue -->
+<div>
+  <input 
+    type="text"
+    @focus.emit
+    @blur.emit
+    ...
+    />
+</div>
+  
+<!-- Parent --> 
+<div>
+  <my-input
+    @focus="onFocus"
+    @blur="onBlur" />
+</div>
+```
+
+# Motivation
+
+The main goal of this feature is to reduce boilerplate when building deeply nested components in which you have to emit events upwards through several component layers.
+
+If we make this process easier, this could help to avoid less visual patterns like `provide/inject`, or a custom event bus.
+
+# Detailed design
+
+This RFC is meant to be _only_ syntax sugar for this pattern:
+
+```html
+<my-component @click.emit />
+
+<!-- Transpiles to... -->
+
+<my-component @click="$emit('click', $event)" />
+```
+
+The new modifier could be chained with the rest of the modifiers too:
+
+```html
+<!-- Without .emit -->
+<form @submit.prevent="$emit('submit', $event)"> ... </form>
+
+<!-- With .emit -->
+<form @submit.prevent.emit> ... </form>
+```
+
+```html
+<!-- Without .emit -->
+<button @click.stop="$emit('click', $event)"> ... </button>
+
+<!-- With .emit -->
+<button @click.stop.emit> ... </button>
+```
+
+#### Templating engine
+Currently, it is not allowed to create a handler without a value (`@click`). So the template compiler has to be modified in order to allow empty values when `.emit` modifier is present.
+
+Another thing to take into consideration is the opposite. I think it doesn't make sense to have a handler with the `.emit` modifier (`@click.emit="onClick"`). Maybe a warning indicating that `.emit` modifier is meant to avoid handling the event in the current component is enough.
+
+#### Render Function
+For this case, I'd say that `.emit` doesn't need to be implemented in any way because of the nature of the render function.
+
+I'd leave to the user to implement it as the docs say for `.stop` or `.prevent`.
+
+# Drawbacks
+
+##### Another modifier
+It's something "new" to learn and could confuse the newcomers.
+
+##### It's a less explicit way of emitting an event
+Although I think the `.emit` modifier is a readable way of re-emitting an event, I understand that anyone could see this as a way of obfuscating what is happening.
+
+##### The relation with the rest of modifiers
+Vue has more modifiers as _key modifiers_, _mouse modifiers_, _system modifiers_, etc... In some edge cases could be confusing because of the verbosity.
+
+```html
+<input @keyup.alt.67.emit />
+
+<!-- Translated to: -->
+
+<input @keyup.alt.67="$emit('keyup', $event') />"
+```
+
+```html
+<input @click.ctrl.exact.emit />
+
+<!-- Translated to: -->
+
+<input @click.ctrl.exact="$emit('keyup', $event') />"
+```
+
+# Alternatives
+
+If this RFC is rejected, the alternative is to keep things as is right now.
+
+# Adoption strategy
+
+The strategy is simple. This RFC only adds a feature on top of Vue, so it only needs to be documented and explained (Similar as v-model is explained).
+
+# Unresolved questions
+
+##### Alternative names of the modifier
+I think the word `emit` makes sense, especially in Vue ecosystem. It's a known word and it's a known behaviour. The word `emit` is directly related to the events, so putting it next to an event handler, makes clear what the component it's trying to do. (`@click.emit`)
+
+I've thought in different names for the modifier that could be a better option in terms of user engagement:
+
+1. `.forward`: This could be a better fit because of the meaning of the word. Furthermore, this word can be associated directly with proxies, which this feature tries to mimic.
+2. `.propagate`: Although I like this one, I consider it a bit verbose and I think people could confuse it with the concept of `event.stopPropagation()` (Which is exactly the opposite).
--- a/active-rfcs/0000-forwarding-events.md
+++ b/active-rfcs/0000-forwarding-events.md
@@ -108,21 +108,19 @@ The new modifier could be chained with the rest of the modifiers too:
 ```
 
 #### Templating engine
-Currently, it is not allowed to create a handler without a value (`@click`). So the template compiler has to be modified in order to allow empty values when `.emit` modifier is present.
-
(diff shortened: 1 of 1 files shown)


visualfanatic:
How about including an option to specify what should be emitted? By default it would be `$event`, but you could also use any other variable:
#### Before:
```vue
<div>
  <input
    type="checkbox"
    @change="$emit('change', $event.target.checked)"
  >
</div>
```

#### After:
```vue
<div>
  <input
    type="checkbox"
    @change.emit="$event.target.checked"
  >
</div>
```

panstromek:
@Aferz well, now it becomes more a rearrangement than shortcut...

In a similar note, what about forwarding as some different event?

Something like:
```html
  <children-component @click.emit:clack />

...
<parent-component @clack="handleClack" />
```

Also, in most cases you can just use `v-on="$listeners"` (which of course is not the same)

panstromek:
I would actually prefer to  have `v-on:click="$on.clack"` but nullable - if you write this now, it will throw if the `clack` is not defined.

maxim-usikov:
May be add some options for `$emit`?

```
<div>
  <input
    type="checkbox"
    @change="$emit('change', payload, options)"
  >
</div>
```

```
// options
{
  bubbling: true,
  once: true,
  ...
}

**Comment by rellect** [comment]
I think this encourages bad design, there are better patterns to deal with nested components communication.

**Comment by Aferz** [comment]
Could you please be more specific about those better patterns to deal with nested components?

**Comment by rellect** [comment]
The emit is fine for child-parent communication, but less for child-grandparents communication, which is what this RFC encourages to do.

Don't get me wrong, for some cases it's okay, but I don't think it should be encouraged because it gets very hard to follow as it goes deeper.

There are other ways to deal with components communication: props, provide/inject, event bus, vuex, portals...

Each pattern has different cons/pros so it all depends on the specific situation.

**Comment by backbone87** [comment]
@rellect i think you miss the point. Event forwarding isnt bad practice. From the grandparents pov the event is emitted by the parent, and that is also exactly what happens, because the parent has explicit control over what is emitted/forwarded. The thing which is considered bad practice is a child emitting to grandparents around the parents control, and this is a violation of hierarchy.

**Comment by Aferz** [comment]
@rellect I agree with @backbone87 that you may miss the point of my proposal.

I don't want to work around __grandparent/parent/child__ communication. In fact, what I proposed is a way to write less to encourage the recommended approach of component communication.

The patterns you proposed are, in fact, workarounds to communicate between deep components hierarchies:

* __provide/inject__ is even discouraged in the official Vue docs because it would be unclear from which component the data is coming. From docs:

> provide and inject are primarily provided for advanced plugin / component library use cases. It is NOT recommended to use them in generic application code.

* __event bus__ suffers from the same that __provide/inject__ problems but it's even worse because the injected data could come from a different component hierarchy. Vue already has a first class API for communication.

* __vuex__ and __portals__ looks like over-engineering for simple communication. Yes, you can do it, but IMO these libraries are designed for resolving another kind of problems.

**Comment by rellect** [comment]
I don't think event forwarding is bad practice, I used it countless of times, but to a certain level.
After 3 levels deep it's confusing and hard to follow.

I understand that this RFC is only to reduce boilerplate for a common pattern, but my concern is it will be considered "official" approach for communication and abused. (Especially by newcomers).

It's only my opinion that's all

**Comment by bandleader** [comment]
On a similar note, how about passing props down and events up in the same syntax? Something I've had to do when creating a custom input

Right now:
```html
<!--Child Input Component-->
<div>
    <input :value="value" @input="$emit('input', $event.target.value)" />
</div>
```
With this proposal:
```html
<!--Child Input Component-->
<div>
    <input :value="value" @input.emit />
</div>
```
How can do this better? Maybe
```html
<!--Child Input Component-->
<div>
    <input :value.pass />
</div>
```

In this way `:value.pass="x"` is akin to `:value.sync="x"`, except that `:value.sync="x"` traps the `@update:value` event and runs `x=$event`, whereas `:value.pass` instead passes that event up: `$emit('update:value', $event)`

See [here](https://github.com/vuejs/vue/issues/8403) for a similar proposal.

(Caveat with my example above: HTML inputs (and existing custom Vue components designed to be compatible with `v-model`) don't raise `@update:value`, but rather `@input`. I propose to make this 'just work' as above, but won't go in to detail here as to how. An alternate suggestion would be to have a directive `v-pass`, or `v-model.pass`, which keeps the `v-model` model of `:value`/`@input`. See the proposal linked above for details.)

lloydjatkinson:
I thought this was removed from Vue 2 (the feature existed in Vue 1) specifically because it’s accepted to be a bad practice and strongly couples components needlessly.

backbone87:
@lloydjatkinson Event bubbling is implicitly propagating an event along an entire path in a tree. This proposal isn’t event bubbling. It is syntactic sugar for event forwarding, which has clear boundaries on which elements receive an event and no such notion as „along an entire path of arbitrary length“ exists

Aferz:
@backbone87 That's correct. I couldn't explained It better.

Don't forget this RFC is just syntactic sugar for keeping your code DRY and don't loose readability. Futhermore, is entirely optional.

bandleader:
In case people don't understand the use case, I think the simplest one is a custom input. Let's say that in a few places in your app you have an `<input />`, and you realize that you want to add some fancy features: auto-completion, checking against a database, calendar, validation, whatever. So you replace the `<input />` with a new `<custom-input />` that you create, that contains a regular `<input />` inside its template and adds the advanced features. Now you still want to support events like `@click`,` @focus`, `@blur`, `@input`, `@change` etc. So you want to pass those events up from the inner `<input />` So with this proposal, you would do `<input @blur.emit @focus.emit />`, etc.

Also you probably want your custom input to support `v-model`, i.e. you want the value of the input to come from your prop `value` and updates to raise the `@input` event. So you would do `<input :value="value" @input.emit />`. Or, with the extension I described above, you would simply do `<input :value.pass />` or similar.

panstromek:
@bandleader this use case is even better covered with transparent wrapper pattern with `v-on="$listeners" v-bind="$attrs"`

TBH I don't think this feature is worth it, I don't even remember last time I had a use case for it, and in that case, just writing like 5-10 characters is not that big of a problem.

Akryum:
If we merge [this RFC](https://github.com/vuejs/rfcs/blob/attr-fallthrough/active-rfcs/0000-attr-fallthrough.md) for Vue 3, you'll just need `<input v-bind="$attrs">` for a transparent wrapper.

Aferz:
Just to clarify, the example of @bandleader is exactly that, an example.

This RFC doesn't pretend to handle the creation of transparent wrappers or HOCs. It aims to reduce the amount of **boilerplate code** you write when you prefer to explicitly define the events the component is gonna handle or simply emit upwards. 

This:

```html
<input 
  @focus="$emit('focus', $event)"
  @click="$emit('click', $event")
  @input="$emit('input', $event)" />
```
Becomes this: 

```html
<input 
  @focus.emit 
  @click.emit 
  @input.emit />
```
Sure, I could make use of `v-on`, but what if I just don't want to pass along _all_ the events attached from the parent? You'd have to pluck those events from `$listeners` before passing them to `v-on` directive.

Anyway, this is completely optional, and you could not use it at all if you don't like it, but I think this feature could improve readability in our own components because of the fact that we explicitly define which events we want to handle instead of relying on `$listeners` container.

Cheers.

bandleader:
@Aferz 👍 

@panstromek Custom inputs aren't necessarily made up of just one child input. Think of a date control for instance: it's a <span> that contains 3 children: a textbox, an 'open calendar' button, and a hidden div with the actual calendar. Maybe `@focus` and `@blur` should be passed up from the textbox, but `@click` from any of the controls, and the mouse events should probably passed up from the parent <span>...

panstromek:
@bandleader I don't deny that there is a use case for this. I just don't think it's worth implementing in the compiler. Every change like that comes with a maintanence burden for core team so it better be worth it - ie. used by many and saving a lot of work. This is a syntax sugar for very narrow use case, saving just a few keystrokes in a very specific case. Not worth it IMO.

panstromek:
Another point is that you could implement your own directive for this. Something like
```
<input v-pass:input>
``` 

or even better with modifiers 

```vue
<input v-pass.input.blur.click >
```

bandleader:
@panstromek I like that, but don't forget checking if the node is a Vue component, because then you need to run `$on` instead of `addEventListener`. Also, if it's an HTML input, you have to extract out `e.target.value`, otherwise it won't work with v-model, since the parent node is a custom component. 

Also, the scenario I mentioned earlier `<input v-bothways:value />` which would pass `value` down and also pass up `update:value` probably can't be done at all through a directive, because you can't set v-bind manually at runtime, right?
```js
 Vue.directive('bothways', {
    bind: (el, binding, vnode) => {
      el.addEventListener("update:" + binding.arg, e => {
        vnode.context.$emit("update:" + binding.arg, e)
      })
      // But how do we pass the prop down?
      vnode.addVBind(binding.arg, () => binding.value) //??
      // No such syntax, obviously!
    }
  })
```


Also, we'd have to re-impeeother "magic" of v-model (there is quite a bit).

bandleader:
@panstromek I actually got a bit further in implementing it as a directive, by skipping the binding and simply manually updating `el.value` on `updated()`. [Fiddle here](https://jsfiddle.net/Lvbwjmec/5/)

But it's going to be really hard to re-implement everything that's in v-model and v-bind. That's why built-in directives would be helpful, or else access to some of the v-model and v-bind internals via an API...

yyx990803:
In Vue 3, this will no longer be an issue as all listeners are now part of attrs and will be automatically inherited if the child component returns another component as root.

JosephSilber:
@yyx990803 it very frequently isn't the root.

yyx990803:
@JosephSilber in that case you can use `inhertiAttrs: false` with explicit `v-bind="$attrs"`.

Aferz:
Sorry @yyx990803 but I don't get it. Could you explain with a code example? Because what you said is already available in Vue2 and I don't see how it fixes what this RFC proposes. Thanks!

yyx990803:
More details can be found in the new attrs fallthrough behavior RFC: https://github.com/vuejs/rfcs/pull/92

I assume the use case mostly happens with "wrapper components" where you not only want to forward a specific event, but rather anything that has not been declared as a prop by a component in the hierarchy. Therefore this is should cover a lot of such cases:

``` html
<!-- Child that wraps a non-root target element: use inheritAttrs: false with explicit v-bind="$attrs" -->
<div>
  <input v-bind="$attrs">
</div>

<!-- If the Parent renders Child as root: nothing needed, all attrs implicitly falls through -->
<children-component />

<!-- Grand Parent -->
<div>
  <parent-component @click="handleClick" />
</div>
```

The `@click` handler is compiled into an `onClick` prop and gets passed all the way to the final `<input>` element.

Aferz:
Understood.

Thanks @yyx990803 !

bandleader:
@yyx990803  What about the case where you want to forward specific events? For example: a component that provides a form control with some extra elements (validation feedback, calendar icon, label, etc.) before/after the `<input />`.  You probably want to pass up `input`, `focus` and `blur` from the <input />, whereas the other events (and attrs) would be on your whole root element (which doesn't have input/focus/blur).

```html
<div>
  <label>
    <input :value="value" @input.emit @focus.emit @blur,emit />
  </label>
</div>
```

yyx990803:
@bandleader you can do something like

``` html
<div v-bind="excludeEvents($attrs)">
  <label>
    <input :value="value" v-bind="pluckEvents($attrs)" />
  </label>
</div>
```

Where you include/exclude the events you want in the method. JavaScript provides full flexibility in these niche cases without the need for introducing additional template syntax.

bandleader:
@yyx990803 Fair enough. In a similar vein, how would you recommend doing what I mentioned in [this comment above](#issuecomment-527250041) -- passing an event up and a prop down in a single syntax? Can it be done similarly: 
```html
<div>
  <input v-bind="twoWayValue" />
</div>
```

```js
computed: {
  twoWayValue: function() {
    return { value: this.value, oninput: e => this.$emit('input', e.target.value) }
  }
}
```

Or via a helper:
```js
// Helper function: returns a computed property function that you can v-bind to your element.
function twoWayHelper(propName) {  
  return function() {
    const eventName = propName === "value" ? "input" : propName
    const eventHandler = e => this.$emit(eventName, e instanceof InputEvent ? e.target.value : e)
    return { [propName]: this[propName}, ['on' + eventName]: eventHandler }
  }
}

```
```html
<div>
   <input v-bind="valueBinding" />
</div>
```
```js
computed: {
  valueBinding: twoWayHelper("value")
}
```

Would this work?

yyx990803:
@bandleader yeah that should work.

Aferz:
@bandleader that looks very clever. Thanks for sharing.

PR #18, opened 2019-03-01
v-bind .prop shorthand proposal

yyx990803:
``` html
<!-- full syntax -->
<video v-bind:muted.prop="isMuted"></video>

<!-- with shorthand -->
<video .muted="isMuted"></video>
```

[Rendered](https://github.com/vuejs/rfcs/blob/v-bind-prop-shorthand/active-rfcs/0000-v-bind-prop-shorthand.md)

### DIFF
--- a/active-rfcs/0000-v-bind-prop-shorthand.md
+++ b/active-rfcs/0000-v-bind-prop-shorthand.md
@@ -1,0 +1,65 @@
+- Start Date: 2019-03-01
+- Target Major Version: 2.x
+- Reference Issues: https://github.com/vuejs/vue/issues/7582
+- Implementation PR: [commit](https://github.com/vuejs/vue/commit/d2902ca8ec5fd184fe81479fea1318553fdb8323) (merged but hidden behind feature flag)
+
+# Summary
+
+Introduce a new shorthand for `v-bind` with `.prop` modifier (binding as DOM properties instead of attributes).
+
+# Basic example
+
+``` html
+<!-- current syntax -->
+<video v-bind:muted.prop="isMuted"></video>
+
+<!-- with shorthand -->
+<video :muted.prop="isMuted"></video>
+
+<!-- with proposed shorthand -->
+<video .muted="isMuted"></video>
+```
+
+# Motivation
+
+Provide more succinct syntax when binding directly to DOM properties. This is not a very common use case
+
+# Detailed design
+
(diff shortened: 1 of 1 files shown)


donnysim:
Would `.innerHTML` and` .textContent` still work when used in html? does html case insensitive not apply as in `.innerHTML` would become `.innerhtml`?

jacekkarczmarczyk:
.inner-html - it's mentioned in the rfc

donnysim:
So `.inner-html` will be compiler specific? otherwise how it will know if it should be `innerHtml` or `innerHTML`?

troy351:
@donnysim AFAIK, there are only 2 **NOT** read-only props that not strictly follow the camel-case rules. `innerHTML` and `outerHTML`.

donnysim:
@troy351 I know, that's just a question on implementation 🤷

troy351:
@donnysim I think just add additional check for these two props as an exception

leopiccionia:
I'm favorable to the shorthand. It remembers me [lit-html](https://lit-html.polymer-project.org/guide/writing-templates#bind-to-properties).

I'm neutral to `v-html` and `v-text` deprecation.

chrisvfritz:
I like the deprecation of v-text and v-html to simplify the API. As an alternative to a shorthand though, I wonder if it might be better to use the same behavior as Svelte, assigning to a DOM property if one exists and falling back to an attribute otherwise. Then we'd achieve as brief a syntax as this shorthand while further simplifying the API to deprecate the `.prop` modifier, instead of adding to the API with difficult-to-google syntax.

I can't immediately think of any edge cases and couldn't find any open issues for Svelte reporting any. Is this potentially feasible?

yyx990803:
@chrisvfritz I think that's worth trying, I'm mostly concerned with potential behavior differences from how Vue handles it now. Would need some testing in real world use cases.

chrisvfritz:
@yyx990803 Perhaps as an interim step, detected DOM properties could not only have the DOM property set, but also get added as an attribute, as happens now. That way, if someone is relying on the attribute existing (e.g. using a `ref` and calling `getAttribute`), it will still exist. I agree though that either way, we'd need some real-world testing to ensure this doesn't create edge cases we can't currently foresee.

yyx990803:
Closing since this is technically no longer needed in Vue 3 - if a property with the same name exists on the target DOM node, it will automatically be bound as a prop.

PR #20, opened 2019-03-12
Slots Unification

yyx990803:
[Rendered](https://github.com/vuejs/rfcs/blob/slots-unification/active-rfcs/0000-slots-unification.md)


posva:
Is

```js
h(Comp, () => [h('div', this.msg)])
```

equivalent to

```js
h(Comp, {
  slots: {
    default: () => [h('div', this.msg)],
  },
})
```

in v3?

yyx990803:
@posva yes.

YunSun-CN:
Instead of using keyword `$stable`, why not design `slot` functions as `computed` to cache vnodes?Developers should be noted in documentation that a `slot` function should be defined preserved outside the `render` function scope.

yyx990803:

PR #21, opened 2019-03-22
Transparent v-model

backbone87:
[Rendered](https://github.com/vuejs/rfcs/blob/6fb672c452bd45ccd390778d1372dfd10bc906bc/active-rfcs/0000-transparent-v-model.md)

```js
const child = {
  template: `<input type="text" v-model="value">`,
  props: {
    value: String,
  },
};
const parent = {
  components: { child },
  template: `<child v-model="value">`,
  data() {
    return { value: '' };
  },
};
```

### DIFF
--- a/active-rfcs/0000-transparent-v-model.md
+++ b/active-rfcs/0000-transparent-v-model.md
@@ -1,0 +1,222 @@
+- Start Date: 2019-03-22
+- Target Major Version: 2.x
+- Reference Issues: https://github.com/vuejs/vue/issues/9749 https://github.com/vuejs/vue/issues/8403
+- Implementation PR:
+
+# Summary
+
+Transparently use the `v-model` directive on data **and props**.
+
+# Basic example
+
+## Vue 2.6
+
+```js
+const child = {
+  template: `<input type="text" :value="value" @input="$emit('input', event.target.value)">`,
+  props: {
+    value: String,
+  },
+};
+const parent = {
+  components: { child },
+  template: `<child v-model="value">`,
+  data() {
+    return { value: '' };
+  },
+};
+```
+
+## Proposed behavior
+
+```js
+const child = {
+  template: `<input type="text" v-model="value">`,
+  props: {
+    value: String,
+  },
+};
+const parent = {
+  components: { child },
+  template: `<child v-model="value">`,
+  data() {
+    return { value: '' };
+  },
+};
+```
+
+
+# Motivation
+
+Wrapping model exposing components is a rather cumbersome and error prone task,
+which involves writing boilerplate code. This gets even worse when dealing with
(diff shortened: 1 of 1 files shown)


posva:
I don't think this is a good idea, mostly because of the implicit behaviour. But even if a modifier is added, there is still the problem of detecting what event to emit and if it has to be emitted by looking at the component's parent.

I would rather see something on the prop itself than v-model, so that if the prop is assigned to a new value, it emits an event. This also intersects with https://github.com/vuejs/rfcs/pull/10 which is in part also addressing the same issue and could be solved by adding a plugin that reads an option from props like `emitOnAssign`, creating an internal prop with a setter that emits an event

backbone87:
> the problem of detecting what event to emit

The current v-model does this lookup?

I can do:
```js
const child = {
  template: `<input type="text" :value="value" @input="$emit($options.model.event, $event.target.value)">`,
  model: { prop: 'model', event: 'updateModel' },
  props: ['model'],
  computed: {
    value() {
      return this[this.$options.model.prop];
    },
  },
};
const parent = {
  components: { child },
  template: `<child v-model="foo">`,
  data() {
    return { foo: 'bar' };
  },
};
```

posva:
You are right, nvm that

backbone87:
> I don't think this is a good idea, mostly because of the implicit behaviour.

What do you mean by implicit behavior? Wrapping components is a common use-case and so is wrapping model-exposing components. The boilerplate to wrap model-exposing components is even used in the vue guide https://vuejs.org/v2/guide/components-custom-events.html#Customizing-Component-v-model. And then you only have the basic handling, which does not respect keyed models or dynamic components.

posva:
implicit as being something that happens without user consent, passing a prop to v-model could be a mistake on the dev's end while they meant to use `:value="` instead

backbone87:
Thats the tradeoff for expressiveness, isnt it?

We could make it explicit:
```js
const child = {
  model: {
    // ...
    transparent: true,
  },
  // ...
};
```

backbone87:
To still have an indication for data/prop mixups, we could provide a dev runtime notice, in case there is no listener for the update event in the parent (and therefore the bound model would never update anywhere).

agronick:
To build on this how about an option for forwarding events? Something like `forwards: ['change', 'input', 'hover']` in the js instead of `@change="$emit('change', $event) @input="$emit('input', $event) @hover="$emit('hover', $event)` in the template. Sometimes you end up writing code like that 3 components deep just to get events where they need to be.

gustojs:
Forwarding events can be dealt with by the developer with a reusable function, doesn't have to be solved on the API level.

yyx990803:

PR #22, opened 2019-03-26
Advanced Reactivity API

yyx990803:
[Rendered](https://github.com/vuejs/rfcs/blob/advanced-reactivity-api/active-rfcs/0000-advanced-reactivity-api.md)

API for creating and observing standalone reactive values outside components.

``` js
import { state, value, computed, watch } from '@vue/observer'

// reactive object
// equivalent of 2.x Vue.observable()
const obj = state({ a: 1 })

// watch with a getter function
watch(() => obj.a, value => {
  console.log(`obj.a is: ${value}`)
})

// a "ref" object that has a .value property
const count = value(0)

// computed "ref" with a read-only .value property
const plusOne = computed(() => count.value + 1)

// refs can be watched directly
watch(count, (count, oldCount) => {
  console.log(`count is: ${count}`)
})

watch(plusOne, countPlusOne => {
  console.log(`count plus one is: ${countPlusOne}`)
})
```

### DIFF
--- a/active-rfcs/0000-advanced-reactivity-api.md
+++ b/active-rfcs/0000-advanced-reactivity-api.md
@@ -277,7 +277,7 @@ export default {
 
 ## Beyond the API
 
-The APIs proposed here are just low-level building blocks. Technically, they provide everything we need for global state management, so Vuex can be rewritten as a very thin layer on top of these APIs. In addition, when combined with [the ability to programmatically hook into the component lifecycle](TODO), we can offer a logic reuse mechanism with capabilities similar to React hooks.
+The APIs proposed here are just low-level building blocks. Technically, they provide everything we need for global state management, so Vuex can be rewritten as a very thin layer on top of these APIs. In addition, when combined with [the ability to programmatically hook into the component lifecycle](https://github.com/vuejs/rfcs/pull/23), we can offer a logic reuse mechanism with capabilities similar to React hooks.
 
 # Drawbacks
 
--- a/active-rfcs/0000-advanced-reactivity-api.md
+++ b/active-rfcs/0000-advanced-reactivity-api.md
@@ -1,0 +1,302 @@
+- Start Date: 03-05-2019
+- Target Major Version: 2.x & 3.x
+- Reference Issues: N/A
+- Implementation PR: N/A
+
+# Summary
+
+Provide standalone APIs for creating and observing reactive state.
+
+# Basic example
+
+```js
+import { state, value, computed, watch } from '@vue/observer'
+
+// reactive object
+// equivalent of 2.x Vue.observable()
+const obj = state({ a: 1 })
+
+// watch with a getter function
+watch(() => obj.a, value => {
+  console.log(`obj.a is: ${value}`)
+})
+
+// a "ref" object that has a .value property
+const count = value(0)
+
+// computed "ref" with a read-only .value property
+const plusOne = computed(() => count.value + 1)
+
+// refs can be watched directly
+watch(count, (count, oldCount) => {
+  console.log(`count is: ${count}`)
+})
+
+watch(plusOne, countPlusOne => {
+  console.log(`count plus one is: ${countPlusOne}`)
+})
+```
+
+# Motivation
+
+## Decouple the reactivity system from component instances
+
+Vue's reactivity system powers a few aspects of Vue:
+
+- Tracking dependencies used during a component's render for automatic component re-render
+
+- Tracking dependencies of computed properties to only re-compute values when necessary
+
+- Expose `this.$watch` API for users to perform custom side effects in response to state changes
+
+Until 2.6, the reactivity system has largely been considered an internal implementation, and there is no dedicated API for creating / watching reactive state without doing it inside a component instance.
+
+However, such coupling isn't technically necessary. In 3.x we've already split the reactivity system into its own package (`@vue/observer`) with dedicated APIs, so it makes sense to also expose these APIs to enable more advanced use cases.
+
+With these APIs it becomes possible to encapsulate stateful logic and side effects without components involved. In addition, with proper ability to "connect" the created state back into component instances, they also unlock a powerful component logic reuse mechanism.
+
+# Detailed design
+
+## Reactive Objects
+
+In 2.6 we introduced the `observable` API for creating reactive objects. We've noticed the naming causes confusion for some users who are familiar with RxJS or reactive programming where the term "observable" is commonly used to denote event streams. So here we intend to rename it to simply `state`:
+
+``` js
+import { state } from 'vue'
+
+const object = state({
+  count: 0
+})
+```
+
+This works exactly like 2.6 `Vue.observable`. The returned object behaves just like a normal object, and when its properties are accessed in **reactive computations** (render functions, computed property getters and watcher getters), they are tracked as dependencies. Mutation to these properties will cause corresponding computations to re-run.
+
+## Value Refs
+
+The `state` API cannot be used for primitive values because:
+
+- Vue tracks dependencies by intercepting property accesses. Usage of primitive values in reactive computations cannot be tracked.
+
+- JavaScript values are not passed by reference. Passing a value directly means the receiving function will not be able to read the latest value when the original is mutated.
+
+The simple solution is wrapping the value in an object wrapper that can be passed around by reference. This is exactly what the `value` API does:
+
+``` js
+import { value } from 'vue'
+
+const countRef = value(0)
+```
+
+The `value` API creates a wrapper object for a value, called a **ref**. A ref is a reactive object with a single property: `.value`. The property points to the actual value being held and is writable:
+
+``` js
+// read the value
+console.log(countRef.value) // 0
+
+// mutate the value
+countRef.value++
+```
+
+Refs are primarily used for holding primitive values, but it can also hold any other values including deeply nested objects and arrays. Non-primitive values held inside a ref behave like normal reactive objects created via `state`.
+
+## Computed Refs
+
+In addition to plain value refs, we can also create computed refs:
+
+``` js
+import { value, computed } from 'vue'
+
+const count = value(0)
+const countPlusOne = computed(() => count.value + 1)
+
+console.log(countPlusOne.value) // 1
+count.value++
+console.log(countPlusOne.value) // 2
+```
+
+Computed refs are readonly by default - assigning to its `value` property will result in an error.
+
+Computed refs can be made writable by passing a write callback as the 2nd argument:
+
+``` js
+const writableRef = computed(
+  // read
+  () => count.value + 1,
+  // write
+  val => {
+    count.value = val - 1
+  }
+)
+```
+
+Computed refs behaves like computed properties in a component: it tracks its dependencies and only re-evaluates when dependencies have changed.
+
+## Watchers
+
+All `.value` access are reactive, and can be tracked with the standalone `watch` API.
+
+**NOTE: unlike 2.x, the `watch` API is immediate by default.**
+
+`watch` can be called with a single function. The function will be called immediately, and will be called again whenever dependencies change:
+
+``` js
+import { value, watch } from 'vue'
+
+const count = value(0)
+
+// watch and re-run the effect
+watch(() => {
+  console.log('count is: ', count.value)
+})
+// -> count is: 0
+
+count.value++
+// -> count is: 1
+```
+
+### Watch with a Getter
+
+When using a single function, any reactive properties accessed during its execution are tracked as dependencies. The **computation** and the **side effect** are performed together. To separate the two, we can pass two functions instead:
+
+``` js
+watch(
+  // 1st argument (the "computation", or getter) should return a value
+  () => count.value + 1,
+  // 2nd argument (the "effect", or callback) only fires when value returned
+  // from the getter changes
+  value => {
+    console.log('count + 1 is: ', value)
+  }
+)
+// -> count + 1 is: 1
+
+count.value++
+// -> count + 1 is: 2
+```
+
+### Watching Refs
+
+The 1st argument can also be a ref:
+
+``` js
+// double is a computed ref
+const double = computed(() => count.value * 2)
+
+// watch a ref directly
+watch(double, value => {
+  console.log('double the count is: ', value)
+})
+// -> double the count is: 0
+
+count.value++
+// -> double the count is: 2
+```
+
+### Stopping a Watcher
+
+A `watch` call returns a stop handle:
+
+``` js
+const stop = watch(...)
+
+// stop watching
+stop()
+```
+
+If `watch` is called inside lifecycle hooks or `data()` of a component instance, it will automatically be stopped when the associated component instance is unmounted:
+
+``` js
+export default {
+  created() {
+    // stopped automatically when the component unmounts
+    watch(() => this.id, id => {
+      // ...
+    })
+  }
+}
+```
+
+### Effect Cleanup
+
+The **effect** callback can also return a cleanup function which gets called every time when:
+
+- the watcher is about to re-run
+- the watcher is stopped
+
+``` js
+watch(idRef, id => {
+  const token = performAsyncOperation(id)
+
+  return () => {
+    // id has changed or watcher is stopped.
+    // invalidate previously pending async operation
+    token.cancel()
+  }
+})
+```
+
+### Non-Immediate Watchers
+
+To make watchers non-immediate like 2.x, pass additional options via the 3rd argument:
+
+``` js
+watch(
+  () => count.value + 1,
+  () => {
+    console.log(`count changed`)
+  },
+  { immediate: false }
+)
+```
+
+## Exposing Refs to Components
+
+While this proposal is focused on working with reactive state outside of components, such state should also be usable inside components as well.
+
+Refs can be returned in a component's `data()` function:
+
+``` js
+import { value } from 'vue'
+
+export default {
+  data() {
+    return {
+      count: value(0)
+    }
+  }
+}
+```
+
+**When a `ref` is returned as a root-level property in `data()`, it is bound to the component instance as a direct property.** This means there's no need to access the value via `.value` - the value can be accessed and mutated directly as `this.count`, and directly as `count` inside templates:
+
+``` html
+<div @click="count++">
+  {{ count }}
+</div>
+```
+
+## Beyond the API
+
+The APIs proposed here are just low-level building blocks. Technically, they provide everything we need for global state management, so Vuex can be rewritten as a very thin layer on top of these APIs. In addition, when combined with [the ability to programmatically hook into the component lifecycle](TODO), we can offer a logic reuse mechanism with capabilities similar to React hooks.
+
+# Drawbacks
+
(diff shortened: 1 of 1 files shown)


skyrpex:
The `watch` parameters difference between a reactive object and a ref object is a bit confusing. Maybe `state` and `value` could work seamlessly in some way?

Aaron-Pool:
I, personally, would like to see some heavy warning verbiage around many of these features, as well as some "don't use this in 'x' scenario, just use simpler option 'y'" examples. I think that would go a long way towards helping newer developers get a good grasp of when and when not to use this level of the api. A lot of these apis look like they could result in some real spaghetti architecture if perceived as "first-choice" candidates for state management. As someone who has experienced some angular.js pre-1.5.x trauma, I'd certainly like to avoid that.

crutchcorn:
If `$watch` was removed in favor of the import, would `computed` still rely on it under-the-hood?
If so, I'd be curious to hear how many applications use Vue without `watch` or computed props 

Otherwise, I think that removing `$watch` would be worth it for the tree shaking capabilities alone

LinusBorg:
@skyrpex 

What in particular do you find confusing? I assume it's that you for a`reactive object you would do something like `() => ob.a` (accessing a property) while you can simply pass return a ref without having to do `ref.value`?

I see how this could trip up some people, and in fact, `() => ref.value` would work just as well. 

But since we can recognize that `ref` is a ref object, we can infer that the only sensible thing to do is to watch it's `value` prop, which allows us to provide users the shortcut of just passing the ref directly.

We can't do the same for "normal" reactive objects, so in their case, we have to provide a function to actually access the property(s)  to watch.

Do you think it would be fine if we document the `getter function` way for both first, and then explicitly document directly passing a ref as a shortcut?

LinusBorg:
@Aaron-Pool We definitely agree. We called this RFC **Advanced** Reactivity API for that reason, basically.

The plan would be to thoroughly extend the `Reactivity in depth` chapter that we have in our current docs with documentation for these advanced APIs as well as useage guidelines.

LinusBorg:
@crutchcorn 

`$watch` would be a wrapper around the standalone `watch()`, and computed properties would also internally rely on `watch()`, so removing the component APIs `$watch` wouldn't break or change anything in that regard.

thedamon:
I don't have much experience with them, but this looks like it could make writing render functions feel a lot more 'native' to Vue. (currently it feels a little bit like its own language). Is that an accurate assessment?

I have a niggling concern about all this `refs` potentially becoming confusing in relation to `this.$refs`.. might it make sense to rename the latter to `this.$domRefs` (and potentially keep the old one as a deprecated alias for backward compatability?) (this potential confusion would likely predate this particular proposal, though)

skyrpex:
> What in particular do you find confusing?

Exactly what you mentioned (sorry, I was on mobile so I couldn't expand much). Basically, this:

```js
const count = value(0)
watch(count, (countVal) => {
  // countVal === count.value
})
```

> Do you think it would be fine if we document the getter function way for both first, and then explicitly document directly passing a ref as a shortcut?

Yes, I think so!

LinusBorg:
@thedaemon

This particular API won't have much of an influence on how we write render functions, no. 

But we'll have a separate RFC about render functions and the virtual dom soon. 

About the ref <-> $ref thing: i raised the same concern in internal discussions. We will certainly have to rename one or the other i think. But that's secondary i think.

dsonet:
I think since the `value` actually is an exception of `state`, so doesn't have to introduce this `value` ref API, wrap the primitive value by user when really needed(since mostly we just use object anyway in a real world). Then we only need to know reactive object, make things much easier.

CyberAP:
Do `state` and `value` really must remain as two separate functions? I can imagine getting confused when to use which. 
Maybe it could work like this?

```javascript
const obj = state({ count: 1 });
computed(() => obj.count + 1);

const count = state(1);
computed(() => count.value + 1);
```

Also they don't declare action in their name. Maybe `setState` or `useState` would be better (but I understand that they might get confused with appropriate React functions).

CyberAP:
What do you think about pointers having custom `valueOf` method, so it's possible just to use the return of the `value` function?

```javascript
function value(initialValue) {
  return {
    value: initialValue,
    valueOf() {
      return this.value;
    },
  }
}

const count = value(1);
computed(() => count + 1);
```
That way it could interoperate with the `state` function.

Aferz:
I'm really excited about this new API. I've already tried my best attempting to implement it un user land, but It was a bit hacky (https://github.com/vuejs/vue/issues/9509).

I hope this gets approved/implemented/merged as soon as possible.

Thank you for this, vue team!!

thecrypticace:
In general, I love the overall API, it is very well thought out and designed.

I do have some thoughts and questions, however:

1. I would merge `value` and `state` such that state, when given a primitive, `state` returns a pointer. This removes the need for `value` entirely.

2. Use the `valueOf` method on pointers to provide the primitive value (awesome idea @CyberAP).

3. Would it be possible to return a pointer from `data()` as long as it returns an object? I'm thinking of being able to define the components state entirely based on the props passed to it. Kind of like how functional components are entirely dependent on props (well that and injections and what not).

I'm guessing that this might already be possible in the proposed API by using computed + object spread (… is it?) so this could be a nice shorthand.

```js
import { computed } from 'vue'

export default {
  data() {
    return computed(() => {
      switch (this.$props.state) {
      case 'pending':
        return {loading: true}
      case 'resolved':
        return {data: "something"}
      case 'rejected':
        return {error: "oh no"}
      }
    })
  }
}
```

4. I have, on occasion, wanted explicit, immtable state updates in components. Kinda like a simplified Vuex on a component level. I remember some past discussion, post, slide, or something which mentioned some way to temporarily toggle reactivity in a Vuex action so that updates would be operating on a temporarily mutable copy of Vuex state (which is otherwise immutable). I believe this was an idea for a future direction to get rid of the need to define mutations? This could probably also be compared to React's `setState` variant that takes a function. Would it be possible to disable automatic reactivity on a component and/or enforce that all state updates must be done through some mechanism making them explicit?

I'm thinking of something like below (The API names below probably aren't the greatest — just for demonstration purposes):

```js
import { immutable, update } from 'vue'

export default {
  data() {
    return immutable({
      state: null,
    })
  },

  methods: {
    loadData() {
      // In `update` this components data is temporarily
      // unlocked so we can mutate it freely
      update(() => {
        this.state = "foo bar"
      })
    }
  },
}
```

With this, any updates to state returned by immutable must be unlocked first. This API would ensure that Vue is still observing changes (so not like `Object.freeze` where Vue doesn't observe anything in the tree) and that the data itself remains effectively immutable. So you can't just pass around the object and mutate it. *However,* being that it is still reactive, reading data from it might return a new value if it has been updated / replaced.

Secondary idea: The update function wouldn't be global. It'd be per immutable instance so only the thing that created the object has access to the mechanism to unlock its mutability.

This is just one approach. I'm 100% open to other ideas in this area.

backbone87:
Isnt this like a poor mans mobx?

LinusBorg:
Ignoring your choice of words, the answer is: yes, in a way. 

MobX provides the same kind of reactivity they Vue does (long before the former), plus some abstractations over it that Vue doesn't have in core. 

Since we are now exposing our reactivity system as a standalone library, the basic functionality is pretty similar.

CyberAP:
I like the way React hooks solved the issue with primitives passed by value, instead of reference. You get a getter and a setter for these values and do not have this issue anymore. Maybe the same could be applied to Vue as well.

```javascript
const [primitive, setPrimitive] = value(0);
watch(primitive, val => console.log(`Value is: ${val}`));
setPrimitive(1); // Value is: 1
```

c01nd01r:
About Computed

Maybe, computed pointers/states should be "depends" on state, like `computed(...states, getterFn)`?

For example: 

```js
const { age } = state({ age: 18 })
const name = value('John');

const userInfo = computed(name, age, (name, age) => `${name} is ${age} years old `);
```

Pros:
* Getter function is pure.
* Encapsulation `value` field of pointer.
* Similar to `watch`.

Cons:
* Computed setter? No ideas :\ Is this feature is necessary?
* Anything else?

laander:
The most common use case for mixins I currently see, is to define a set of helper data/computed/methods/life-cycle actions that can be reused across components for common behaviour. For instance, a simple pagination helper might look something like this:

```javascript
// Vue 2.x mixin
export default {
  data() {
    return {
      currentPage: 1,
      lastPage: null
    }
  },
  computed: {
    nextPage() {
      if (this.currentPage === this.lastPage) return null
      return this.currentPage + 1
    }
  }
}
```

With the new reactivity API, am I correct in assuming that the same could be achieved with something like the following:

```javascript
// Vue 3.x with reactivity helpers
import { value, computed } from '@vue/observer'

const paginationHelper = {
  currentPage: value(1),
  lastPage: value(null)
  nextPage: computed(() => {
    if (currentPage === lastPage) return null
    return currentPage + 1
  })
}

export default {
  data() {
    return {
      ...paginationHelper,
      // local data here
    }
  }
}
```

In a real-life scenario, the `paginationHelper` would obviously be extracted into its own module and exported. What I'm digging at is that the RFC could benefit with some examples of how this works in tandem with components, especially as a means to replace mixins.

LinusBorg:
A couple of replys:

@thecrypticace 

1. Technically, that shouldn't be an issue. However I think the expressiveness of a seperate `value()`, which returns something inherently different than `state()` is important.

2. We'll definitely look into the `valueOf()` idea.

3. No, `data` will have to return an object so we can use its keys as names to inflect the properties onto the component instance.

4. Regarding immutability, I think we even have a rough implementation of a lock mechanism in our prototype somewhere, but so far it's mainly intended to make state immutable in certain situations, i.e. forbid to mutate objects that were passed through props.

It's not part of this RFC though, and currently not externally usable I think. /cc @yyx990803 any input here?

@c01nd01r 

> Maybe, computed pointers/states should be "depends" on state [...]

I don't like it. We are basically taking away the convenience of automatically registering dependencies to make this function "pure", and I don't see how it's useful to have a pure function here.

@laander 

> With the new reactivity API, am I correct in assuming that the same could be achieved with something like the following [...]

Correct.

> What I'm digging at is that the RFC could benefit with some examples of how 
this works in tandem with components, especially as a means to replace mixins.

Fair enough. But this RFC is focussed on the standalone capabilites of the reactivity system. Mixins often also involve lifecycle methods, and we have another RFC for those, in #23.

In that RFC, you can find examples how that RFC and this one we are discussing right now can be used to do the same that mixins do, and more/better.

panstromek:
This is basically mobx, I love it 😄

chriscalo:
Small point, but I had to re-read the writable pointer example a few times before I understood it.

``` js
const writablePointer = computed(
  // read
  () => count.value + 1,
  // write
  val => {
    count.value = val - 1
  }
)
```

Might this be more explicit?

``` js
const writablePlusOne = computed(
  // read
  () => count.value + 1,
  // write
  val => {
    count.value = val - 1
  }
)
console.log(count.value) // 1
console.log(writablePlusOne.value) // 2
writablePlusOne.value = 5
console.log(count.value) // 4
console.log(writablePlusOne.value) // 5
```

eladFrizi:
We can pass to watch multiple values . 
```
watch([plusOne, anotherReacriveRef] , [countPlusOne, anotherReacriveRefValue ] => {
  console.log(`count plus one is: ${countPlusOne} , ${anotherReacriveRefValue}`)
})
```
```
watch([currentArenaRef, CrentUserRef] ,
 ([currentArena, currentUser], [prevArena, prevUser]) => {
   findMatches(currentArena,currentUser ) .... 
})
```
I find it very usefull.

David-Desmaisons:
Mobx exposes API to observe an object as a whole:

```javascript
import {observable, observe} from 'mobx';

const person = observable({
    firstName: "Maarten",
    lastName: "Luther"
});

const disposer = observe(person, (change) => {
    console.log(change.type, change.name, "from", change.oldValue, "to", change.object[change.name]);
});
```
In this API, the change callback receives the name of property that changed as well as the old and new value of the property. This is very helpful when you need to listen to all the potencial changes of an object with only one callback. It will be very practical if Vue 3 provides such an API.

jhoffner:
You'r obviously concerned with establishing with the low level utilities first, but any plan to include a macro builder similar to how the Vue instance works now?

```javascript
const person = reactive({
  data: {
    firstName: 'Joe',
    lastName: 'Smith',
    nameChanges: 0,
  },
  computed: {
    fullName() {
      return `${this.firstName} ${this.lastName}`;
    }
  },
  watch: { 
    fullName() {
      this.nameChanges++;
    }
  }
})
```

It would be easy enough to build on top of the provided low level APIs, but might be better to provide it up front so that a bunch of miscellaneous implementations don't start popping up all over the place.

dodas:
Seriously love that you decided to split reactivity system out, but isn't this just re-implementing MobX?
Couldn't we use MobX directly, or with some thin layer on top of it?
What is advantage of Vue having its own implementation?

Thanks for reply and your work!

backbone87:
mobx also comes with a es6 proxy version

panstromek:
@dodas Well, that would mean that you will have two reactivity systems in the app unnecessarily. Vue will use its reactivity system anyway, this is only a way to expose it as public API. Having this `in-house` is a good way too ensure compatibility, I doubt Vue would ever use 3rd party lib for such a core feature as reactivity system.

You can already use MobX with `mobx-vue`, though ;) But I don't really know what's the cost of having two systems tracking dependencies.. MobX is great, but I guess I'll wait for this API to land, just to be sure I don't get into some trouble ;)

HendrikJan:
Could the watchers probably be written without having to specify what you are watching? 
Like so:

```javascript
// watch with a getter function
watch(() => {
  console.log(`obj.a is: ${obj.a}`)
})

// refs can be watched directly
watch(() => {
  console.log(`count is: ${count}`)
})

watch(() => {
  console.log(`count plus one is: ${plusOne}`)
})
```

The passed function would then recalculate on changes of any reactive value. 
You could also easily watch multiple reactive values:

```javascript
watch(() => {
  console.log(`count is: ${count}, and count plus one is ${plusOne}`)
})
```

LinusBorg:
@David-Desmaisons That's an interesting suggestion! 

However if we implement it, maybe we should consider making it a dedicated API, since I'm not sure that we back backport that functionality to the compatibility build which reimplements the old getter&setter based Reactivity for IE compatibility.

But I'm liking the feature, definitely.

LinusBorg:
@jhoffner I personally don't  see a reason for providing such a mechanism in our own core, since it would be trivial to implement in userland while it doesn't really provide any objectively tangible advantage over doing this:

```javascript
const person = {
  firstName: value('Joe'),
  lastName: value('Smith'),
  nameChanges: value(0),
  fullName: computed(() =>`${person.firstName} ${person.lastName}`)
}

watch(person.fullName, () => person.nameChanges.value++)
```

>  Also I just realized that it's called state when you are working with the low level API, but data when working with a view instance. Any concerns about trying to keep them the same term

We are aware of this and are internally debating what to do with this. Naming things one of the two hard problems in CS, you can imagine it's not easy ^^

LinusBorg:
@dodas Basically what @panstromek said...

LinusBorg:
@HendrikJan Not really, for multiple reasons.

1. This wouldn't allow for lazy watchers as the side effect has to be ru immediately to collect the dependencies
2. Since wachers are meant to have side effects, you could easily create infinite loops:
```javascript
const lock = value(false)
const counter = value(1)
watch(() => {
  if (lock.value && counter !== 0) {
    counter.value++
  }
})
```
The eabove would result in us watching the `lock` and `counter` pointers for  changes. Since we *also* mutate the counter as a side effect, this would lead to an endless update loop.

3. Collecting dependencies in async operations doesn't work.

HendrikJan:
@LinusBorg 

Infinite loops might be easy to detect and give warnings about, but point 1 and 3 are good arguments.

LinusBorg:
> Infinite loops might be easy to detect and give warnings about

Sure. But the consequence is that you *can't* modify any reactive dependency from a watcher that you also need to access.

And trying to technically prevent such an infinite loop in `watch`'s implemantation won't work realibly since there would be *so many* edge cases to consider.

agronick:
Looks good. I'd like to have the same API for objects and primitives. The ability to provide an array to watch on is a must have.

David-Desmaisons:
@LinusBorg 

> However if we implement it, maybe we should consider making it a dedicated API
A dedicated API would be fine indeed. I have a specific scenario for [Neutronium integration](https://github.com/NeutroniumCore/Neutronium) where I need to listen to all properties of an object and to know which property has been changed. With the current implementation, I need to register a callback for each properties. Having a single callback as in mobx would result in a potentially substantial memory optimization.

agronick:
@David-Desmaisons I agree. I'd like to have these internals exposed so I'm not adding duplicate code when I need to do things outside the framework. Things like event handlers and animations work well enough when they can be set up with HTML but when they need to be done in JS I end up adding a bunch of redundant helper functions. This is similar.

yyx990803:
@LinusBorg @HendrikJan note in the RFC this is actually supported:

``` js
watch(() => {
  console.log(obj.a)
})
```

There isn't anything inherently wrong with this, since mutating a watcher's own dependency in the callback would result in infinite loops too:

``` js
watch(somePointer, () => {
  somePointer.value++
})
```

yyx990803:
@eladFrizi note watching multiple values can be easily done in userland:

``` js
watch(
  () => [currentArenaRef.value, currentUserRef.value],
  ([currentArena, currentUser], [prevArena, prevUser]) => {
    findMatches(currentArena,currentUser ) ...
  }
})
```

nerdcave:
What's the advantage of these patterns over simply creating additional Vue instances to encapsulate what you want? Performance? Or am I missing other advantages? For example:

```html
<div id="app">
  <p>app: {{ x }} {{ y }}</p>
  <child></child>
</div>
```

```javascript
const mouseTracker = new Vue({
  data: { x: 0, y: 0 },
  created() {
    window.addEventListener('mousemove', this.update)
  },
  destroyed() {
    window.removeEventListener('mousemove', this.update)
  },
  methods: {
    update(e) {
      this.x = e.pageX
      this.y = e.pageY
    }
  }
})


new Vue({
  el: '#app',
  components: {
    child: {
      template: `<p>child: {{ x }}, {{ y }}</p>`,
      computed: {
        x: () => mouseTracker.x,
        y: () => mouseTracker.y,
      }
    },
  },
  computed: {
    x: () => mouseTracker.x,
    y: () => mouseTracker.y,
  }
})
```
https://jsfiddle.net/cfy8ensb/

LinusBorg:
A direct advantage would be a smaller memory footprint as you don't have to create whole Vue instances when you only need the reactivity functionality itself.

Another comes to light when combining it with #23 (*Dynamic Lifecycle injection*). It allows to colocate behaviours that have to be triggered at various points in a component's lifecycle in a way that is much more fluent and can be effortlessly be combined with other behaviours that are written in that way. You can find a few examples of that in that linked RFC.

Together they form a pattern close to what React's new Hooks API provides. It can  replace Mixins and has none of their drawbacks.

alinnert:
@LinusBorg one question about the package itself. In March you wrote:

> Since we are now exposing our reactivity system as a standalone library, (...).

Does that mean I could install and use the reactivity system in a project without depending on Vue itself? Like `npm install @vue/observer`?

LinusBorg:
Yes.

yyx990803:
Closing as part of #42.

PR #23, opened 2019-03-26
Dynamic Lifecycle Injection

yyx990803:
[Rendered](https://github.com/vuejs/rfcs/blob/dynamic-lifecycle/active-rfcs/0000-dynamic-lifecycle-injection.md)

``` js

import { onMounted, onUnmounted } from 'vue'

export default {
  beforeCreate() {
    onMounted(() => {
      console.log('mounted')
    })

    onUnmounted(() => {
      console.log('unmounted')
    })
  }
}
```

### DIFF
--- a/active-rfcs/0000-dynamic-lifecycle-injection.md
+++ b/active-rfcs/0000-dynamic-lifecycle-injection.md
@@ -1,0 +1,214 @@
+- Start Date: 03-05-2019
+- Target Major Version: 2.x & 3.x
+- Reference Issues: N/A
+- Implementation PR: N/A
+
+# Summary
+
+Introduce APIs for dynamically injecting component lifecycle hooks.
+
+# Basic example
+
+```js
+import { onMounted, onUnmounted } from 'vue'
+
+export default {
+  beforeCreate() {
+    onMounted(() => {
+      console.log('mounted')
+    })
+
+    onUnmounted(() => {
+      console.log('unmounted')
+    })
+  }
+}
+```
+
+# Motivation
+
+In advanced use cases, we sometimes need to dynamically hook into a component's lifecycle events after the component instance has been created. In Vue 2.x there is an undocumented API via custom events:
+
+``` js
+export default {
+  created() {
+    this.$on('hook:mounted', () => {
+      console.log('mounted!')
+    })
+  }
+}
+```
+
+This API has some drawbacks because it relies on the event emitter API with string event names and a reference of the target component instance:
+
+1. Event emitter APIs with string event names are prone to typos and is hard to notice when a typo is made because it fails silently.
+
+2. If we were to extract complex logic into external functions, the target instance has to be passed to it via an argument. This can get cumbersome when there are additional arguments, and when trying to further split the function into smaller functions. When called inside a component's `data()` or lifecycle hooks, the target instance can already be inferred by the framework, so ideally the instance reference should be made optional.
+
+This proposal addresses both problems.
+
+# Detailed design
+
+For each existing lifecycle hook (except `beforeCreate`), there will be an equivalent `onXXX` API:
+
+``` js
+import { onMounted, onUpdated, onDestroyed } from 'vue'
+
+export default {
+  created() {
+    onMounted(() => {
+      console.log('mounted')
+    })
+
+    onUpdated(() => {
+      console.log('updated')
+    })
+
+    onDestroyed(() => {
+      console.log('destroyed')
+    })
+  }
+}
+```
+
+When called inside a component's `data()` or lifecycle hooks, the current instance is automatically inferred. The instance is also passed into the callback as the argument:
+
+``` js
+onMounted(instance => {
+  console.log(instance.$options.name)
+})
+```
+
+### Explicit Target Instance
+
+When used outside lifecycle hooks, the target instance can be explicitly passed in via the second argument:
+
+``` js
+onMounted(() => { /* ... */ }, targetInstance)
+```
+
+If the target instance cannot be inferred and no explicit target instance is passed, an error will be thrown.
+
+### Injection Removal
+
+`onXXX` calls return a removal function that removes the injected hook:
+
+``` js
+// an updated hook that fires only once
+const remove = onUpdated(() => {
+  remove()
+})
+```
+
+# Appendix: More Examples
+
+Pre-requisite: please read the [Advanced Reactivity API RFC](https://github.com/vuejs/rfcs/pull/22) first.
+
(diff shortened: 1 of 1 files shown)


jvbianchi:
In the [Data Fetching](https://github.com/vuejs/rfcs/blob/dynamic-lifecycle/active-rfcs/0000-dynamic-lifecycle-injection.md#data-fetching) example:
```js
const res = value({
    status: 'pending',
    data: null,
    error: null
  })
```
could we use `state` instead of `value`?

LinusBorg:
Yes, that would work as well.

jvbianchi:
Is there any reason to use one or the other in this case?

LinusBorg:
No, except to demonstrate that refs work in this context as well, I guess. They are the new concept, while `state` is just `Vue.observable()`

crutchcorn:
I love the verbosity in the later examples where custom logic was extracted to their own variables. These have an easy to read/understand intuitive API in my opinion. 

I hadn't seen it explicitly mentioned in the docs, but can there be many of these variables in a single data component method? (I'm assuming so, but just double checking)

LinusBorg:
> I hadn't seen it explicitly mentioned in the docs, but can there be many of these variables in a single data component method?

Yes, that's possible, and kind of the point. This allows us to pass state from one of these functions to another, which is something that was not possible before with mixins.

Pseudocode:

```javascript
data() {
  const showsidebar = value(false)

  const { x, y } = useMousePosition()

  watch(x, value =>  {
    showSidebar.value = value < 20
    }
  })

  return  {
    x,
    y,
    showSideBar,
  }
}
```

voluntadpear:
Are there any intentions to deprecate mixins on version 3? I think that the new Advanced Reactivity API along with this proposal are enough for building reusable bits of logic (and besides, we have renderless components and higher-order components if for some reason this is not enough) without the usual drawbacks of mixins (implicit and hard to follow population of state values in `data()` etc.)

LinusBorg:
It's being discussed, yes. But nothing final. 

We obviously won't *remove* mix ins to ensure backwards compatibility in this regard, but *personally* i think we can drop them in v4.

LinusBorg:
@aztalbot 

> Would it be a read-only property in the component’s state same as an ordinary computed property?

Yes.

> In practice, would you expect react-hooks-like external functions to return read-only computed pointers alongside mutable pointers (or even methods/functions)? If so I’m not sure how clear it would be to quickly tell which data object properties are read-only vs mutable (I don’t mean it’s bad, I only mean it’s less obvious than ordinary getters/computed properties).

That's indeed a valid point - if such a helper returns a normal stateful object, a pointer, and a computed pointer, it's not entirely obvious which is which. 🤔

> what order do these injected lifecycle hooks fire? If I inject onUpdated in the data method and onUpdated in the created method, which onUpdated fires first? Is it FIFO based on component lifecycle? I only see it mattering if there is an immediate side effect to the DOM or some external API.

I think we should specificy this, but haven't so far.

I would say it should work like mixins would: Mixed-in lifecycle hooks run first (and here order is FIFO, yes), component's own hooks after that.

amatthieu:
Hello there.

Will there be some restrictions on an dynamic lifecycle injection usage ?
For example, using `onMounted()` in the `beforeDestroy()` lifecycle.

LinusBorg:
I guess that would be something we could with lint rules in solve in eslint-plugin-vue

gwardwell:
@LinusBorg This might not be the right place for this question, so please feel free to point me to the right place to ask. Picking up your dynamic lifecycle injection illustration as the future replacement for mixins:

> ```js
> data() {
> 
>   const { x, y } = useMousePosition()
> 
>   const showSideBar = computed(() => x.value < 10)
> 
>   return  {
>     x,
>     y,
>     showSideBar,
>   }
> }
> ```

I see examples of data, computed values and I get how it would work for lifecycle hooks, but how would I handle props and methods?

trickstival:
## Awesome RFC! 

I'm really excited about seeing the power of hooks being brought to Vue in such native feeling.

In the data fetching example, could the computed() function be used inside the "useFetch" hook?

Instead of this:
```js
  data() {
    return {
      postData: useFetch(computed(() => `/api/posts/${this.id}`))
    }
  }
```

Maybe something like that:
```js
function useFetch(endpointFunc) {
  const endpointRef = computed(endpointFunc)
  // rest of the function
}

const App = {
  // ...
  data() {
    return {
      // Now there's no need to call computed() everytime fetching api
      postData: useFetch(() => `/api/posts/${this.id}`)
    }
  }
}
```

TotooriaHyperion:
What's the recommended pattern to use this when the injected is dependent on props change?
And when with custom functionality that declared through function call to compose with each other?

I mean to achieve something like this.
```javascript
watch: {
    someUniqueToken(newToken, oldToken) {
        // unsubscribe old service
        this.XXXunsub && this.XXXunsub();

        // get a service instance that related to this token. inst.in/inst.out is a Rx Subject
        const inst = getServiceXXX(newToken);
        const subscription = inst.out.subscribe(....);

        this.XXXunsub = () => subscription.unsubscribe();
        this.XXXin = inst.in;
    },
    XXXin(newSubject, oldSubject) {
        // do some logic composition.
    },
}
```

LinusBorg:
```js
data() {
  const unsub = value(null)
  const XXXIn = value(null)

  watch(
    vm => vm.someUniqueToken,
    (newToken, oldToken) => {
      unsub.value && unsub.value()

      const inst = getServiceXXX(newToken);
      const subscription = inst.out.subscribe(....);

        unsub.value = () => subscription.unsubscribe();
        XXXin.value = inst.in;
    },
  )

  watch(
    () => XXXin,
    (newSubject, oldSubject) {
        // do some logic composition.
    }
  )

  return {
    XXXIn
  }
}
```

You wouldn't even have to return `XXXIn` from `data` for it to work (unless you need it in the template or another method).

And could extract all of this beahvour in a function that you import run in `data`.

```js
useMyRxLogic(propName) {
  const unsub = value(null)
  const XXXIn = value(null)

  watch(
    vm => vm[propName],
    (newToken, oldToken) => {
      unsub.value && unsub.value()
      const inst = getServiceXXX(newToken);
      const subscription = inst.out.subscribe(....);

      unsub.value = () => subscription.unsubscribe();
      XXXin.value = inst.in;
    },
  )

  watch(
    () => XXXin,
    (newSubject, oldSubject) {
        // do some logic composition.
    }
  )

  return {
    XXXIn,
    unsub,
  }
}
```
```js
data() {
  const { unsub, XXXIn } = useMyRxLogic('someUniqueToken')

  return {
    myUnSub: unsub,
    in: XXXIn
  }
}
```

yyx990803:
Closing as part of #42.

PR #26, opened 2019-04-08
Attr fallthrough behavior

yyx990803:
- Disable implicit attribute fall-through to child component root element

- Remove `inheritAttrs` option

[Rendered](https://github.com/vuejs/rfcs/blob/attr-fallthrough/active-rfcs/0000-attr-fallthrough.md)

### DIFF
--- a/active-rfcs/0000-attr-fallthrough.md
+++ b/active-rfcs/0000-attr-fallthrough.md
@@ -55,7 +55,7 @@ In 3.0, we are planning to make attribute fallthrough an explicit decision of co
 
 - `inheritAttrs` option will be removed.
 
-- `.native` modifier will be removed.
+- `.native` modifier for `v-on` will be removed.
 
 - Non-prop attributes no longer automatically fallthrough to the root element of the child component (including `class` and `style`). This is the same for both stateful and functional components.
 
@@ -74,6 +74,8 @@ In 3.0, we are planning to make attribute fallthrough an explicit decision of co
   ```
 
   Both `bar="2"` AND `class="bar"` on `<child>` will be ignored.
+ 
+- `this.$listeners` will be removed.
 
 - `this.$attrs` now contains **everything** passed to the component except those that are declared as props or custom events. **This includes `class`, `style`, `v-on` listeners (as `onXXX` properties)**. The object will be flat (no nesting) - this is possible thanks to the new flat VNode data structure (discussed in [Render Function API Change](https://github.com/vuejs/rfcs/pull/28)).
 
--- a/active-rfcs/0000-attr-fallthrough.md
+++ b/active-rfcs/0000-attr-fallthrough.md
@@ -181,7 +181,7 @@ This will be one of the changes that will have a bigger impact on existing code 
 
   - Declare props and events that are consumed by the HOC itself (thus removing them from `$attrs`)
 
-  - Refactor the component and explicitly add `v-bind="$attrs"` to the target inner component or element. For render functions, apply `$attrs` with the `cloneVNode` helper.
+  - Refactor the component and explicitly add `v-bind="$attrs"` to the target inner component or element. For render functions, apply `$attrs` with the `mergeData` helper.
 
   - If a component is already using `inheritAttrs: false`, the migration should be relatively straightforward.
 
--- a/active-rfcs/0000-attr-fallthrough.md
+++ b/active-rfcs/0000-attr-fallthrough.md
@@ -108,25 +108,17 @@ In 3.0, we are planning to make attribute fallthrough an explicit decision of co
   }
   ```
 
-  In render functions, if simple overwrite is acceptable, `$attrs` can be merged using object spread. But in most cases, special handling is required (e.g. for `class`, `style` and `onXXX` listeners). Therefore a `cloneVNode` helper will be provided. It handles the proper merging of VNode data:
+  In render functions, if simple overwrite is acceptable, `$attrs` can be merged using object spread. But in most cases, special handling is required (e.g. for `class`, `style` and `onXXX` listeners). Therefore a `mergeData` helper will be provided. It handles the proper merging of VNode data:
 
   ``` js
-  import { h, cloneVNode } from 'vue'
+  import { h, mergeData } from 'vue'
 
   const Child = {
     render() {
-      const inner = h(InnerComponent, {
-        foo: 'bar'
-      })
-      return cloneVNode(inner, this.$attrs)
+      return h(InnerComponent, mergeData({ foo: 'bar' }, this.$attrs))
     }
   }
   ```
-
-  The 2nd argument to `cloneVNode` is optional. It means "clone the vnode and add these additional props". The `cloneVNode` helper serves two purposes:
-
-  - Avoids mutating the original VNode
-  - Handles special merging logic for `class`, `style` and event listeners
 
   Inside render functions, the user also has the full flexibility to pluck / omit any props from `$attrs` using 3rd party helpers, e.g. lodash.
 
--- a/active-rfcs/0000-attr-fallthrough.md
+++ b/active-rfcs/0000-attr-fallthrough.md
@@ -1,0 +1,196 @@
+- Start Date: 2019-04-08
+- Target Major Version: 3.x
+- Reference Issues: N/A
+- Implementation PR: N/A
+
+# Summary
+
+- Disable implicit attribute fall-through to child component root element
+
+- Remove `inheritAttrs` option
+
+# Basic example
+
+To replicate 2.x behavior in templates:
+
+``` html
+<div v-bind="$attrs">hi</div>
+```
+
+In render function:
+
+``` js
+import { h } from 'vue'
+
+export default {
+  render() {
+    return h('div', this.$attrs, 'hi')
+  }
+}
+```
+
+# Motivation
+
+In 2.x, the current attribute fallthrough behavior is quite implicit:
+
+- `class` and `style` used on a child component are implicitly applied to the component's root element. It is also automatically merged with `class` and `style` bindings on that element in the child component template.
+
+  - However, this behavior is not consistent in functional components because functional components may return multiple root nodes.
+
+  - With 3.0 supporting fragments and therefore multiple root nodes for all components, this becomes even more problematic. The implicit behavior can suddenly fail when the child component changes from single-root to multi-root.
+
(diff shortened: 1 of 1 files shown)


alexsasharegan:
Could we get clarification on how best to reproduce the current class/style attr merging in the 3.x api when these conditions exist:
- the component already has a class/style applied to it's root node (just 1 in this case)
- the component wants to also spread the other attrs (to support things like aria as mentioned), but declares a few of its own that it doesn't want over-written

I think for myself and many others with a lot of JSX experience, we know there are gotchas for exactly where you spread attrs and the desired behavior--spreading before defining your own attrs ensures they aren't over-written, while spreading after has the effect of providing default attr values.

yyx990803:
@alexsasharegan that's a good question. In templates `v-bind="$attrs"` handles `style/class` merging for you. Like you mentioned, we should make sure the overwriting behavior is determined by its position relative to other existing attributes. So if you don't want certain attributes to be overwritten, place them after the `v-bind`:

``` html
<div v-bind="$attrs" aria-hidden="true"></div>
```

In render functions, Vue will expose a `mergeData` helper (see updated RFC) that performs proper merging. Our JSX plugin will also compile spread props into `mergeData` calls so that `style/class` merging is handled properly:

``` jsx
<div { ...this.$attrs } aria-hidden="true"/>
```

Both the template and JSX examples above compiles into:

``` js
return h('div', mergeData(this.$attrs, { 'aria-hidden': true }))
```

Justineo:
Now in 2.x, single bindings always overwrite the `v-bind` object and I think it's quite straightforward. Not sure if relying on attribute order is a good idea (applying JavaScript semantics in HTML scope).

alexsasharegan:
As the author of [vue-functional-data-merge](https://github.com/alexsasharegan/vue-functional-data-merge), I like the sound of that!

Justineo:
Emmm...It seems that we have inconsistent behavior for components and elements now:

https://jsfiddle.net/Justineo/hx1nfbry/

mika76:
> we should make sure the overwriting behavior is determined by its position relative to other existing attributes.

This doesn't sound so great to me - it's hidden from anybody who hasn't read docs in great detail. I don't think Vue has anything else like it. Isn't there a more declarative way? Adding a modifier or something?

posva:
@Justineo I think that's working normally because in the case of inputs, the `value` attribute is used during the first render but then v-bind is used by Vue to change the value, whereas, with components, the attribute value takes the upper hand.

If you use a `v-bind` on input, it shows off consistent

alexsasharegan:
@mika76, this kind of stuff gets into the weeds of the underlying abstraction--the virtual dom. Component attributes are described by objects, so when you have to merge attribute objects from multiple sources, the simplest tool is `Object.assign`, which is sensitive to position such that the last source to provide a key will set that key's final value in the resulting object (`argument[0]`, the target of `Object.assign`).

Vue has been really friendly and helpful when it comes to merging attributes, especially given the deep merging of things like `class` & `style`. I understand that class/style merging will continue, and anyone getting into the weeds writing function components will now have access to those merge behaviors by way of exposed helper functions.

I agree that this whole area could benefit from additional detail in the docs. One thing to consider, what assumptions do you have about what would happen currently if you saw a component written with duplicate attributes like this:

```html
<x-button disabled type="button" :disabled="false">
  Example Component
</x-button>
```

Justineo:
> If you use a `v-bind` on input, it shows off consistent

@posva Yeah you are right! I prefer this behavior than relying on the order of attributes in templates.

dima74:
As a developer who worked both in vue and react projects, I strongly dislike disabling fall-through for `class` and `style`. For me it was always painful to style child components in react, because I had to create unnecessary wrappers (which are ugly and also are bad from both semantically and performance point of view) and use tedious helpers for merging `class`es. And I always was exciting about how elegant vue allows to style child components (thanks to fall-through behaviour).

alexsasharegan:
Maybe we can get some clarification, @dima74, because my assumption was that class/style merging would continue except when you author your own render functions in function components. I think this does get confusing when you consider there are many ways to describe how a component should render:

- Instance Component
  - using a dom template
  - using a string template
  - `.vue` `<template>`
  - render method
    - createElement calls
    - JSX vue dialect
- function component
  - `<template functional>`
  - createElement calls
  - JSX vue dialect

And maybe more (are dom/string templates okay for function components?).

posva:
it's the same fall through behaviour except now it has to be explicit. This is to make it consistent with Fragments components (more than one element at the root), which wasn't possible before. You won't need any wrappers

sdras:
In class and style bindings that are dependent on a parent/child relationship such as position: relative on a parent and position: absolute on child to place it on a containing unit, this becomes a bit more messy. There are workarounds but I'm concerned that this impedes the ability to position things without some annoyances. Happy to talk through it if anyone would like clarification.

casey6:
So to enable the 2.x functionality you need to add `v-bind="$attrs"` to every single component in your application? That feels pretty inconvenient.

I understand the argument that it encourages a more "correct" way of building components. But in the real world people aren't perfect and I'd argue using style/class for child component styling is a great way to achieve your goals in the absence of perfection.

I'd be interested to see how many people currently write code like `<my-component class="some-class"/>` for styling purposes. I suspect the number is extremely high.

LinusBorg:
I came across one blind spot with this concept: Event modifiers.

We currently provide a couple of handy modifiers such as `.prevent` for native event listeners:

```html
<a @click.prevent="handleLinkClick">
```

Some of these modifiers (i.e. `.prevent`) interact with the event object that the native event passes to the listener, i.e. to call `event.preventDefault()`. These modifiers don't work for component events, obviuosly, and aren't meant to.

In Vue 2, if we wanted to listen to a native event and prevent its default behaviour, we could do this:

```html
<fancy-a @click.native.prevent="handleLinkClick">
```

However, this RFC drops support for `.native`. Instead, the child component is now in full control over where to attach any and all `v-on` listeners that were defined for it by the parent. Converning modifiers, this seems to create leaky abstractions - the consumer of our component has to be aware how an event listener would be used in the child in order to determine wether using `.native` would work or not.

```html
<fancy-a @click.prevent="handleLinkClick">

// implementation in FancyA:
<a v-bind="$attrs"> <!-- WORKS -->

<a v-on:click.prevent="prepareStuff">
<script>
  export default {
    methods: {
      prepareStuff(e) {
        // do something else, then:
        this.$emit('click', e.target.value) <!-- BREAKS -->
      }
    } 
  }
</script>
```

aminimalanimal:
I like the improvements this makes for `v-model` and passing properties to form elements, but I'm alarmed about the proposed changes to fallthrough classes and styles.

I use fallthrough classes to handle making adjustments to the layout or positioning of child elements given _circumstances that the child components need not be aware of_.

The documentation attached to this RFCS states the following (emphasis mine):
> This may cause some inconvenience for users accustomed to the old behavior, especially when using class and style for styling purposes, _but it is the more "correct" behavior when it comes to component responsibilities and boundaries_.

I _think_ we can all agree on the following rules for the “correct” behavior for component responsibilities and boundaries (regarding styling):
- components should handle their own appearance
- components should handle their own “internal” layout
- a parent component shouldn't target any class specified by a child component
- parent components should handle in-context/situational “outer” layout, positioning, and sizing of child components

The beauty of Vue 2's behavior is that if a parent component needs to style a child component's root element directly:
- the override stays entirely in one file—in its proper context
- no additional containers are added to the DOM
- no semantics are ever impacted
- _the child component does not need to be explicitly set up to play nice_
- the API for adding the class is consistent (at least within templates)

With the proposed changes, third-party components would have to _allow_ your parent components to pass a class onto them, which from a component's contractual standpoint is, I _suppose_, more “correct,” but I really struggle to see any way in which this limitation is an improvement to the developer experience since the API for adding a class to the root element would no longer be consistent.
- it may not be present at all
- it may be present, but implemented in various ways
  - best-case `$attrs` method: if `$attrs` needs be applied to a form element for `v-model` support, a developer could presumably create a `computed` property to _omit_ the class/style portions from the form element, and apply only the class/style portions to the root element; otherwise, just place `$attrs` on the root element
  - clumsy `$attrs` method: `$attrs` is just dumped in both spots entirely, causing an internal element to gain a class you probably only wanted on the root
  - prop method: set up a prop for a custom class, forcing the parent component to adhere to a non-standard prop name (and pug users miss out on the syntactical sugar of just sticking it on as `MyComponent.my-class`, but oh well...)

I really think that the current fallthrough capability for classes and styles is important to maintain.

There may be use-cases where a parent component needs to add/remove/adjust a component's default `margin`, `height`, `width`, `position`, directional properties, `overflow`, (or, and I hesitate to say this—`float`)... As @sdras pointed out, depending on the context, a wrapper component is not always sufficient for handling these things... and there are certainly plenty of situations I've ended up in where a flatter DOM structure was better or even necessary to accomplish the styling task at hand.

To @casey6 's point, I use fallthrough classes in almost every parent/child relationship, so if this behavior ships in Vue 3, I'm probably going to have to/want to implement an explicit `$attrs` solution in practically every component I make.

As @dima74 mentioned, having to wire this stuff up manually in React was **painful**... Vue 3 would still be _better_ than that thanks to class merging, but not better than Vue 2 due to the lack of consistent application. I think that the inconvenience of having to manually wire this up is likely to cause developers to add styles that impact outer-layout directly to their components.

**Overall**, I think that `class` and `style` should be treated differently from other properties in this regard, maintaining their current functionality of being placed on the root element, and that there should be a version of `$attrs` that excludes them for the purpose of solving the `v-model`/property issue for forms that this RFCS addresses.

In cases where there are multiple root elements, I would want my fallthrough class applied to each of them. (But if that isn't possible, I'd rather have the inconsistency than the proposed inconveniences, as instances in which I need fragments are fairly rare (`dd`/`dt`)).

This is one of the bits of functionality that I think makes Vue great, and makes it a cut above the rest. I hope it's possible to keep. (And... _I just spent five hours writing this instead of sleeping. Awesome_.)

tmorehouse:
What about using an options flag, that denotes that the component renders (or could render) more than one root element?

It many cases, adding an outer wrapper for styling positioning adjustments can break layout for some CSS libraries that expect a certain element structure, specifically CSS sibling selectors (when components are children of a parent element).

How would consumers know if a component will merge style/classes on it's root element?

yyx990803:
After core team discussion I'm closing this since it is a major breaking change for a lot of use cases, and it's very difficult to provide a smooth upgrade path. We will be submitting a new RFC with a smaller scope of changes instead.

PR #31, opened 2019-04-10
Component v-model API change

yyx990803:
Adjust `v-model` API when used on custom components.

This builds on top of #8 (Replace `v-bind`'s `.sync` with a `v-model` argument).

[Rendered](https://github.com/vuejs/rfcs/blob/v-model/active-rfcs/0000-v-model-api-change.md)


jvbianchi:
How will this be implemented in JSX?

CyberAP:
Having an additional attribute for custom elements to support native-like behaviour for v-model is kind of strange. I would love to see v-model support an extra argument for that case with a special symbol, for example:

```
<custom-input v-model:enabled@checkbox="inputEnabled" />
```

posva:
@CyberAP I also thought about that, but a modifier woud make more sense instead of `@`: `v-model:enabled.checkbox="inputEnabled"`

leopiccionia:
It would be nice if `lazy` was a `type`, whichever syntax is ultimately chosen.

GerardRodes:
Why
```
v-model:name="value"
```
Instead of
```
:name.sync="value"
```

edcoreweb:
This will be breaking a lot of packages which implement v-model on custom components.

> In 3.0, the model option will be removed. v-model="foo" (without argument) on a component compiles to the following instead:

Why not leave the model option for the first few 3.x versions, deprecate it so that the upgrade path will be easier for existing Vue 2.x packages ?

beeplin:
@edcoreweb i think such option removals must happen in major version bumps like 2->3 or 3->4. If you don't remove it in 3.0.0, you will not be able to remove it until 4.0.0.

backbone87:
wouldnt it be better to name the default model value prop `default`? this aligns with slots and es modules.
maybe the BC break is a little harder since `default` is more likely to conflict than `modelValue`, but not **that** likely and there are BC breaks regarding this anyway, so might as well take pill now for improved consistency.

atilkan:
This is much better.

yyx990803:

active-rfcs/0000-v-model-api-change.md
```diff
@@ -0,0 +1,152 @@
...
+``` html
+<InviteeForm
+  v-model:name="inviteeName"
+  v-model:email="inviteeEmail"
+/>
+```
+
+## Handling Modifiers
+
+In 2.x, we have hard-coded support for modifiers like `.trim` on component `v-model`. However, it would be more useful if the component can support custom modfiers. In v3, modifiers added to a componen `v-model` will be provided to the component via the `modelModifiers` prop:
```
jacekkarczmarczyk:
```suggestion
In 2.x, we have hard-coded support for modifiers like `.trim` on component `v-model`. However, it would be more useful if the component can support custom modfiers. In v3, modifiers added to a component `v-model` will be provided to the component via the `modelModifiers` prop:
```

sqal:
Regarding `modelModifiers` prop. I noticed that if I use more than one v-model on the component with different modifiers, then compiler creates modelModifiers prop with modifiers only from the first v-model. is this a bug?

Demo: https://vue-next-template-explorer.netlify.com/#%7B%22src%22%3A%22%3CComp%20%20%5Cr%5Cn%20%20v-model%3Aname.a.b%3D%5C%22name%5C%22%5Cr%5Cn%20%20v-model%3Aemail.c%3D%5C%22email%5C%22%5Cr%5Cn%2F%3E%22%2C%22options%22%3A%7B%22mode%22%3A%22module%22%2C%22prefixIdentifiers%22%3Afalse%2C%22hoistStatic%22%3Afalse%2C%22cacheHandlers%22%3Afalse%7D%7D

yyx990803:
@sqal yeah, good catch! We need to generate different prop for `v-model` with args.

CyberAP:
Why it's not possible to write model handler in render function with modifiers? `onUpdate:foo.number`

denisinvader:
I think, writing `'onUpdate:modelValue'` in JSX is quite ugly. So, to implement an input in a shared library that fits templates and render functions ways you have to support two props and two events (for single-prop v-model) or write uncommon (less native/classic) JSX

yyx990803:
@CyberAP @denisinvader the format is optimized for the consuming component and the compiler. For manual render functions you can always build a little helper to make the usage read nicer:

``` js
withModel(h(SomeComp), {
  prop: 'foo', // optional
  value: foo,
  update: value => (foo = value),
  modifiers: { trim: true }
})
```

@denisinvader Vue will have its own JSX transform which gives us the ability to handle the transform for you. The 2.x transform only handles `<Comp vModel={foo} />`, we do need to discuss how the syntax should look like for v-model with arguments and modifiers in JSX for 3.0.

mediafreakch:
How do the changes discussed in this RFC differ from the `:prop.sync` syntax that already exists today? Isn't it the same?

If it really is it's not clear to me when to use which and should be addressed by a unique API design.

But maybe it isn't and someone could explain the difference?

zero7u:
我们可以用`v-bind`传递父组件属性，有没有方法可以传递父组件的`models`？

PR #32, opened 2019-04-12
Custom Directive API Change

yyx990803:
- Re-design custom directive API so that it better aligns with component lifecycle

- Custom directives usage on components will follow the same rules as discussed in the [Attribute Fallthrough Behavior RFC](https://github.com/vuejs/rfcs/pull/26). It will be controlled by the child component via `v-bind="$attrs"`.

[Rendered](https://github.com/vuejs/rfcs/blob/custom-directive/active-rfcs/0000-custom-directive-api-change.md)


zhangenming:
image

this will be better?

componentUpdated -> removed ,use updated instead

Gnuk:
@zhangenming, I think it's the same case as `bind -> beforeMount`, `inserted -> mounted` and `unbind -> unmounted`, so this suggestion may not be needed for me.

yyx990803:

PR #33, opened 2019-04-13
Slot framing

backbone87:
[Rendered](https://github.com/vuejs/rfcs/blob/15d959b8a2ae3361a226df0a4d343cede7c6920d/active-rfcs/0000-slot-framing.md)

```html
<template>
  <div class="wrapper">
    <slot>
      <div class="frame">
        <slot-content>
          Fallback content
        </slot-content>
      </div>
    </slot>
  </div>
</template>
```

```html
<template>
  <my-component>
    Custom content
    <slot-fallback />
  </my-component>
<template>
```

### DIFF
--- a/active-rfcs/0000-slot-framing.md
+++ b/active-rfcs/0000-slot-framing.md
@@ -1,0 +1,251 @@
+- Start Date: 2019-03-28
+- Target Major Version: 3 (2 is possible)
+- Reference Issues: TBD
+- Implementation PR:
+
+# Summary
+
+Controlling framing markup around (function-based) slots is hard or impossible
+to manage in template based rendering. This RFC proposes new (compiler-level)
+components to allow clear and concise handling of advanced slot usage scenarios.
+
+# Basic example
+
+```html
+<template>
+  <div class="wrapper">
+    <slot>
+      <div class="frame">
+        <slot-content>
+          Fallback content
+        </slot-content>
+      </div>
+    </slot>
+  </div>
+</template>
+```
+
+```html
+<template>
+  <my-component>
+    Custom content
+    <slot-fallback />
+  </my-component>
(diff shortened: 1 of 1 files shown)


posva:
I don't follow here. What problem is this solving? is it about using `$slots` and `$scopedSlots` in the template? If that's the case, it would make more sense to have some kind of api to check if the slot is provided

backbone87:
Its about the question, if the rendered slot is empty or not (which is different from the test `v-if="$slots.default"`). Using an API for this, either involves rendering the slot and "throwing away" the result (so it must be rendered again), or, to avoid the multiple rendering, using some memoizing pattern, which is not that easy with "any value"-arguments (e.g. a simple map is not enough).

While having an API for this could solve even more use cases, from my experience the proposed solution in this feature resolves the most common use cases. And like i wrote in "Detailed design" all of the proposed features can be solved by using custom render functions already.

LinusBorg:
> Using an API for this, either involves rendering the slot and "throwing away" the result (so it must be rendered again), or, to avoid the multiple rendering, using some memoizing pattern, which is not that easy with "any value"-arguments (e.g. a simple map is not enough).

This is a good point that the new slots API doesn't provide an answer for. So thanks for bringing it up.

I'll have a closer look over the easter holidays now.

backbone87:
@LinusBorg any news about this?

LinusBorg:
Hey, so this is still open and to be discussed, but honestly the other bigger RFCs took up most of our time recently, especially with the most recent influx of feedback.

I still see the point you address and we will come to it. 

Thanks for your patience. :)

yyx990803:

---

I personally believe such use cases should be dealt with using render functions rather than further complicating the template syntax.

PR #34, opened 2019-05-10
Scoped slot API for router-link

posva:
[Rendered](https://github.com/posva/rfcs/blob/router/link-scoped-slot/active-rfcs/0000-router-link-scoped-slot.md)

### DIFF
--- a/active-rfcs/0000-router-link-scoped-slot.md
+++ b/active-rfcs/0000-router-link-scoped-slot.md
@@ -1,0 +1,122 @@
+- Start Date: 2019-04-29
+- Target Major Version: Vue (2.x / 3.x) Vue Router (3.x / 4.x)
+- Reference Issues: https://github.com/vuejs/vue-router/issues/2611,
+- Implementation PR: (leave this empty)
+
+# Summary
+
+- Remove `tag` prop
+- Remove `event` prop
+- Stop automatically assigning click events to inner anchors
+- Add a scoped-slot API
+
+# Basic example
+
+```vue
+<router-link to="/">
+  <Icon>home</Icon> Home
+</router-link>
+```
+
+# Motivation
+
+Current implementation of Router Link has many limitations:
+
+- Active state customization is not complete ([#2611](https://github.com/vuejs/vue-router/issues/2611)
+- Cannot be integrated with custom components ([#2021](https://github.com/vuejs/vue-router/issues/2021))
+- click event cannot be prevented (through `@click.prevent` nor through a `disabled` attribute [#2098](https://github.com/vuejs/vue-router/pull/2098))
+
+The idea of this RFC is to solve those issues by providing a scoped slot that allow app developers to easily extend Links based on their applications and to allow library authors to provide an even easier integration with Vue Router.
+
+# Detailed design
+
+## Slot with content
+
+A simple use case would be slot with content (no nested anchors or buttons)
+
+```vue
+<router-link to="/">
+  <Icon>home</Icon> Home
+</router-link>
+```
+
+This implementation would:
+
+- generate an anchor (`a`) element and apply the corresponding properties:
+  - `href` with the destination
+  - `class` with `router-link-active` or `router-link-exact-active` (can be changed through prop or global option)
+  - click listener to trigger navigation through `router.push` or `router.replace` with a `event.preventDefault`
+- Put anything passed as the children of the anchor
+- Pass down any attributes that aren't props to the `a` element
+
+**Breaking changes**:
+
+- no longer accept a `tag` prop -> use the scoped slot instead (see the point below)
+- no longer accepts `event` -> use the scoped slot instead
+- no longer works as a wrapper automatically looking for the first `a` inside -> use the scoped slot instead
+
+## Custom `tag` prop
+
+I am not sure about keeping the `tag` prop if it can be replaced which a scoped slot because it wouldn't handle custom components and except for very simple cases, we will likely use custom UI components instead of the basics ones:
+
+```vue
+<router-link to="/" tag="button">
+  <Icon>home</Icon><span class="xs-hidden">Home</span>
+</router-link>
+```
+
+is equivalent to
+
+```vue
+<router-link to="/" v-slot="{ navigate, classes }">
+  <button role="link" @click="navigate" :class="classes">
+    <Icon>home</Icon><span class="xs-hidden">Home</span>
+  </button>
+</router-link>
+```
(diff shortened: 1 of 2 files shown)


Akryum:
I love it so much 😻

Rolanddoda:
That would be awesome and very flexible

adamwathan:
The scoped slots API looks fantastic, very well thought-out 👍 Nice job!

Akryum:
I would change `isActive` to `isSamePath` and `isExactActive` to `isSameExact` or `isSame` (or similar), to be more consistent with `isSameQuery` and `isSameHash`.

posva:
we can add `isSamePath` for consistency but `isActive` is more of an inclusion like `string.startsWith` than an equality. It's an equality if in #35 we use `exact="path"` though

Akryum:
For `isActiveExact`, I'm not sure changing the behavior of passed slot props depending on other props would be a good idea.

My suggestion of slot props:

- `isActive`: true whenever `router-link-active` is applied. This is when the _path_ section of the href (without the query and hash) is included (as in `currentLocation.path.startsWith(location.path)`)
- `isSameURL`: true whenever the _fullPath_ is matched (including query and hash) (as in `currentLocation.fullPath === location.fullPath`)
- `isSamePath`: true if the path of the link matches current path
- `isSameQuery`: true if the query of the link matches current query
- `isSameHash`: true if the hash of the link matches current hash
- `isDescendant`: true if the route record matched by the `to` location is a descendant of the record matched by the current route (as in `children` (works for nested `children` as well))
- `isAscendant`: true if the route record matched by the `to` location is an ascendant of the record matched by the current route (as in current route being in `children`(works for nested routes as well))

Akryum:
> It's an equality if in #35 we use exact="path" though

The `exact` prop shouldn't change `router-link-active` though

posva:
Adding a `isSameURL` (or `isSameFullPath` for consistency) makes me wonder if we should just leave `isSameQuery` as it is the only one that requires comparing objects and cannot be achieved comparing values in `location` and `$route` 🤔 

The reasoning behind `isExactActive` is to provide a way to know whether the `router-link-exact-active` would be applied or not. That's why it makes sense to change its behavior based on the `exact` prop. The same goes for `isActive`, its behavior also changes based on `exact`. Which just made me realize there is a confusion part on my exact prop proposal: right now providing exact changes the behavior of the active class, which will be applied only with an exact match, alongside the exact class. Having a string value would still follow that reasoning, so the link always has both classes or none of them

Akryum:
So maybe we can just add `isSameFullPath` in addition to `isExactActive`

Akryum:
> if we should just leave `isSameQuery` as it is the only one that requires comparing objects and cannot be achieved comparing values in `location` and `$route` 🤔 

But it would be very handy to have the others. Less cognitive load imo.

Akryum:
## Helper slot props

### Classes

- `isActive`
- `isExactActive`

### Route comparison

- `isPathIncluded`
- `isSameFullPath`
- `isSamePath`
- `isSameQuery`
- `isSameHash`

### Route tree

- `isDescendant`
- `isAscendant`

CesarGomezTissini:
This awesome rfcs take me with a new question: is it possible to use named router views in future (like Angular) assuming in that framework both implementations have a relation?

JamesCoyle:
This gets a 👍 from me. I believe this would solve an issue I'm currently having.

I'm currently trying to implement a generic `tab-item` element that can be any tag (`button`, `a`, `router-link`, etc). I've currently implemented this functionality but I don't always want the tab item to be a router link. I need a clean way to handle active state by either manually setting it based on certain application state if it isn't a router-link or having it set automatically if it is a router-link. 

Ideally I'd like to be able to define a tab-item as its own component that can render as a button, anchor, or other custom component and then, if desired, wrap the router-link functionality around that without adding unnecessary DOM elements. I like this proposal as it provides access to the data router-link uses such as the href and active state which AFAIK aren't currently accessible in a reactive state. 

### Examples

Render as `a`  just using `tab-item`
```html
<!-- template -->
<tab-item tag="a" href="/tab-test" active>Tab Link</tab-item>

<!-- rendered -->
<li class="tab-item active">
    <a href="/tab-test">Tab Link</a>
</li>
```

Render as `a` by wrapping in a `router-link`
```html
<!-- template -->
<router-link to="/tab-test" v-slot="{ isActive, href }">
  <tab-item tag="a" :href="href" :active="isActive">Tab Link</tab-item>
</router-link>

<!-- rendered -->
<li class="tab-item active">
    <a href="/tab-test">Tab Link</a>
</li>
```

Render as `button` by wrapping in a `router-link`
```html
<!-- template -->
<router-link to="/tab-test" v-slot="{ isActive, navigate }">
  <tab-item tag="button" @click="navigate" :active="isActive">Tab Link</tab-item>
</router-link>

<!-- rendered -->
<li class="tab-item active">
    <button>Tab Link</button>
</li>
```

donnysim:
Would it recognize active state when it contains children from different route trees? Like they have nothing in common, just grouped under one element?

Resources (<- should be marked as isDescendant?)
    - Users (/users)
    - Clients (/clients, active)

posva:
if you are on Resources and Users and clients and children of that route record, then yes, isDescendant will be true for those two links

donnysim:
Children as in template composition wise, not router configuration wise?

posva:
oh, always related to routing configuration. We are comparing routing records at the end

donnysim:
Was hoping it could also do it template wise, as there's often groups of routes that are not in the same tree, but I think it will be solvable if there will be child route addition functionality on the router. We have modules that register separate routes, and because of the current router state, all we can do is add a duplicate base route to the root that wraps the intended route as children.

I also wonder if/how it'll make certain scenarios simpler? Taking sidebars as an example, on page open, you want the active tree to be expanded by default, but it also should not force it to stay active as the user might collapse it. Though this might be something to solve in userland with a wrapper component?.

plehnen:
I propose renaming `navigate` to `onNavigate` making it more implicit that this is an event handler (same reason why isActive isn't called just active)

posva:
Thanks but navigate is an action that you attach as an event handler, so having a verb makes more sense imo

Eduardo San Martin Morote

posva:
I updated the RFC to include only the safe information. This API has already been implemented in vue-router v3 so any new breaking changes would only be implemented in v4

cawa-93:
@posva Shouldn't the Rendered link be updated?

PR #35, opened 2019-05-10
propsal for exact active behavior customization in router-link

posva:
[Rendered](https://github.com/posva/rfcs/blob/router/link-exact-active-behavior/active-rfcs/0000-router-link-active-behavior.md)

### DIFF
--- a/active-rfcs/0000-router-link-active-behavior.md
+++ b/active-rfcs/0000-router-link-active-behavior.md
@@ -1,0 +1,77 @@
+- Start Date: 2019-05-10
+- Target Major Version: Vue (2.x / 3.x) Vue Router (3.x / 4.x)
+- Reference Issues: https://github.com/vuejs/vue-router/issues/2040
+- Implementation PR: (leave this empty)
+
+# Summary
+
+Customize current _exact_ matching behaviour by allowing a string value to the `exact` prop.
+
+# Basic example
+
+```vue
+<!-- same as `exact` boolean prop -->
+<router-link to="/some">/some</router-link>
+<!-- matches only the path, eg: `/?query`, `#foo`, `?q=hey`. But not `/a`, `/some` -->
+<router-link to="/" exact="path">Home</router-link>
+<!-- matches the path and query eg: `?q=hey&a=b`, `?a=b&q=hey`, But not `/` or `?q=hey` -->
+<router-link to="/?q=hey&a=b" exact="path+query">Home</router-link>
+```
+
+# Motivation
+
+Current `exact` behavior is not flexible enough. A boolean value can only handle two cases but there are many more when dealing with query params or hashes as part of the link and making them responsible for marking the link as active.
+
+# Detailed design
+
+Right now, the active state can either be _exact_ or normal by passing the `exact` prop. However this is quite limiting for path containing query params or a hash which are taked into account when exact matching.
+If the `exact` prop is not supplied, the link is considered active when it is included by the current location path as a string:
+eg: `<router-link to="/">Home</router-link>`/`<router-link :to="{ name: 'home' }">Home</router-link>` would be highlighted by any link (parent-child relationship is not necessary). The same goes for these two routes:
+
+```js
+const routes = [{ path: '/some' }, { path: '/some/option' }]
+```
+
+Both links are actives while being on `/some/option`:
+
+```vue
+<router-link to="/some">/some</router-link>
+<router-link to="/some/option">/some/option</router-link>
+```
+
+If `/some/option` is matched, the link `/some` will be marked as active but if another link like `/some-option` was matched (note the `/` replaced by `-`), which is not an _append_ of `/some`, `/some` would not be marked as active.
+
+If `exact` is supplied, the link matches only if it matches the `record` matched by current location and link component **as well as** query + hash params.
+To allow an `exact` match to also allow different query params and hash params, we could allow a string value for the `exact` match so we don't break existing behavior:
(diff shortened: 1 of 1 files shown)


Akryum:
What if we provide a `active` prop instead that gets a function so that the user can customize the behavior however he wants?

posva:
it could be an extra option, indeed, but I wonder if it's necessary since it could also be done through the scoped slot (although it depends on what arguments you want to pass to the function):

```js
const isExact = (current, location) => location.path === current.path
```

```vue
<router-link v-slot="{ location }">
  <a :class="matchLocation($route, location) && 'is-active'">Link</a>
</router-link>
```

Akryum:
Yeah I think it should all be achievable with the slot API, and we should make sure it's easy to match several parts of the location. Then we could remove the `exact` prop altogether? :cat2:

```html
<router-link v-slot="{ location, navigate, matched }">
  <a
    :href="location"
    :class="{ 'router-link-active-exact': matched.path && matched.query }"
    @click="navigate"
  >
    Link
  </a>
</router-link>
```

Akryum:
I feel this new `extact="path+query"` syntax is not necessary with the slot API. To much new API surface and learning cost for little benefit imo.

posva:
I agree it can all be achieved through the scoped slot haha but it does make it much easier to use, so I don't think we should remove it altogether hehe. The most common use would be `exact="all"` and `exact="path"`, so removing the `+` syntax and only allowing those values seems like a fair trade off to me while we cover most use cases and people can still match only the query or the hash by using the scoped slot

Akryum:
So instead of having a new way of using `exact`, we could have a new boolean prop named `exact-path`:

```html
<router-link to="/" exact-path>Home</router-link>
```

Akryum:
In the end, I think it would be less confusing and easier to learn that having different ways of using `exact`.

posva:
yes, we could but because it's either using `exact` or `exact-path` so it would be confusing (it's my point in the alternatives, but I think you saw it), it means they should be refactored into one prop for it to be cleaner. It's also not possible to extend to new usages while the string version is.
To me having a new possible value is less to learn than having two props

Akryum:
Anyway, I think having this:
image
means the API is too complicated.

Akryum:
Maybe we could just remove the `exact` prop and make the user rely on the default classes added by `router-link`, like `.router-link-active-exact` and maybe a new one,  `.router-link-active-exact-path`?

Akryum:
So instead of having an `exact` prop, you could do:

```html
<router-link class="exact" to="/">Home</router-link>
<router-link to="/foobar">Foobar</router-link>
```

```css
:not(.exact).router-link-active,
.exact.router-link-active-exact {
  background: green;
}
```

posva:
About the table

It's funny, I felt the same way when I wrote it, but it should be treated as examples not as a table that would be included in the documentation. I realized it doesn't feel that complicated if I ignore the table and read it as a sentence that says we can match some specific parts of a location only and match multiple ones with a `+` operator

> router-link, like .router-link-active-exact and maybe a new one, .router-link-active-exact-path

that's another alternative I forgot to add but it's more complicating and also adds even more props to the router-link (active-link-class companions). Also, if we have one for path, why not query and hash? That's what people could ask for and that would create even more props

Akryum:
>  That's what people could ask for and that would create even more props

Then they can just use the slot API.

I really feel this is all overkill, and that the `exact` prop is kinda only handy for the `/` path. The more complex use cases should be covered by the slot API. The `exact` prop is currently necessary because we don't have this slot API.

Akryum:
In the end I think we should just leave the `exact` prop as it is, being mostly useful for `to="/"` and use the slot API which *will* cover all possible use case the `exact` prop may not.

Akryum:
Another possible change: remove `exact` prop and change the default behavior of the `router-link-active` class for the `/` path, so it's like `exact` by default. I mean in how many app you don't put `exact` on the `home` link in the nav?

posva:
> In the end I think we should just leave the exact prop as it is, being mostly useful for to="/" and use the slot API which will cover all possible use case the exact prop may not.

It's a possibility but there is a real need for matching the path only without the query and the hash. Since using a scoped slot will always be a lot more work than using the regular slot api, I think we should allow at least `exact="path"` for convenience. TBH the rest is really there as a proposal to gather feedback if this is necessary and I don't like the `+` because it's magic.

```css
:not(.exact).router-link-active,
.exact.router-link-active-exact {
  background: green;
}
```

I didn't think about this one! It's a good one. I still want framework/library users to be able to provide a single class for `.router-link-active` (in tailwindcss for example) and provide the `exact` prop to have their styling without writing CSS

Akryum:
I just though about an issue with the proposal: what if you want to have styling for let's say `router-link-active-exact` and `router-link-active-exact-path` (both of them)?

posva:
> I just though about an issue with the proposal: what if you want to have styling for let's say router-link-active-exact and router-link-active-exact-path (both of them)?

Do you mean all of the them at the same time, different styling?

the way I was thinking about the exact prop is:

- no prop -> apply active and active-exact separately
- `true` -> apply active **and** active-exact at the same time while the fullPath matches
- `'path'` -> apply active **and** active-exact at the same time while the path matches

Akryum:
> Do you mean all of the them at the same time, different styling?

Yes

posva:
Closing in favor of a much more simple RFC https://github.com/vuejs/rfcs/pull/37/files

PR #36, opened 2019-05-10
Exact should default to `true` for `/` path on router-link

Akryum:
[Rendered](https://github.com/Akryum/rfcs/blob/home-exact/active-rfcs/0000-home-exact.md)

### DIFF
--- a/active-rfcs/0000-home-exact.md
+++ b/active-rfcs/0000-home-exact.md
@@ -1,5 +1,5 @@
 - Start Date: 2019-05-10
-- Target Major Version: (2.x / 3.x) (Breaking)
+- Target Major Version: Vue 2.x/3.x, vue-router 4.x (breaking change)
 - Reference Issues: https://github.com/vuejs/vue-router/issues/454
 - Implementation PR:
 
--- a/active-rfcs/0000-home-exact.md
+++ b/active-rfcs/0000-home-exact.md
@@ -43,6 +43,7 @@ This is a breaking change.
 
 Migration:
 - Remove `exact` prop on router links targeting the root path.
+- For `<router-link>` that actively needs it, put `:exact="false"`.
 
 # Alternatives
 
--- a/active-rfcs/0000-home-exact.md
+++ b/active-rfcs/0000-home-exact.md
@@ -43,7 +43,7 @@ This is a breaking change.
 
 Migration:
 - Remove `exact` prop on router links targeting the root path.
-- For `<router-link>` that actively needs it, put `:exact="false"`.
+- For `<router-link>` on the root `'/'` path that actively needs it, put `:exact="false"`.
 
 # Alternatives
 
--- a/active-rfcs/0000-home-exact.md
+++ b/active-rfcs/0000-home-exact.md
@@ -1,0 +1,52 @@
+- Start Date: 2019-05-10
+- Target Major Version: (2.x / 3.x) (Breaking)
+- Reference Issues: https://github.com/vuejs/vue-router/issues/454
+- Implementation PR:
+
+# Summary
+
+Change default behavior of `router-link-active` class for the `'/'` path, so it's only applied on the `<router-link>` when the current location is exactly `"/"`.
+
+# Basic example
+
+Before:
+
+```html
+<router-link to="/" exact>Home</router-link>
+<router-link to="/about">About</router-link>
+```
+
+After:
+
+```html
+<router-link to="/">Home</router-link>
+<router-link to="/about">About</router-link>
+```
+
+# Motivation
+
+The big benefits of this change is to make the `.router-link-active` class just work out of the box for every router-link.
+
+Downsides of current implementation:
+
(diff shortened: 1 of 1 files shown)


dominik-bln:
What would happen on something like `<RouterLink
    :to="{name:'home'}">`?

What would happen if the application is hosted at `/my-app/`?

Akryum:
@dominik-bln The resolved path in vue-router would still be `/`.

ycmjason:
I personally prefer things with less exceptional cases and make things more explicit. I agree that `/` should always be `exact` but adding this exceptional case could be confusing for developers.

Instead, I think the correct way to solve this problem is to have all `router-link` to be `exact: true` by default and let user to specify something like `not-exact`. (need to think of a better name for that attr)

posva:
> An alternative way to solve this problem is to have all router-link to be exact: true by default and let developer specify something like not-exact. (need to think of a better name for that attr) I think it is safer to assume routes to always be exact than assuming that each route would have some sub-paths which I wish to add active to.

I also thought about that but the most common use case for active links is navbars and drawers, sometimes with submenus. In those scenarios, you want the active class to be applied while in a subroute. The only case that does not work by default is the `/`. While it is an exception, it isn't something people would have to think about because a `/` that isn't exact is **always** active, it's pretty much the same as having a custom class that is always applied, which seems way more intuitive than relying on the active class to add css

In other words, we want to make apply exact for `/` because that's what we always want, so in most scenarios, people won't even need to use `exact` at all

blake-newman:
I personally agree with @ycmjason 

Adding a different behaviour for a specific use case will lead to more confusion, although there is often a use case that `/` should have exact, it may be not the case. To reverse the default behaviour there will be two input structures:

`<router-link to="/" :exact="false">`
`<router-link to="/path" exact`>`

Without reading the documentation it will not be immediately obvious why there is two different behaviours, although reading the documentation is probably advised if you see unintended behaviour the API should be as intuitive as possible to not require the need for reading the documentation. Thus having the default the same for both cases is preferred IMO. 

I would argue that the default value for exact should be true, however that seems like a subjective default value. It is an assumption that we know the majority use case, and either default will likely result in someone not being happy.

posva:
> I would argue that the default value for exact should be true, however that seems like a subjective default value. It is an assumption that we know the majority use case, and either default will likely result in someone not being happy.

Why would you have `exact` default to true though?

the idea here is really to make things less confusing by eliminating the need to use `exact` except for advanced users, which would have to search the docs for it

blake-newman:
I believe that there is no perfect default in this case. 

However, based on my expectations on first usage I perceived that it would be more intuitive to match the exact path provided, as i would expect to have to provide a prop to allow nested routes to match.

Thus being subjective, as I believe every developer will have separate opinions on what the default behaviour should be. Personally, i would prefer to switch on nested route matching. 

I also believe that this proposal is biased to the use case of a `<nav>` (although an often use case). The use of `<router-link>` in the `<nav>` use case make sense for nested route matching by default. However the opposite can be said in use cases where `<router-link>` is used in places such as content. In this use case _exact_ matching would be preferred, thus making the default value subjective to how the router link is used.

ycmjason:
> I also thought about that but the most common use case for active links is navbars and drawers, sometimes with submenus. In those scenarios, you want the active class to be applied while in a subroute. 

I agree that in navbars/drawers, it is common that we wish to match subroutes. However, it is almost the only place I can think of where we would match subroutes, I would use `exact` in almost every other scenario.

posva:
> However, based on my expectations on first usage I perceived that it would be more intuitive to match the exact path provided, as i would expect to have to provide a prop to allow nested routes to match.

To be clear, there are two classes applied by default, one for active and one for exact

> is used in places such as content. In this use case exact matching would be preferred, thus making the default value subjective to how the router link is used.

Why would you mark links in content with custom style when active? That is much more specific use-case than the navbar and drawer though

> I would use exact in almost every other scenario I can think of. (e.g. breadcrumbs, body etc.)

breadcrumbs is actually another use case of the exact=false, what is the _body_ scenario? I wouldn't mind seeing other scenarios hehe

ycmjason:
> breadcrumbs is actually another use case of the exact=false

Why would you want exact=false in breadcrumbs?

Imagine we have a route `/orgs/vue/members/posva`, the breadcrumbs might show as:

```
Orgs > Vue > Members > Posva
```

In this case, I wouldn't expect `Members` to have the active class.

posva:
I see what you mean now, I was thinking of a deeper breadcrumb, showing further entries that are not active, like being in _Members_ while showing all the way til _Posva_
For that case, you would use the exact-active class instead though

Akryum:
Remember this proposal only concerns the router-link-active class, which is
broken by default for `/`. Currently, the exact prop is to work around this
which is something the new comer has to learn about because router-links
don't behave as expected in his menu.

posva:
This is closed in favour of https://github.com/vuejs/rfcs/pull/136

PR #37, opened 2019-05-14
Change exact prop to only match the path part by default

posva:
[Rendered](https://github.com/posva/rfcs/blob/router/link-exact-prop/active-rfcs/0000-router-link-exact-active.md)

Do you rely on `exact` to match the query and hash? If so for what kind of links? How do you match only the path in those cases


posva:

posva:
This is closed in favour of https://github.com/vuejs/rfcs/pull/136

PR #38, opened 2019-05-23
VueRouter - Allow scrolling other than window

phouri:


pimlie:
Would using [scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) be an option as a way to solve the first drawback about old browser quirks?

phouri:
ScrollIntoView does something else afaik - it scrolls the parent of a child into view, and does not scroll the element itself.

pimlie:
This should work:
```js
if (isObject && typeof shouldScroll.selector === 'string') {
    const el = document.querySelector(shouldScroll.selector)
    if (el) {
      el.scrollIntoView()
    }
```

we'd only need to add additional logic for `isPop`

phouri:
It's not just scrolling into view - it's keeping the last position, scrolling element can be huge - thousands of pixels, scrolling to the top of it (or center or any other option) - is not the required behavior, it should be where it was before navigation.

pimlie:
Hmm, yes. Isnt that actually managed by `saveScrollPosition` anyway? The position calculated in `scrollToPosition` doesnt seem to be used for that.

phouri:
save scroll position saves it only from the window, which is in some cases not the scroller. 

Check out my draft PR for a full example - it saves the new position and uses it on the new element.

PR #39, opened 2019-05-23
Proposal for better v-for

ycmjason:
[Rendered](https://github.com/ycmjason/rfcs/blob/master/active-rfcs/000-simplified-v-for.md)

### DIFF
--- a/active-rfcs/000-simplified-v-for.md
+++ b/active-rfcs/000-simplified-v-for.md
@@ -1,0 +1,128 @@
+- Start Date: 2019-05-23
+- Target Major Version: 3.x
+- Reference Issues: N/A
+- Implementation PR: 
+
+# Summary
+
+- Make `v-for` to behave more similar to  `for` and `Array#forEach` in 
+  Javascript.
+- Remove the "magic" in `v-for`.
+
+# Basic example
+
+```html
+<!-- iterating over arrays -->
+<div v-for="x of xs"></div>
+<div v-for="(x, i) of xs"></div>
+
+<!-- iterating over object keys -->
+<div v-for="key in object"></div>
+<div v-for="key of Object.keys(object)"></div>
+
+<!-- iterating over objects -->
+<div v-for="[key, value] of Object.entries(object)"></div>
+<div v-for="([key, value], i) of Object.entries(object)"></div>
+<div v-for="value of Object.values(object)"></div>
+<div v-for="(value, i) of Object.values(object)"></div>
+
+<!-- iterating through range -->
+<div v-for="(_, i) of Array.from({ length: 10 })"></div>
+```
+
+# Motivation
+
+In Vue 2.x, `v-for` does some "magic" to make iterating over object / 
+array easy for developers. I believe the "magic" was added because 
+`Object.entries` and `Object.values` did not exist or in an unstable 
+stage when `v-for` was first implemented. 
+
+By unifying the API for `v-for` and Javascript's `for` / `forEach`, we 
+make developer's life easier because:
+
+1. For complete beginners, they only need to care about one way of
+   `for`-looping.
+2. For Javascript developers who are new to Vue, they will feel at home
+   when using `v-for`.
+
+# Detailed design
+
+Implement `v-for` so that it maps directly to `for`-loop / `forEach` in 
+Javascript.
+
+I refer to the following as "magic" because they do not behave the same
+as Javascript's `for`. I believe they should be removed in Vue 3.x:
+
+1. Iterating over objects with `for-in` (`v-for-in-object`)
+```html
+<div v-for="(value, key, index) in object"></div>
+```
+
+2. Iterating over array with `in` (`v-for-in-array`)
+```html
+<div v-for="x in xs"></div>
+```
+
+3. Iterating over number (`v-for-in-number`)
+```html
+<div v-for="i in 10"></div>
+```
+
+# Drawbacks
+
+- Upgrading from Vue 2.x requires a lot of changes depending on how
+  heavily `v-for-in-object` and `v-for-in-array` and `v-for-in-number`
+  are used throughout the project.
+- More verbose when using `v-for`.
+
+# Alternatives
+
+We can even go further and remove the `index` passed as the second
+argument to make the `v-for` exactly the same as Javascript's `for`.
+
+For example:
+```html
+<!-- iterating over arrays -->
+<div v-for="x of xs"></div>
+<div v-for="([x, i]) of xs.entries()"></div>
+
+<!-- iterating over object keys -->
+<div v-for="key in object"></div>
+<div v-for="key of Object.keys(object)"></div>
+
+<!-- iterating over objects -->
+<div v-for="[key, value] of Object.entries(object)"></div>
+<div v-for="([[key, value], i]) of Object.entries(object).entries"></div>
+<div v-for="value of Object.values(object)"></div>
+<div v-for="([value, i]) of Object.values(object).entries()"></div>
+
+<!-- iterating through range -->
+<div v-for="i of Array.from({ length: 10 }, (_, i) => i)"></div>
+```
+
+
+# Adoption strategy
+
+To adopt to this simplified `v-for`, we will need to only deal with the three
+cases mentioned.
+
+1. Iterating over objects with `for-in` (`v-for-in-object`)
+```html
+<div v-for="(value, key, index) in object"></div>
+// This will become
+<div v-for="([value, key], index) of Object.entries(object)"></div>
+```
+
+2. Iterating over array with `in` (`v-for-in-array`)
+```html
+<div v-for="x in xs"></div>
+// This will become
+<div v-for="x of xs"></div>
+```
+
+3. Iterating over number (`v-for-in-number`)
+```html
+<div v-for="i in 10"></div>
+// This will become
+<div v-for="i of Array.from({ length: 10 }, (_, i) => i + 1)"></div>
+```
--- a/active-rfcs/000-simplified-v-for.md
+++ b/active-rfcs/000-simplified-v-for.md
@@ -7,7 +7,8 @@
 
 - Make `v-for` to behave more similar to  `for` and `Array#forEach` in 
   Javascript.
-- Remove the "magic" in `v-for`.
+- Introduce `v-for-from` for iterating through objects and range.
+- Introduce range syntax `..`.
 
 # Basic example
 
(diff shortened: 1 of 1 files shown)


active-rfcs/000-simplified-v-for.md
```diff
@@ -0,0 +1,128 @@
...
+# Adoption strategy
+
+To adopt to this simplified `v-for`, we will need to only deal with the three
+cases mentioned.
+
+1. Iterating over objects with `for-in` (`v-for-in-object`)
+```html
+<div v-for="(value, key, index) in object"></div>
+// This will become
+<div v-for="([value, key], index) of Object.entries(object)"></div>
```
tlgreg:
```suggestion
<div v-for="([key, value], index) of Object.entries(object)"></div>
```

ycmjason: thank you!

leopiccionia:
May a benchmark prove me wrong, but current behavior (see [implementation](https://github.com/vuejs/vue/blob/6fe07ebf5ab3fea1860c59fe7cdd2ec1b760f9b0/src/core/instance/render-helpers/render-list.js)) seems faster than proposal.

Notably, `i of 10` is not only much terser than `i of Array.from({ length: 10 }, (_, i) => i)`, but more efficient, as it compiles to a plain *numeric `for`* that can be further optimized by JIT compiler (and even inlined). Perhaps JIT compilers are smart enough to detect this "range" pattern, too?

ycmjason:
@leopiccionia 

I agree. I would imagine the current implementation to be faster. I have added this drawback to the proposal.

I understand that the performance of the render function can do a significant impact on the application. There are some ways which we could reduce the impact. For example, we can make the expression in `i {of,in} [expr]` a computed value under the hood. Details to be discussed.

I believe the main focus of this proposal is to remove the not-too-javascripty things in the current `v-for`.

ycmjason:
For `i in 10`, maybe we can provide a special `range` generator function in vue template which will return a generator from `1` to `10`.

It could be as simple as 

```js
function* range(n) {
  let i = 0;
  while (i <= n) yield i++;
}
```

Or provide more functionality like python's range.

yyx990803:
I like the general direction of this (aligning more with JavaScript), but there are a few issues I am concerned about:

- performance of raw `for...of` + `Object.entries()` is [significantly slower](https://jsperf.com/of-entires-vs-manual) than `for...in` + `object[key]`. The perf difference is so big that may make this a deal breaker. **We have to provide constructs that are fast by default.**

- Code is also much more verbose with `Object.entires()`.

I think there could be a middle ground where the only change is respecting the difference between `in` and `of` (which works interchangeably right now):

- `v-for` with `in` always iterate with keys, so

  ``` vue
  <div v-for="index in arr">
  <div v-for="key in obj">
  ```

  both would be consistent with JavaScript.

- `v-for` with `of` always iterate with values:

  ``` vue
   <div v-for="value of arr">
  ```

  For objects, we provide a sugar so that there's no need for explicit `Object.entries()`:

  ``` vue
   <div v-for="value of obj">
  ```

  This allows the code to be less verbose and gives us the opportunity to use a more efficient internal implementation.

  Accessing index or key with `of` works just like now:

  ``` vue
  <div v-for="(value, i) of arr">
  <div v-for="(value, key, i) of obj"> 
  ```

- Range sugar should be kept, but only with `of`:

  ``` vue
  <div v-for="i of 10" />
  ```

  (maybe also take the opportunity to make it start from `0`?)

---

Migration strategy would also be straightforward: just change all `v-for="... in ..."` to `v-for="... of ..."`.

ycmjason:
@yyx990803  

I agree with most of your points and understand that we should provide the fast implementation by default. 👍

Writing `Object.entries` does not necessarily mean that the templates is compiled down to actual `Object.entries`. I imagine the compiler could compile `for-of-entries` down to `for-in` loops which could potentially solve the problem with performance. The important point IMO is to make the behaviour of `v-for-in` and `v-for-of` as consistent to Javascript's as possible.

It feels strange to use `of` on an object/number as this is not allowed in Javascript. I would suggest choosing another keyword/symbol to avoid mixing up with Javascript.

Here are some alternatives to your suggestions:

- Object could be done with special `entries` function
```html
<div v-for="entry of entries(obj)"></div>
<div v-for="([key, value], i) of entries(obj)"></div>
```

- Object could be done with `entriesOf` keyword
```html
<div v-for="(key, value) entriesOf obj"></div>
```

- Object could be done with `from` keyword
```html
<div v-for="(key, value) from obj"></div>
```

- Range could be done with `v-for-from-to-by` (similar to [PL/I](https://en.wikipedia.org/wiki/For_loop#1964:_PL/I)'s)
```html
<div v-for="i from 1 to 5"></div>
<div v-for="i from 1 to 5 by 2"></div>
```

- Range could be done with a special `range` function (similar to [python](https://en.wikipedia.org/wiki/For_loop#1991:_Python)'s)
```html
<div v-for="i of range(1, 5)"></div>
<div v-for="i of range(1, 5, 2)"></div>
```

- Range could be done with range syntax `..`

```html
<div v-for="i from 1..5"></div>
<div v-for="i from 1,5..15"></div>
```

Thoughts?

yyx990803:
@ycmjason 

- Any magic helper functions (`entries`, `range`) causes potential conflict with user methods with the same name.

- I like the `from` keyword. The potential drawback is the potential conflict if JavaScript one day adds the `for ... from` syntax itself, but I think it might be worth it.

- For range, I like the `..` syntax, but it's not JavaScript. Users could potentially write things like `a..(some + expression)` and this incurs extra complex parsing edge cases. I think for range we should just keep it simple.

leopiccionia:
It's important to be aware that any "helper function" like `entries` and `range` means more "magic", not less, as it's additional microsyntax. `entries` is particularly likely to clash with userland code (I may have used it already myself).

I really like the compromise on [Evan's first comment](https://github.com/vuejs/rfcs/pull/39#issuecomment-495463470).

mika76:
I must say I don't see the benefit of changing this. `v-for` is currently nice and clean, easy to write, easy to read and understand. I understand this would make it closer to js but then again this is template code so why would it _need_ to be closer to js? And all those functions already work so changes here seem totally unnecessary. Just another hurdle from upgrading from vue 2.x...

victornpb:
I'm favorable to adding range looping. I dislike the idea of creating dummy arrays in order to iterate.
```js
for(let i=a; i<b; i++)
```
```html
<template v-for="i=a; i<b; i++">
    <div>{{i}}</div>
</template>
```

For looping through arrays or objects, I find it extremely convenient the way it works right now, I don't see the need for change specially if it breaks current syntax. Templates are one of the hardest things to migrate over a large code base.

CarterLi:
> Code is also much more verbose with Object.entires()

Exactly you may always need to wrap for..in loop with `hasOwnProperty`

```js
for (const key in object) {
  if (Object.prototype.hasOwnProperty.call(object, key)) { // I don't use `object.hasOwnProperty` because `hasOwnProperty` can be overrided or the object can be created by `Object.create(null)`
    sideEffect = [key, object[key]]
  }
}
```

VS

```js
for (const [key, value] of Object.entries(object)) {
  sideEffect = [key, value]
}
```

MarcelloTheArcane:
One thing that would be nice to get rid of is the obligatory `:key="..."`. 

Correct me if I'm wrong, but isn't it quite boilerplate? (Or is there a use for it?)

leopiccionia:
@MarcelloTheArcane `:key` is a bit of context passed to Vue's diff algorithm, so it can preserve state between each re-render.

While it's not strictly required in Vue 2.x, omitting it is not suitable for all use cases: https://vuejs.org/v2/api/#key.

sontd-0882:
The change from `v-for="... in ..."` to `v-for="... of ..."` is not bring much benefit. The current syntax is nice enough, easy to write and understand.

ycmjason:
@sontd-0882 I would argue that it is not beneficial if you are already familiar with the current `v-for-in`. In the long run, having `v-for` more consistent with JS would help new vue users to understand `v-for` much easily.

frlinw:
About performance, I don't know if it's relevant but vue compiler can probably optimize loops
```js
const obj = {};
for (let i = 0;i<0xf;++i) {
  obj[i] = i;
}

console.time('Object.keys + forEach');
for (let i = 0;i<0xFFFFF;++i) {
  Object.keys(obj).forEach(k => {
    const value = obj[k];
  });
}
console.timeEnd('Object.keys + forEach');

console.time('for of + Object.entries');
for (let i = 0;i<0xFFFFF;++i) {
  for (const [key, value] of Object.entries(obj)) {
    // value
  }
}
console.timeEnd('for of + Object.entries');

console.time('for in + hasOwnProperty');
for (let i = 0;i<0xFFFFF;++i) {
  for (const k in obj) {
    if (obj.hasOwnProperty(k)) {
      const value = obj[k];
    }
  }
}
console.timeEnd('for in + hasOwnProperty');
```
Chrome 76
image
Firefox 68
image
Result:
- `for in + hasOwnProperty` is the slowest on Firefox
- `Object.keys + forEach` is the fastest on Firefox & Chrome

Feel free to try this code in the console on others browsers

CarterLi:
@frlinw What about Object.entries + forEach?

frlinw:
@CarterLi same as `for of + Object.entries` in Chrome, close to `Object.keys + forEach` in Firefox. previous post updated

JamesCoyle:
I think the only real thing I like about this proposal is the range syntax as having do deal with the 1 to X ranges gets really annoying the more you need to use that value. 

I think the way I would solve this though is to allow for user defined generator functions in the current range syntax so `index in range(0, 9)` where `range` is a generator function that returns all the numbers between 0 and 9, incrementing by one each time. If the generator function then `return`s once the limit is reached then vue can detect that the result has completed. This would also allow for custom sequences as needed such as exponential growth, odds/evens, reverse ranges etc. without having do do the math everywhere you need to use that value.

leopiccionia:
Vue support custom iterators inside `v-for` since 2.6. See: https://jsfiddle.net/leopiccionia/vg7e6fj1/

JamesCoyle:
Neat! Have I missed some documentation somewhere?

leopiccionia:
> Neat! Have I missed some documentation somewhere?

As far as I know, it's only documented here: https://vuejs.org/v2/api/#v-for

michaeldrotar:
Sounds like this rfc could be split into smaller pieces.. it's hard to differentiate which parts people like vs what they don't like.

Personally while the _idea_ of being closer to js sounds good, I think most of the ideas are not supporting that goal.

These syntaxes don't have an equivalent syntax in js (afaik and as far as trying them in console shows):
```vue
<div v-for="(x, i) of xs"></div>
<div v-for="(key, value) from object"></div>
<div v-for="(key, value, i) from object"></div>
<div v-for="i from 1..10"></div>
<div v-for="i from 1,5..10"></div>
```

Only these syntaxes meet that goal:
```vue
<div v-for="x of xs"></div>
<div v-for="key in object"></div>
```

And while it _sounds_ good to match js 1:1, I think the utility of `(value, key) in object` and other such patterns is worth deviating from js.

---

As to the need for a range operator.. that sounds like a completely separate concern, and it sounds worthwhile, but it sounds like a user lib would be a better solution so it could answer things like how to specify a `start` and a `step` rather than be a loop syntax.

I would also wonder.. in real-world places where a range operator would be useful.. would a `render` function solve things better? They're not too complex, can be limited to a small component, and can have all the flexibility of js.

zhangenming:
> One thing that would be nice to get rid of is the obligatory `:key="..."`.
> 
> Correct me if I'm wrong, but isn't it quite boilerplate? (Or is there a use for it?)

how about
`<li v-for="num in 10, num">{{num}}</li>`
replace
`<li v-for="num in 10" :key="num">{{num}}</li>`

yyx990803:
Closing as Vue 3 has shipped and we are no longer considering breaking changes at this point.

PR #46, opened 2019-06-17
Make props required by default

afontcu:
* Props are required out of the box.
* Providing a `default` key sets the prop as optional.

[Rendered](https://github.com/afontcu/rfcs/blob/master/active-rfcs/0000-props-required-by-default.md).

### DIFF
--- a/active-rfcs/0000-props-required-by-default.md
+++ b/active-rfcs/0000-props-required-by-default.md
@@ -1,0 +1,94 @@
+- Start Date: 2019-06-17
+- Target Major Version: 2.x / 3.x
+- Reference Issues:
+- Implementation PR:
+
+
+# Summary
+
+Make component props required out of the box. Infer optional props from the already-existing `default` key.
+
+# Basic example
+
+```js
+const Component = {
+  props: ['title'], // required prop
+}
+```
+
+```js
+const Component = {
+  props: {
+    // required prop
+    title: String,
+
+    // required prop
+    subTitle: {
+      type: String
+    },
+
+    // optional prop
+    description: {
+      type: String,
+      default: 'My default value'  
+    }
+  }
+}
+```
+
+
+# Motivation
+
+There are two main motivators behind this proposal.
+
+Firstly, current prop declaration API allows users to set illogical scenarios, such as providing a default value for a required prop. There's even a specific [`Vue Eslint rule`](https://eslint.vuejs.org/rules/require-default-prop.html#vue-require-default-prop) to warn developers about it.
+
+Secondly, this proposal lets users write a shorter, terser syntax to define props by ditching the object notation and relying solely on the `name: Type` syntax while defining required props. From my experience both using and teaching Vue, setting a prop as required is the most common scenario, so this would reduce some boilerplate on Vue components.
+
+
+# Drawbacks
+
+People familiar with current implementation would need to unlearn current behavior.
+
+The `default` key would become somewhat "overloaded", as it would not only provide a default value, but also mark the prop as optional. I see this as a feature.
+
+
+# Alternatives
+
+A possible alternative is to keep using the `required` attribute and let user use it to explicitly express the intended behavior:
+
+```js
+const Component = {
+  props: {
+    // required prop
+    title: String,
+
+    // optional prop
+    subTitle: {
+      type: String,
+      required: false,
+    },
+
+    // optional prop
+    description: {
+      type: String,
+      required: false,
+      default: 'My default value'
+    }
+  }
+}
+```
+
+
+# Adoption strategy
+
+A codemod could remove all unnecessary `required: true` statements.
+
+After that, and depending on the prop declaration syntax used on each codebase, developers would need to fix missing required prop warnings.
+
+
+# Unresolved questions
+
+> Is this a breaking change?
+
+While this proposal changes the public API (and thus it could be considered a breaking change), existing applications shouldn't break - they'd only throw additional console warnings.


leopiccionia:
> The default key would become somewhat "overloaded", as it would not only provide a default value, but also mark the prop as optional.

@afontcu It can be ambiguous, e.g. `{ default: undefined }` means that the prop is required or that the default value is `undefined`? Is it possible to distinguish both?

```javascript
const Component = {
  someProp: { default: undefined },
}
```

thecrypticace:
It's not ambiguous because the `in` operator can always tell you if a default property exists.

```js
"default" in {} // false
"default" in {default: undefined} // true
```

leopiccionia:
Good point. The `in` operator can solve this ambiguity and, considering that the prototype chain would generally have three or less levels, still be efficient.

These semantics (implicit vs explicit `undefined`) could have some gotchas in very rare contexts (e.g. some kinds of HOC), but still be bearable.

```javascript
prop = { type, ...(hasDefault ? { default } : {}) }
```

CyberAP:
Please list in the drawbacks section that it's a breaking change, and if you have an established codebase with a lot of optional props you'll have to rewrite all of these props to match the new logic.

ycmjason:
@leopiccionia

I have never seen anyone would write a dynamic `prop` declaration. I would like to know in what situation would that be useful. 🧐

afontcu:
@CyberAP hi and thanks for your feedback!

It is true, thou, that depending on your prop declaration strategy you might need to deal with missing required props warning. Yet, AFAIK, nothing should break (that's why I added the "is it a breaking change?" question in the Unresolved questions section).

As long as props have a `default` key, you'd only need to remove the `required: true` attribute, as it wouldn't be necessary (even though nothing would break).

If you have been using the vue eslint rule that asks you to provide a default value for optional props, you should be good to go.


Again, thanks for your time!

donnysim:
So far I've never seen `required` go with `default`, there's just no point, if you add `default` then you expect it to be optional so I don't see object syntax as a breaking change, now on the other hand, the `props: ['name', 'disabled']` is a breaking change and a moderate one. Having to fallback to object syntax just because 1 from 6 props should be required is inconvenient. I'd suggest keeping the list format as optional while object could be required based on default presence.

filrak:
@afontcu a lot of people rely on empty optional props falsy values (a simple example would be `{{ optionalProp ? ' a' : 'b' }}`  By implementing this feature you'll force them to explicitly declare those false values which in many situations can be significant amount of work. Not to mention many component libraries that will producing the errors. Ofc it's something easily doable via eslint plugin but you can't expect everyone to use it.

Even though it's a nice thing from the code quality PoV I think it's unnecessary breaking change.

gigioSouza:
Nice proposal, I'd rather mark a prop as optional instead of required. As I always set a default value for optional props, I think it's needless to mark others as required...

@donnysim possible workaround for list format optinal markup would be prepend or append a `?`.
For example: 
`props: ['name', '?disabled']` or `props: ['name', 'disabled?']`

casey6:
> So far I've never seen required go with default, there's just no point, if you add default then you expect it to be optional so I don't see object syntax as a breaking change

I _always_ specify `required: false` when adding a `default` value. If I'm quickly looking through a component the following props looks exactly the same to me.

```
props: {
  propA: {
    type: Boolean,
    default: true
  },
  propB: {
    type: Boolean,
    required: true
  }
}
```

YunSun-CN:
It might be debatable.
If we dig into some specs (html or web components), it found no attributes (or say property) is required. Each attribute has its default value. All required values should be validated in component's program scripts. In according to other props, the prop whether is required may be differ. And just throw an error from your program scripts may be better.

Consider if we want to make a component to display user's profile, and it may be rendered different according to its prop `type`: `basic` type only renders user's profile image, or `complex` type renders user's name in addition.
```
{
  props : {
    id    : { type : Number }, // its default value is undefined
    name  : { type : String }, // its default value is undefined
    type  : { type : String, default : `simple` },
    value : { type : String, default : `` },
  },
  watch : {
    type : {
      handler ( type ) {
        if ( `simple` !== type && ! this.name ) {
          // Provide a user-friendly custom error message here
          const error = new Error( `A name is required if type is not simple` );

          // Provide a custom error name to catch and resolve outside (for analysis etc. or passthrough)
          error.name = `MyComponentError`;

          throw error;
        }
      },
      immediate : true,
    },
  },
}
```
This solution handles two different types in one components instead of using a functional component to combine two components.

And if we could use typescript (or flow), it may be more simple.
```
interface MyProps {
  id    ?: number,
  name  ?: string,
  type   : string,
  value  : string,
}

{
  props : {
    type  : `simple`,
    value : ``,
  } as MyProps,
}
```

To coerce, could be written as
```
interface MyProps {
  value : ( value : any ) : string[] => {},
}

{
  props : {
    value : val => [ String( val ) ],
  } as MyProps,
}
```
Now, all complex value types such as Array or Object could be defined as a transformation.
**Note** It should make a proxy to provide coercing like computed values. Thus, we could retrieve the original value by `propsData.value` or `Reflect.get( props.value )`

Obviously, validate options can be moved from `created` hook to coerce format.
```
{
  props : {
    value : ( val, ctx ) => {
      if ( ! ctx.props.id || ctx.propsData.length ) {
        // Throw an error here
      }
      const arr = new Array( ctx.propsData.length );
      arr[ id ] = String( val );
      return arr;
    },
  } as MyProps,
}
```

A Vue 3 Syntax
```
import { createComponent, createElement, value, computed } from "vue";

interface MyProps {
  id     : ( value ?: number | string ) : string => {},
  length : number,
  value  : ( value : any, ctx ?: VueComponent ) : string[] => {},
}

const MyComponent = createComponent( {
  props : {
    id     : ( id = `` ) => String( id ),
    length : 1,
    value  : ( val, ctx ) => {
      if ( ! val ) {
        throw new Error( `[MyComponent Prop Error] Please provide a non-empty string value` );
      }
      const arr = new Array( ctx.props.length );
      arr[ ctx.props.id ] = String( val );
      return arr;
    },
  } as MyProps,
  setup ( props ) {
    props.value // A proxied value
    Reflect.get( props.value ) // Original propsData

    // Observed data
    const data = value( { key : `chunk` } );
    // Computed value
    const hashValue = computed( props => ( { [ data.key ] : props.value } ) );

    return ( props, slots, attrs, vnode ) =>
      createElement( `div`, JSON.stringify( hashValue ) );
  },
} );
```

Cheers🎉

yyx990803:

Side note: I originally was in favor of this idea because it makes the type inference for required props a bit easier - however I just recently managed to get rid of the need to use `as const` for props declaration in order to correctly infer required props, so I believe this change will not be worth it anymore.

afontcu:
thank you for taking the time!

PR #69, opened 2019-07-01
Implicit return initial values

sebhochuli:
[Rendered](https://github.com/sebhochuli/rfcs/blob/implicit-return-initial-values/active-rfcs/0000-implicit-return-initial-values.md)


thecrypticace:
This would require a compilation step. Omitting the `return` in javascript means that the return is `undefined`. If you don't specify what you are returning then the Javascript engine doesn't know what you are trying to return and therefore returns nothing. These "hook" functions are just vanilla javascript.

In addition, I do not believe a compiler step enabling such a step would provide any meaningful benefit to one's workflow but it could hinder readability and understanding because javascript in general does not work this way.

posva:
This could be possible by making the functions (`reactive`, `computed`, etc) stateful but it causes problems:

- They are not generic usage functions that can be used outside of Vue
- They cannot be parallelized and couldn't be properly used in Web Workers in the future without using mutex or semaphores

This could be added to vue-loader but then the JS from the `script` section would not work without a compilation step, which is a gap we want to avoid

Finally, having the explicit return, also allows for switching different solutions easily while developing, by commenting out lines:

```js
setup () {
  const { mouseX, mouseY } = useMouse()
  const { eyeX, eyeY } = useEyeTracking()
  
  // we can swap comments to switch what values are used
  return {
    x: mouseX,
    y: mouseY,
    // x: eyeX,
    // y: eyeY,
  }
}

sebhochuli:
Thanks for your reply

Effectively, add a compilation step would be a problem and wouldn't respect the initial idea.

Have a nice day Eduardo.

On juil. 1 2019, at 8:52 am, Eduardo San Martin Morote <notifications@github.com> wrote:
This could be possible by making the functions (reactive, computed, etc) stateful but it causes problems:


  *
They are not generic usage functions that can be used outside of Vue
  *
They cannot be parallelized and couldn't be properly used in Web Workers in the future without using mutex or semaphores

This could be added to vue-loader but then the JS from the script section would not work without a compilation step, which is a gap we want to avoid

Finally, having the explicit return, also allows for switching different solutions easily while developing, by commenting out lines:


setup () {
  const { mouseX, mouseY } = useMouse()
  const { eyeX, eyeY } = useEyeTracking()

  // we can swap comments to switch what values are used
  return {
    x: mouseX,
    y: mouseY,
    // x: eyeX,
    // y: eyeY,
  }
}

bartlomiej-korpus:
Wouldn't it possibly hurt type inference in TypeScript in the future?

posva:
> Wouldn't it possibly hurt type inference in TypeScript in the future?

Yes, an implicit return will break type inference as well

phiter:
I don't think that's a good idea, because you could have reactive value that you don't want to expose to the template when using in a component.
With return you specify what you want to use on your template, or what you want to return from a reactive function.

posva:
Since this goes against one of the main aspect of the API: typing, let's drop it

PR #75, opened 2019-08-05
Prefer unknown over any when assigning types to props

arianrhodsandlot:
Some believe that `any` is evil in TypeScript, should we use a better type [`unknown`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type) here?


yyx990803:
The function api RFC is outdated. Regardless, this seems to not be a major issue as it's just a matter of preference.

PR #77, opened 2019-08-16
Add a description for dependency injection functions

sh7dm:
Add a description about `provide` and `inject` functions documenting their args.
This PR is also changing API a bit from what we have in https://github.com/vuejs/vue-function-api.

### DIFF
--- a/active-rfcs/0000-function-api.md
+++ b/active-rfcs/0000-function-api.md
@@ -617,7 +617,7 @@ const Ancestor = {
 
 const Descendent = {
   setup() {
-    const { [CountSymbol] } = inject(CountSymbol)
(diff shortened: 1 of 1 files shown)


sh7dm:
Maybe we could return the single value from `inject` if one key is given and object if array of keys is given.
```ts
function inject<T>(key: Key<T> | string): Wrapper<T>;
function inject(keys: (Key<any> | string)[]): UnwrapValue<AnyObject>;
```
maybe types are a bit wrong

yyx990803:
This seems to be submitted against the outdated proposal. Latest is here: https://vue-composition-api-rfc.netlify.com/

PR #80, opened 2019-08-20
Use slots on router-view to display an initial placeholder instead of an empty node

briwa:
TL;DR: There should be a way to render something on the browser, instead of an empty node, when `vue-router` has just been initialized.

[Rendered](https://github.com/briwa/rfcs/blob/initial-placeholder-for-vue-router/active-rfcs/0000-initial-placeholder-for-vue-router.md)

### DIFF
--- a/active-rfcs/0000-initial-placeholder-for-vue-router.md
+++ b/active-rfcs/0000-initial-placeholder-for-vue-router.md
@@ -15,7 +15,7 @@ between the blank page and the rendered component is undesirable and I am trying
 
 I tried using [async component](https://vuejs.org/v2/guide/components-dynamic-async.html#Async-Components) to no avail, and it turned out that the cause was when the router is initialized, it is validating an [initial route](https://github.com/vuejs/vue-router/blob/v3.0.6/src/util/route.js#L52), and it is rendering an empty node since there is [no match](https://github.com/vuejs/vue-router/blob/v3.0.6/src/components/view.js#L51). This explains the blank page/element that I was facing.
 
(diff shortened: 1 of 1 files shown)


michaeldrotar:
Sounds like a great idea. An issue I've faced is that you're still waiting for the app to load before it can do anything. What I've done instead is add the minimum css and html by hand to the index page for showing a splash screen or loading indicator and then once the initial route is rendered I'll hide them.

When using a loading indicator, it's generally also a component, but since I can't wait for the framework and my app to load, I have to copy/paste the necessary code from the component to the index page to get the fastest time to render so possibly a plugin would be really handy that could do this instead so that imports and sass and such things could still be used.

briwa:
Hello @posva, sorry for bumping this 3 months-old PR. Was wondering if you're ok with this RFC? So that I can proceed with https://github.com/vuejs/vue-router/pull/2799. Thank you.

posva:
@briwa Sorry, but we are not sure this is the direction we want to go yet. So please, don't add more commits to the PR. We will come back at you when we can

posva:
If you check `$route.matched.length` to know if there is a match, you could initially display something while the router has nothing to display:

```vue
<router-view v-if="$route.matched.length"/>
<loading-view v-else/>
```

Or by comparing to the initial location (https://github.com/vuejs/vue-router/issues/2718)
```vue
<router-view v-if="$route !== START"/>
<loading-view v-else/>
```

Apart from a builtin way to handle this, why would using the default slot better? Having it as a default makes it impossible to escape, which could be problematic for Declarative Routing (adding routes as we enter a View)

posva:
Closing for the reasons mentioned above and because this can be achieved with https://github.com/vuejs/rfcs/pull/153
It will likely be ported back to router v3 as well

PR #82, opened 2019-09-28
Immutable object when using computed get only

pikax:
[Rendered](https://github.com/pikax/rfcs/blob/immutable-computed-get/active-rfcs/0000-immutable-computed-get.md)

[codesandbox](https://codesandbox.io/s/vue-template-mwcpu)

### DIFF
--- a/active-rfcs/0000-immutable-computed-get.md
+++ b/active-rfcs/0000-immutable-computed-get.md
@@ -1,0 +1,109 @@
+- Start Date: 2019-09-28
+- Target Major Version: 3.x
+- Reference Issues: (fill in existing related issues, if any)
+- Implementation PR: (leave this empty)
+
+# Summary
+
+- Make `computed` only `get` to return an readonly immutable object.
+
+# Basic example
+
+```js
+{
+    data(){
+        return {
+            a: 1,
+            b: 2
+        }
+    },
+    computed: {
+        mixedObject(){
+            return {
+                a,
+                b,
+                c: a + b
+            }
+        }
+    },
+    methods(){
+        incrementA(){
+            //this.mixedObject.a++; // do not allow this
+            this.a++
+        }
+    }
+}
+```
+
+
+# Motivation
+
+Simplifying the api, allowing an readonly computed to be change will create a pitfall and not easy to spot bugs.
+
+Allowing to add properties to a readonly computed it will only work until the computed is refreshed.
+
+This is also motivated by the wording in the new `composition-api` rfc, since the read only computed are described as `immutable` ref object.
+
+# Detailed design
+
+Currently in 2.x you can do:
+```js
+export default {
+  data() {
+    return {
+      test: 1,
+    };
+  },
+  computed: {
+    awesome() {
+      return {
+        test: this.test,
+        test2: 1
+      };
+    }
+  },
+
+  methods: {
+    increment() {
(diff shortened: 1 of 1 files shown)


posva:
I don't think the cost of making every computed property immutable is worth since the concept of a computed property is precisely being readonly. Making the type use `Readonly`, sure, that's definitely a good thing as it has no cost

kahirokunn:
This idea will zero cost resolve many mistakes.
I think this idea is really cool.

pikax:
I'm using `immutable` wording to enforce the fact the readonly computed should be readonly through out, since modifying it leads to unexpected behaviour. 

Can't we warn in development if you try to modify a computed? For example prop type checking is only available in development mode

kahirokunn:
I think zero cost static analysis is good.
Because I think debug on the runtime with typescript is not typescript way ...😢

yyx990803:
In v3 this is the default behavior. In v2, however, implementing such behavior without native ES Proxy will be too costly.

pikax:
@kahirokunn not everyone uses typescript, having some safeguards (such runtime and development only checks) would be helpful for some developers.


> In v3 this is the default behavior. In v2, however, implementing such behavior without native ES Proxy will be too costly.

The wording used in composition-api-rfc suggested that.

Do you think is worth to add this interactions to the documentation as an `warning`?

for example: 
> Object returned by computed property are `non-reactive` and don't affect the original instance and it will be overrided once the computed gets recalculated
example:
```js
var vm = new Vue({
  data: { a: 1 },
  computed: {
    // get only
    double: function () {
      return {
        a: this.a
    },
  }
})
vm.a   // => 1
vm.double.a // => 1
vm.a++ // => 2
vm.double.a // => 2
vm.double.a++ // => 3
vm.a  // => 2
vm.a--  // => 1
vm.double.a  // =>1
```

danielelkington:
Sometimes it makes sense for properties in an object returned by a computed property to be mutable. Consider cases where a component's data contains an array of items and a computed property is used to point to a particular item in the array - properties can be safely mutated as you are really just changing the underlying data.

For example ([CodeSandbox](https://codesandbox.io/s/mutablecomputedproperty-h908v)):
```
<template>
  <div>
    <label>Lookup Person Id:</label>
    <input type="number" v-model="selectedPersonId">
    <br>
    <label v-if="selectedPerson">Edit Person Name:</label>
    <input v-if="selectedPerson" v-model="selectedPerson.name">
  </div>
</template>

<script>
export default {
  data: function() {
    return {
      people: [{ id: 1, name: "Jane" }, { id: 2, name: "Jack" }],
      selectedPersonId: 1
    };
  },
  computed: {
    selectedPerson: function() {
      return this.people.find(x => x.id === Number(this.selectedPersonId));
      // Safe to mutate this object (eg using v-model) because it just points
      // to something in "data"
    }
  }
};
</script>

```

pikax:
@danielelkington thanks for bringing that point, I think that behaviour is expected. 

Then reason for this rfc is to prevent non-reactive for being modified, for example change the type of the people to be an array of numbers, and you will see your change doesn't take effect. 

The definition would be: the `computed property` is **readonly** if the property value is `not` **reactive** or **ref**.

I will update the rfc to incorporate that too.

pikax:
https://github.com/vuejs/vue-next doesn't have this limitations.

PR #87, opened 2019-10-22
Rename `:key` (i.e. `v-bind:key`) to `v-key`

ThomasLandauer:
See https://github.com/vuejs/vue/issues/10583

### DIFF
--- a/active-rfcs/0000-rename-key-to-v-key.md
+++ b/active-rfcs/0000-rename-key-to-v-key.md
@@ -1,0 +1,37 @@
+- Start Date: 2019-10-22
+- Target Major Version: 3.x
+- Reference Issues: https://github.com/vuejs/vue/issues/10583
+- Implementation PR: (leave this empty)
+
+# Summary
+
+Rename `:key` (i.e. `v-bind:key`) to `v-key`
+
+# Basic example
+
+<li v-for="product in products" v-key="product.id">
+
+# Motivation
+
+The current syntax of `key` ( https://vuejs.org/v2/guide/list.html#Maintaining-State ) is inconsistent with the `v-bind` directive:
+https://vuejs.org/v2/guide/syntax.html#Attributes says that `<div v-bind:foo="...">` will result in `<div foo="...">`.
+
+However, `v-bind:key` (shorthand: `:key`) will not bind anything to the HTML attribute `key`! The result will _not_ be `<div key="...">`.
+
+So I suggest to just rename `key` to `v-key`, to show that this is something _internal_ to Vue.
+
+Besides its inconsistency, the current implementation breaks the ability to actually introduce an attribute named `key`.
+
+# Detailed design
+
+# Drawbacks
+
+# Alternatives
+
+Leave it as it is :-)
+
+# Adoption strategy
+
+For a transition period, support both syntaxes, and log a deprecation warning to the console.
+
+# Unresolved questions


yyx990803:
Please follow the proper RFC format, especially consider:

- The impact of this change (this is a major breaking change).
- The gain of this change
  - AFAIK there are no valid use of `key` as an attribute on an HTML element. If anything, semantically correct usage would be binding to `data-key` via dataset API.

ThomasLandauer:
I tried to follow your RFC format.

The gain of this change is mainly **consistency** (see "Motivation"). And there is no section in the template for "impact".

stefnotch:
This RFC could be fleshed out a bit more. For instance, an obvious drawback is breaking backwards compatibility.  This breaks a large number of existing Vue.js components, and developers have to learn a new approach.


One part of the adoption strategy could be an automatic conversion from the old format to the new one? And maybe a warning if :key is used anywhere?

LucasLeandro1204:
for me, it never make much sense binding key. Myself had a number of situations where I could use `key` as a prop, but wasn't able to do so.
+1 for v-key

michaeldrotar:
Some real world examples of where people would want to use `:key` would be useful.

yyx990803:
- `key` and `ref` have both been explicitly reserved props that work with `v-bind` since v2. If you want to change the usage of `key`, you'd have to change the usage of `ref` too, so the potential breakage impact would be even larger. This should be discussed in the drawback section.

- As @michaeldrotar mentioned, a real world use case of actually binding `key` as an attribute would not only be useful, but actually required. "Consistency" alone, IMO, does *not* justify the breaking change involved.

Aferz:
This would break the directives named as `key` in userland.

pikax:
I think having `v-key` and `v-ref` is a good idea, since `key` and `ref` are reserved props, in my opinion vue reserved props should always start with `v-`, even tho that might be a pain to get used to in the beginning.

smolinari:
@pikax - things starting with v- in Vue are **directives**. v-for is a directive. v-if is a directive. v-bind is a directive. And in that line of thinking and where `ref` and `key` sit and what they do, they make much more sense as props, even if they are bound with v-bind. Or put another way, they make much less sense as directives. 

From the docs:

>Directives are special attributes with the v- prefix. Directive attribute values are expected to be a single JavaScript expression (with the exception of v-for, which will be discussed later). A directive’s job is to reactively apply side effects to the DOM when the value of its expression changes.

So, reading that, it's clear `v-key` and `v-ref` wouldn't directly fit the definition.

Scott

pikax:
@smolinari `:key`, `:ref` and `:is` are already an exception for attributes and props...

What I'm suggesting is having a way to allow the user to use the current v2 reserved keywords as an html attribute.

[Special Attributes](https://vuejs.org/v2/api/#Special-Attributes) `key`, `ref` and `is`.

Aferz:
If breaking changes are not so dramatic (It looks like it can be automatically migrated with a script, with the exception of custom directives), I think I'd prefer that `key`, `ref` and `is` were a *special* directive better than a *special* attribute, to avoid limitation about which attributes a DOM node can has and, for example, a new html attribute that could collide with them.

smolinari:
@Aferz - Welp. I personally disagree. Making them directives breaks the concept of a directive.  And, they simply make more sense as props. Plus, naming something another name, i.e. `key-data` for `key` and `reference` for `ref`, etc. is notoriously easy to do. 

@pikax - They are reserved props for Vue to work. Just like you have reserved keywords in JavaScript you can't use as variables. 

Scott

yyx990803:
I'll take this opportunity to clarify our general stance on the category of "minor syntax improvement" RFCs - it's not really about what is conceptually more "consistent" or "correct", but rather what practical benefits you'd get by making a change like this, and what cost comes with it.

As far as we can tell, these reserved props are working just fine (otherwise we'd have much more interest in this proposal), and don't really cause usability issues. The fact that they are somewhat Vue-special but not starting with `v-` may feel like a pet peeve to some, but not so much to others (e.g. I don't like making them directives for the exact reason @smolinari stated: they do not fit into the concept of directives.)

The only objective gain from making this change is then allowing users to bind `key` and `ref` as native attributes - which, at this moment, doesn't seem to serve any real purpose. If that is something absolutely needed, it can be done via a custom directive as well. At the same time, changing them to `v-key` and `v-ref` also breaks existing userland custom directives of the same names - which cannot be easily upgraded with codemods.

Finally, the downside: there are tons and tons of documentation, education materials and existing codebase that would need to be updated just for this change, and tons of developers used to `:key` and `:ref` would have to unlearn the habit. This is an ecosystem stability concern. Vue 2's API has been stable for 3 years and has huge amount of existing usage. Frankly, it's simply no longer at a stage where we can make minor adjustments with marginal/subjective improvements (this is especially true for the template syntax), because every such change comes at a considerable cost. A breaking change RFC worth considering should at least fit one of the following:

- Affects only advanced features that only impact a very limited set of users;
- Can be fully automatically upgraded via codemods, with no edge cases;
- Provides more objective improvements (e.g. unlocking a use case that had no workarounds before, or making a common use case drastically simpler) than just syntax consistency.

cwxyz007:
As @yyx990803 said, It not a good option to make a break change. And I think the reason for this RFC is how to resolve when prop name is the same with reserved props (`key` and `ref`). So, how about use a modifier like `:key.prop`, and it means `key` is a prop. Since we have `.sync` modifier first.

Example:

```html
<comp
  v-for="(o, idx) in 10"
  :key="idx" // this is a key for v-for 
  :key.prop="o" // this is a prop for component
  :other-prop="o" // this is a custom prop
/>
```

michaeldrotar:
This is getting way over-engineered... every language has reserved words, there's no need for adding a modifier for this.

Nobody has offered a real-world use case where `key` itself is the optimal name so I don't think this is an actual problem.

cwxyz007:
@michaeldrotar Yes, you are right, Before solve the problem, we need to confirm it is an actual problem. Sorry I miss that.

yyx990803:

PR #92, opened 2019-11-06
Attr fallthrough behavior

yyx990803:
This is a follow up of #26, but with much less breakage.

- Make the attrs fallthrough behavior more consistent;
- Make it easier to pass all extraneous attrs to child elements / components.

[Rendered](https://github.com/vuejs/rfcs/blob/attr-fallthrough/active-rfcs/0000-attr-fallthrough.md)

### DIFF
--- a/active-rfcs/0000-attr-fallthrough.md
+++ b/active-rfcs/0000-attr-fallthrough.md
@@ -55,7 +55,7 @@ In 3.0, we are planning to make attribute fallthrough an explicit decision of co
 
 - `inheritAttrs` option will be removed.
 
-- `.native` modifier will be removed.
+- `.native` modifier for `v-on` will be removed.
 
 - Non-prop attributes no longer automatically fallthrough to the root element of the child component (including `class` and `style`). This is the same for both stateful and functional components.
 
@@ -74,6 +74,8 @@ In 3.0, we are planning to make attribute fallthrough an explicit decision of co
   ```
 
   Both `bar="2"` AND `class="bar"` on `<child>` will be ignored.
+ 
+- `this.$listeners` will be removed.
 
 - `this.$attrs` now contains **everything** passed to the component except those that are declared as props or custom events. **This includes `class`, `style`, `v-on` listeners (as `onXXX` properties)**. The object will be flat (no nesting) - this is possible thanks to the new flat VNode data structure (discussed in [Render Function API Change](https://github.com/vuejs/rfcs/pull/28)).
 
--- a/active-rfcs/0000-attr-fallthrough.md
+++ b/active-rfcs/0000-attr-fallthrough.md
@@ -1,4 +1,4 @@
-- Start Date: 2019-04-08
+- Start Date: 2019-11-05
 - Target Major Version: 3.x
 - Reference Issues: N/A
 - Implementation PR: N/A
--- a/active-rfcs/0000-attr-fallthrough.md
+++ b/active-rfcs/0000-attr-fallthrough.md
@@ -181,7 +181,7 @@ This will be one of the changes that will have a bigger impact on existing code 
 
   - Declare props and events that are consumed by the HOC itself (thus removing them from `$attrs`)
 
-  - Refactor the component and explicitly add `v-bind="$attrs"` to the target inner component or element. For render functions, apply `$attrs` with the `cloneVNode` helper.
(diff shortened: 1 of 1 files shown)


tmorehouse:
For class and style bindings, and `inheritAttrs` is false (or implied false), how would one make the style/class binding apply to the root element, but not the inner element?

```html
<template>
  <div :class="$attrs.class" :style="$attrs.style">
    <!-- How does one easily prevent style and class from also being applied here? -->
    <div v-bind:"$attrs">
       Foobar
    </div>
  </div>
</template>
```

Perhaps there needs to be similar flags for `inheritStyle` and `inheritClass`?

Meekohi:
> For class and style bindings, and `inheritAttrs` is false (or implied false), how would one make the style/class binding apply to the root element, but not the inner element?
> 

Specifically, you may often wish to apply all event listeners to a child element... maybe a simple helper that plucks off the listeners from `$attrs` would cover most cases? `v-bind:"listenerProps($attrs)"`?

yyx990803:
@tmorehouse you can create a computed property based on `$attrs`, in which you can pluck off anything you don't want to pass down to the inner element.

tmorehouse:
@yyx990803 so something like this could be done to omit style/class:

```js
export default {
  props: {
    foo: { type: Boolean, default: false }
  },
  computed: {
     computedAttrs() {
       return { ...this.$attrs, style: null, class: null }
     }
  }
}
```

Although is `this.$attrs` reactive (to trigger the update of the computed prop)?

Similarly, if one wanted the v-on listeners (which are now top level `onXXX` properties in `$attrs`), one would need to use a bit more convoluted code to eliminate the listeners (or extract the listeners) so they can be targetted to another element:

```js
export default: {
  computed: {
    computedListeners() {
      return Object.keys(this.$attrs).reduce((obj, key) => {
        if (/^on[A-Z]/.test(key)) {
          obj[key] = this.$attrs[key]
        }
        return obj
      }, {})
    },
    computedAttrs() {
      return Object.keys(this.$attrs).reduce((obj, key) => {
        if (!/^on[A-Z]/.test(key)) {
          obj[key] = this.$attrs[key]
        }
        return obj
      }, {})
    }
  }
}
```

Perhaps exposing a few utility methods like `omit(obj, keysArr)` and `pluck(obj, keysArr)` might be useful.

yyx990803:
@tmorehouse it does remind me that `$attrs` is not reactive and a simple method call is probably more consistent.

As for helpers like `omit` and `pluck` - these are generic JS utilities and there are already plenty of existing libraries out there, and Vue will provide helpers that are Vue-specific.

michaeldrotar:
this probably isn't the best approach, and I haven't Vue'd in almost a year so I'm sure I have some syntax wrong.. but could you do something like this?

```vue
<!-- search-component.vue -->
<template>
  <form v-bind="$attrs.form" @click="submit">
    <input type="text" v-bind="$attrs.input">
    <button type="submit" v-bind="$attrs.button">
       Search
    </button>
  </form>
</template>
```

Usage:
```vue
<search-component
  input="{ name: 'foo', value: searchText }"
  form="{ onClick: fancySubmit }"
/>
```

leopiccionia:
@michaeldrotar I think it should be `<search-component :input={ ... } :form="{ ... }/>`, to bind an object instead of a string. But I think that registering a prop, in this case, would be more idiomatic than using `$attrs`.

cjpearson:
Does this also affect v-slot props? Would listeners be exposed as onXXX props there?

yyx990803:
@cjpearson yes.

yyx990803:

tmorehouse:
I still like the idea of $listeners which pre-separates events from attributes to make it easier for plucking out event handlers instead of creating wildcard RegExp's to find anything that matches `/^on[A-Z]/`

Perhaps instead of `$attrs` having everything, maybe `$all` would have everything, and then `$attrs` is any non event, and `$listeners` has only events.

This would follow more closely the template convention of `@xxx` for events, and `:yyy` (or yyy) for attributes/props. unless instead event handlers will be placed on elements/components as `:on-foo-event="handler"`

eliottrobson:
I quite like the use of `$listeners` to enable components to be re-used more easily (for different use cases). For example I might be tempted to add a `cursor: pointer;` css class if there is an event listening to `click`.

I appreciate I could also pass this class down or handle it in other ways but it sometimes makes more sense to let the component handle its own rendering without passing two sources of truth - the event handler and if there is one.

yyx990803:
This proposal has been replaced by #137.

PR #95, opened 2019-11-08
Remove keyCode support in v-on

yyx990803:
- Drop support for using numbers (keyCodes) as `v-on` modifiers
- Remove `config.keyCodes`

[`KeyboardEvent.keyCode` has been deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) and we already have support for using [`KeyboardEvent.key`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) as modifiers.

[Rendered](https://github.com/vuejs/rfcs/blob/drop-keycode/active-rfcs/0000-drop-keycode-support.md)


Meekohi:
Browser support for `.code` is still pretty bad: https://caniuse.com/#feat=keyboardevent-code

leopiccionia:
`KeyboardEvent.key` may often return values that aren't template-friendly, like `.`, `=`, `>`, etc. I can't imagine common cases where those would be handy, though. On the worst hypothesis, one can handle the `key` inside event handler.

An exception would be the space bar (` `), that even gained a "named" modifier in 2.x (`.space`). Will "named" 2.x modifiers like this be kept? Some of them (`.esc`, the four arrows) have non-standard `key`s in IE11, according to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values).

yyx990803:
@leopiccionia yes, 2.x built-in aliases will be kept.

yyx990803:

PR #101, opened 2019-11-15
Guards Stack in VueRouter

Aferz:
[Fully rendered proposal](https://github.com/Aferz/rfcs/blob/master/active-rfcs/0000-vue-router-guards-stack.md)

### In-route guard

```javascript
import needsToBeAuthenticated from '@/router/guards/needsToBeAuthenticated'
import needsToBeAuthorized from '@/router/guards/needsToBeAuthorized'

const routes = [
  {
    path: '/authenticated',
    component: () => import('@/pages/Authenticated'),
    beforeEnter: needsToBeAuthenticated,
  },
  {
    path: '/authorized',
    component: () => import('@/pages/Authorized'),
    beforeEnter: needsToBeAuthorized,
  },
  {
    path: '/authenticated-and-authorized',
    component: () => import('@/pages/AuthenticatedAndAuthorized'),
    beforeEnter: [
      needsToBeAuthenticated,
      needsToBeAuthorized
    ],
  }
]
```

### In constructor guard

```javascript
const router = new VueRouter({
  ...,

  // As array
  beforeEach: [
    (to, from, next) => { next() },
    (to, from, next) => { next() }
  ],

  // As Function
  beforeResolve: (to, from, next) => { next() },

  // As array, again
  afterEach: [
    (to, from) => { ... }
  ],

})

```

### DIFF
--- a/active-rfcs/0000-vue-router-guards-stack.md
+++ b/active-rfcs/0000-vue-router-guards-stack.md
@@ -1,0 +1,57 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- Target Major Version: (2.x / 3.x)
+- Reference Issues: (fill in existing related issues, if any)
+- Implementation PR: (leave this empty)
+
+# Summary
+
+Brief explanation of the feature.
+
+# Basic example
+
+If the proposal involves a new or changed API, include a basic code example.
+Omit this section if it's not applicable.
+
+# Motivation
+
+Why are we doing this? What use cases does it support? What is the expected
+outcome?
+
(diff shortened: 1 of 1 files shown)


posva:
Thank you for this but I don't think we need an RFC since the unresolved questions you talk about are answered by the order of guard execution. Composition api will only provide in-component level guards.

There is already an opened issue at https://github.com/vuejs/vue-router/issues/2688 and arrays will be supported in the rewrite 🙂

Aferz:
Mmh that looks pretty much like this RFC. We can close this then.

Thanks @posva

PR #105, opened 2019-11-29
<transition> API changes

yyx990803:
- Transition class name adjustments [[Rendered](https://github.com/vuejs/rfcs/blob/transition/active-rfcs/0000-transition-class-change.md)]
  - Rename the `v-enter` transition class to `v-enter-from`
  - Rename the `v-leave` transition class to `v-leave-from`

- Using `<transition>` as component's root will no longer trigger transitions when the component is toggled from the outside. [[Rendered](https://github.com/vuejs/rfcs/blob/transition/active-rfcs/0000-transition-as-root.md)]


CyberAP:
What would be the best way to refactor such code regarding we can no longer use `v-if`?

```html
<TransitionExpand v-if="user">
  <UserInfo :user="user" />
</TransitionExpand>
```

`user` is a required prop and `<UserInfo>` component is not valid until it receives `user`.

If we refactor it like this we would get an error:

```html
<TransitionExpand :visible="Boolean(user)">
  <UserInfo :user="user" />
</TransitionExpand>
```

Should we write `v-if` for slot contents as well now? Would that trigger a transition?

```html
<TransitionExpand :visible="Boolean(user)">
  <UserInfo v-if="user" :user="user" />
</TransitionExpand>
```

robertmoura:
You'd probably just do it the normal way:
```html
<TransitionExpand>
    <UserInfo :user="user" v-if="user" />
</TransitionExpand>
```

yyx990803:

PR #112, opened 2020-01-20
<Teleport> built-in component

LinusBorg:
This RFC introduces a `<teleport>` component, which allows to move its slot content to another part of the document.

[Rendered](https://github.com/vuejs/rfcs/blob/rfc-portals/active-rfcs/0025-teleport.md)


kiaking:
Great work! 🎉  Is it worth mentioning about SSR? I know that [Portal Vue is suffering](https://portal-vue.linusb.org/guide/SSR.html) with it.

active-rfcs/0000-portals.md
```diff
@@ -0,0 +1,355 @@
...
+    const { Portal } = window.Vue;
+    new Vue({
+      el: "#app",
+      components: {
+        Portal,
+        Child: { template: "<div>Placeholder</div>" }
+      }
+    });
+  </script>
+  <body></body>
```
potato4d: What does this body tag mean?

cawa-93:
The name can lead to conflict: https://web.dev/hands-on-portals/

I find this proposal useful, but I do not have enough information on how it should behave.
What to do when a portal points to an item that already contains content or another component?

```vue
<Portal target=".content">Portal content</Portal>
<div class="content">Div Content</div>

<Portal target=".component">Portal content</Portal>
<my-component class="component">Component Content</my-component>

<Portal class="portal" target=".another-portal">Portal content</Portal>
<Portal class="another-portal" target=".portal">Another Portal content</Portal>
```

ycmjason:
@cawa-93 is right. We should use another name? `v-portal`?

gustojs:
If we start calling new components `v-stuff`, we'd have to rename current ones for them to follow same naming style. Also `v-portal` suggests it's a Vuetify component.

cawa-93:
> @cawa-93 is right. We should use another name? `v-portal`?
Personally, I like @aztalbot  suggestion: `<Mount target=“”>` - this name reflects the behavior of the component well.

But I would change the name of the attribute:
```
<Mount to=“”></Mount>
```

posva:
@cawa-93 @ycmjason did you check https://github.com/vuejs/rfcs/blob/rfc-portals/active-rfcs/0000-portals.md#naming-conflict-with-native-portals ?

brandonpittman:
+1 for `<wormhole>`

michaelpumo:
If <portal> will create a conflict against the HTML spec, perhaps we could simply utilise the existing <template> tag with a directive? 

Example:

<template v-portal=“example”>
  Hello Vue 3!
</template>

udany:
I'm currently using a custom written mixin to make components that can be portaled around, I'll include it in the end in case it's of use. 

Two things I'd like to point out are:

1. I think it would be better if the target could also support an actual DOM element instead of only a query selector. 

2. I'd like to be able to choose between an append mode besides replace mode (i.e. Instead of replacing the target's content, merely append to it) 

```
export default {
	props: {
		parent: {
			type: HTMLElement,
			required: true
		}
	},
	watch: {
		parent(newValue, oldValue) {
			if (oldValue) this.detachFromElement(oldValue);
			if (newValue) this.attachToElement(newValue);
		}
	},
	methods: {
		attachToElement(parent) {
			if (!parent) return;
			parent.appendChild(this.$el);
			this.$emit('attachedToElement', parent);
		},
		detachFromElement(parent) {
			if (!parent) return;
			this.$emit('detachedFromElement', parent);
		}
	},
	mounted() {
		this.attachToElement(this.parent);
	},
	beforeDestroy() {
		this.detachFromElement(this.parent);
	}
};
```

ycmjason:
> @cawa-93 @ycmjason did you check https://github.com/vuejs/rfcs/blob/rfc-portals/active-rfcs/0000-portals.md#naming-conflict-with-native-portals ?

yes. so I was just suggesting names? 😄

CyberAP:
How does Portal work with refs? Should you put a ref on a Portal or on its contents only?

JosephSilber:
If we can't use the same target for multiple portals, how can we use it, for example, for multiple stacked modals? Would we need to manually created additional portal targets in setup?

This is slowly starting to feel less data driven, and more like raw DOM manipulation.

LinusBorg:
> If we can't use the same target for multiple portals, how can we use it, for example, for multiple stacked modals? Would we need to manually created additional portal targets in setup?

@JosephSilber That's what the RFC is for: To answer these questions.

LinusBorg:
> How does Portal work with refs? Should you put a ref on a Portal or on its contents only?

I it should not be püossible to put a ref on the `<portal>` - it doesn'T havce (or at least doesn't require) a resal component instance as it's just symbol telling the virtualdom patch algo to move its contents somewhere else.

If we find, over the course of this discussion, that portals need to have an instance to provide additional functionality not yet in the RFC, then this might change. But putting it on the content should always work.

Do you have a use case for putting a ref on the `<portal>`?

LinusBorg:
@udany 

> How does Portal work with refs? Should you put a ref on a Portal or on its contents only?

I agree, will add this.

> I'd like to be able to choose an append mode besides replace mode (i.e. Instead of replacing the target's content, merely append to it)

Should be considered, would also solve @JosephSilber's issue, presumably.

active-rfcs/0000-portals.md
LinusBorg: It means I have to fix a typo and remove the superflous opening tag.

That there's a body tag at all is because the example is means to be in HTML, not in an SFC.

LinusBorg:
@kiaking SSR and Portals are tricky as usually the app is rendered in one go. 

I didn't add it to the RFC for now as we are still exploring SSR architecture for Vue 3 and want to see what we come up with in general before re-visiting this in the RFC

kiaking:
@LinusBorg Make sense. Thanks for the confirmation! 🤝

LinusBorg:
@aztalbot 

> Selector
> Is the behavior the same as querySelector such that Portal will mount its children only to the first node it finds (and in the case of a less specific selector ignore all other nodes)?

Yes. I see you suggest to limit it to ids. I'm unsure about that, it feels unnecessessarily limiting.

> Vue Control

I'll need to think about this  bit more, will get back to you about it.

> Multiple Portals
> 
> What happens to any existing children under the target? If it’s just a matter of managing order, might it suffice to have a prop called insert-before which accepts a Dom node reference before which to mount the children (this can be queried in setup)? That way a small amount of custom logic or a library can manage placement of multiple portals by determining where to insert based on what already exists under that target (eg: examine data-order attribute and sort). By default just append to end.

Yeah I see that many people like to have multiple portals moving stuff to the same place, but I'm torn if this should be part of core or we can design it in a way that a small lib can make this happen with little code.

sqal:
@JosephSilber 

> If we can't use the same target for multiple portals, how can we use it, for example, for multiple stacked modals?

Who said you can't ? You can use Portal to mount as many elements as you wan in the same target node. You can create a functional component that renders Portal and mounts target element in the DOM just once, then you can reuse this component anywhere you want.

// DialogPortal.js
```js
import { h, Portal } from 'vue';

let target = null

export function DialogPortal(_, { slots }) {
  if (!target) {
    target = document.createElement('div')
    target.id = 'dialog-root';
    document.body.appendChild(target)
  }

  return h(Portal, { target }, slots.default())
}
```

// Dialog.vue
```vue
<template>
  <DialogPortal>
    <div class="dialog">
      dialog content
    </div>
  </DialogPortal>
</template>
```

emironov-via:
Maybe make portals look like scoped slot?

Source component:
```
<template>
  <div>
    ...
    <template v-portal:example="{ foo, bar }">
      <div>{{ foo + bar }}</div>
    </template>
    ...
  </div>
</template>
```
Target component:
```
<template>
  <div>
    ...
    <portal
      name="example"
      foo="Hello"
      :bar="'world'"
    >
      default content, when source not found 
    </portal>
    ...
  </div>
</template>
```

**Pros:**
- familiar syntax and behavior;
- access to target data in the source template;
- default content;
- can be used for deep transfer slots, without `this.$slots` transition;
- source template it's like render function, and can be used for many targets;
- without DOM selector, because it feels like a leaky abstraction for Vue.

**Cons:**
- name collision.

P.S. It's may be a repetition of the situation with slots. At first there were slots, then scoped slots were added, after they were merged (more precisely, slots were removed).

codebender828:
🙏🏽❤️ Great work with this RFC, @LinusBorg  , contributors and core team with Vue 3.

I have a few thoughts on handling portals with no visible children and I'd be happy to hear some thoughts about this. It's not a very urgent requirement/feature for the `Portal` but it would certainly make my live easier. I use the `PortalVue` in a lot of my projects. However I often find myself having to create my own `Portal` component that wraps around the `MountingPortal` component from PortalVue to implement the below mentioned desired behaviour.

I'll also go ahead and say that I might be doing this wrong, and there may be a better way to implement this than I am doing. But this currently is what would work for me. So feel free to suggest your thoughts about this.

### Handling portal targets with no children
One of the things I've noticed as a library author is that I might have multiple portals for different kinds of components. I'd have a single portal for toast components, a separate one for popovers, and for tooltips as well, etc. My preference would be to have them exists in different portal targets. So, naturally I'd build my components using this pattern.

In consumer applications, however, if their production app has many implementations of popovers and tooltips, etc, they would end up having multiple portal targets in the DOM since I'm creating the portal targets and mounting them in the DOM only when the Tooltip/Popover/Menu components are consumed by the user. So we would wind up with markup somewhat similar to this:
```html
<!DOCTYPE html>
<html lang="en" dir="ltr">
  <head>
    <meta charset="utf-8">
    <title>Consumer App</title>
  </head>
  <body>
    <!-- Vue app is mounted here -->
    <div id="app"></div>

    <!-- Portal targets are mounted here -->
    <div id="popover-portal"></div>
    <div id="toast-portal"></div>
    <div id="menu-portal"></div>
    <div id="tooltip-portal"></div>
  </body>
</html>
```
### On demand Portal target creation and removal. 
This results in having multiple portal targets in the markup that (may not be used). This led me down a path of desiring to create the my own portal targets only when it's children are present/visible either by `v-if` or `v-show` directives and then remove the targets from the DOM when their children are no longer displayed.

This may require to some specific internal behaviour from the `Portal` component. `Portal` would need to:
- Create the portal target provided by the consumer if it does not exist in the DOM.
- Remove the `Portal` target node from the DOM if it's children are not displayed. It would need to detect this the same way that Vue's `Transition` and `TransitionGroup` component detects the presence of it's children components/nodes. If they're present, create and mount the portal target in the DOM provided in the `target` prop of the Portal component.
- We could perhaps add an `onDemand` Boolean prop to implement the above mentioned behaviour. It would default to `false` so as to maintain the current behaviour as mentioned in [rendered rfc](https://github.com/vuejs/rfcs/blob/rfc-portals/active-rfcs/0000-portals.md)

Implementing this could them result in markup shown below:
###  📭 Targets with invisible children
In Vue template
```vue
<template>
  <main>
    <Portal on-demand target="#popover-portal">
      <p v-if="false">Portal target will not be mounted because I'm set to v-if=false</p>
    </Portal>
  </main>
</template>

<script>
import { Portal } from "vue"
export default {
  components: {
    Portal
  }
}
</script>
```
Rendered markup
```html
<html lang="en" dir="ltr">
  <body>
    <!-- Vue app is mounted here -->
    <div id="app">...</div>

    <!-- #popover-target not created and mounted -->
  </body>
</html>
```

###  ✅ Targets with visible children
In Vue template
```vue
<template>
  <main>
    <!-- 
      on-demand prop set to true to activate
      internal portal target creation and and removal
    -->
    <Portal on-demand target="#popover-portal">
      <p v-if="true">Mounted because I'm set to v-if true</p>
    </Portal>
  </main>
</template>

<script>
import { Portal } from "vue"
export default {
  components: {
    Portal
  }
}
</script>
```
Rendered markup
```html
<html lang="en" dir="ltr">
  <body>
    <!-- Vue app is mounted here -->
    <div id="app">...</div>

    <!-- Portal targets are mounted here -->
    <div id="popover-portal">
      <p>Mounted because I'm set to v-if true</p>
    </div>
  </body>
</html>
```
Having this behaviour would keep markup clean and also be useful for authors to not have to create their own wrapper Portals  to create this kind of behaviour for their consumers :)

What do you think about this? 🙏🏽❤️

cawa-93:
@codebender828 Sounds pretty helpful. However, I see the need to specify WHERE it is necessary to create a target for the portal. And I don't see a way to conveniently set the location of this target. Maybe someone more experienced will point it out.

It might be better for these tasks to implement a new **universal** component, or directive, etc. that would remove a node if it had no child nodes?

In Vue template
```vue
<template>
  <main>
    <div remove-if-empty>
      <p v-if="false">Portal target will not be mounted because I'm set to v-if=false</p>
    </div>
  </main>
</template>
```

Rendered markup
```html
<html lang="en" dir="ltr">
  <body>
    <!-- Vue app is mounted here -->
    <div id="app">...</div>

    <!-- div not created and mounted -->
  </body>
</html>
```

In Vue template
```vue
<template>
  <main>
    <div remove-if-empty>
      <p v-if="true">Mounted because I'm set to v-if true</p>
    </div>
  </main>
</template>
```

Rendered markup
```html
<html lang="en" dir="ltr">
  <body>
    <!-- Vue app is mounted here -->
    <div id="app">...</div>

    <div>
      <p>Mounted because I'm set to v-if true</p>
    </div>
  </body>
</html>
```

Janne252:
What if the syntax for controlling a component's mounting root was a global directive; `v-mount`? This feature behaves quite similarly to [$mount(...)](https://vuejs.org/v2/api/#vm-mount). Wouldn't it make sense to name them consistently?

```html
<template>
  <main v-mount="$root">
    <!-- content -->
  </main>
</template>
```
*Assuming the mounting target is `string` | `VNode` | `Element`

udany:
@codebender828

While I appreciate the concerns you have for keeping the dom uncluttered, I feel like this could be premature optimization, I can hardly see how a few (even tens) empty divs could harm an application's performance to make the complexity this behavior would add to the implementation worth it.

@cawa-93 suggestion is great for it makes it an independent feature rather than making portals more complex, even feels like a directive that could be implemented in userland

codebender828:
@cawa-93 Interesting. I also like the directive approach. 

@udany Thanks for sharing. I agree that this might actually be a premature optimization. 🤔 A few divs would not be a performance constraint. 

It's only based on an observation I've seen with React's `createPortal` API. Since I was familiar with this it quite naturally made sense to me to remove the target if it has no children. But there's also no harm in having a few nodes in the DOM. In fact I've seen some cases where hundreds of portal targets were used and performance was still fine. Thanks ❤️

LinusBorg:
> Perhaps to make mounting within a Vue app more predictable [...]

I thought about this a bit and I think that the functionality proposed within this RFC should primarily focus on the 80% use case, which is moving stuff to a place *outside* of the app, usually as a direct child of the body element, in order to prevent styling/positioning/z-index problems.

I understand that a lot of people like what you can do today with portal-vue, namely move content around within your Vue app freely, but due to the dynamic nature of Vue apps and the component lifecycle, this requires some kind of orchestration to that both source and target can realise when the other is removed, or e.g. the target is added after the source portal has rendered. Taking things like order, scoped slots and other features into account, some of which were mentioned in this discussion here, this leads to a significant footprint in terms of code required.

That's basically why portal-vue weighs a closer to 3kb gzipped, why my less-known lib, `vue-simple-portal`, which imlements the aforementioned "80%" use case only, is as light as 500b.

Therefore

LinusBorg:
Concering the proposals of as directive ... I don't really feel it. a component feels more natural to me to encapsulate this behaviour.

donnysim:
Current portal-vue also has problems with transitions where the element is instantly removed by the portal without waiting for leave animation to finish as there's no way to check if the element is wrapped in transition as the transition component is not referenced on the DOM element, it just hangs in the air. While there are other ways to achieve portal functionality that also works with transitions, I do prefer it being supported out of the box without hacking around the framework, and just rendering it to another place/element is more than enough for majority of cases.

Could you clarify about the slots?

Would this work:
```jsx
const popupContent = <popper-transition key="popper-transition">{this.$scopedSlots.popup()}</popper-transition>;

return (
  <this.tag class={this.classList} onClick={this.onClick}>
    {this.toBody ? <to-body key="popper-to--body">{popupContent}</to-body> : popupContent}
    {this.$scopedSlots.default()}
  </this.tag>
);
```

as in `<to-body` (component that moves the element to then end of body, kind of like portal) would be replaced with `<portal`?

LinusBorg:
@donnysim 

> Current portal-vue also has problems with transitions 

Indeed. the root cause is that transitions are handled at the virtual dom patch level, whereas portal-vue moves virtual DOM around before it's being processed.

As mentioned, we would handle the portal funcionality at the virtual DOM level now, so transitions should work properly with this.

> Could you clarify about the slots?
> Would this work:

Yes, that would work.

bladerunner41:
The [Vis.js timeline library](https://visjs.github.io/vis-timeline/examples/timeline/) has an [example](https://visjs.github.io/vis-timeline/examples/timeline/other/usingReact16.html) (right click - view source) of using React v16 Portals to render a React component in the library's standard renderItem() template function to customise an item's display/text instead of a developer constructing static HTML manually.

Will the vue.js portal implementation also support the equivalent syntax needed:
`ReactDOM.createPortal( ReactDOM.render( <ItemTemplate item={item} />, element ), element, () => { window.timeline.redraw()} );`

nash403:
I am really excited with this proposal ! thank you @LinusBorg  😃 

Is it possible to consider a `disabled` prop like in `portal-vue` that would have the effect of rendering the content in place when truthy instead of moving it to the target ?

Or maybe is it better to emulate this behavior with the following boilerplate ? : 

```html
<Content v-if="someData" />
<Portal v-else>
	<Content />
</Portal>
```

But I think this will destroy the `Content` instance and recreate it whenever someData is toggled whereas with a prop the instance could be kept alive ? 🤔

donnysim:
I wonder how could we tailor Portal for dropdowns? Taken a menu as an example, on desktop the dropdown element is placed on body, but on mobile it is kept in original placement and toggled as a sliding menu - would be nice if we didn't have to resolve to v-if to toggle between portal and normal content (but maybe could be wrapped into a separate component), another need is to catch the re-positioning (from body to initial place) of the element for actions like detaching or reattaching Popper.js. Would love if it could be (or maybe already can be) achieved with minimal effort.

CyberAP:
I'd like to be able to choose Portal strategy in working with DOM node. Let's imagine we have a sidebar that's out of the scope of the main application. We would like to be able to push content to that sidebar, but with Portal every time you target this sidebar the whole element is replaced. Having `<Portal target="[data-sidebar] " strategy="append">` would solve this issue. I would also like for Portal to retain the original root node it's using as a target and always replace the inner content of that node. That way you'd be able to have dynamic Portal strategy.

LinusBorg:
@donnysim 

> Taken a menu as an example, on desktop the dropdown element is placed on body, but on mobile it is kept in original placement and toggled as a sliding menu - would be nice if we didn't have to resolve to v-if to toggle between portal and normal content (but maybe could be wrapped into a separate component), *[...]*

A way t essentially "disable" a portal conditionally was already proposed, and I think it would be quite useful:

```
<portal :disabled="isMobile">
  <!-- -->
</portal>
```

> another need is to catch the re-positioning (from body to initial place and vice verse) of the element for actions like detaching or reattaching Popper.js. 

Not sure I totally get that usecase - I get the direction but, lack a detailed image in my head. can you add some details for me?

LinusBorg:
@CyberAP 

> *[...]* Portal strategy used when working with DOM node. *[...]*
> `<Portal target="[data-sidebar] " strategy="append">`

I like this, but it has to be kept simple. 

What kinds of strategies would we see here? 

* `append`
* `prepend`
* `replace` (would mean we have to unmount the other component, which might be challenging with the current implementation)

I wonder though, if it would make more sense to let these cases be solved in userland as long as we can make sure in Vue's core that they are possible to be implemented.

> I would also like for Portal to retain the original root node it's using as a target and always replace the inner content of that node. 

That's already part of the RFC.

donnysim:
> Not sure I totally get that usecase - I get the direction but, lack a detailed image in my head. can you add some details for me?

Was kind of thinking about being able to trigger callbacks based on action from the portal - done moving the element etc. to be able to execute further actions without having to resort to `$nextTick` etc. like:

```html
<portal :disabled="screenWidth < 768">
  <div @after-move-to-portal="attachPopper" @before-move-to-host="detachPopper"/>
</portal>
```
(I know the event names make no sense)
```js
methods: {
  attachPopper(el) {
    el._popper = createPopper(el);
  },

  detachPopper(el) {
    el._popper.destroy();
  },
},
```

or something along those lines, but I realize that this kind of makes no sense? like if it's a component and contains multiple root elements etc. it would fail pretty fast. Most likely this should be encapsulated as a component and just toggle the "use-popper" on and off based on the same condition as the parent. Main problem with something like popper is that it saves the parent element on bind, but if you move it using portal, it breaks real fast.

At the moment for something similar I use directives `v-inserted`, `v-removed` (basic `inserted` and `unbind` hooks) to execute callbacks on element insert/remove to trigger additional actions on that element to simplify dealing with v-if's and similar conditions, but probably not going to work with portal.

yyx990803:
We have implemented multi-portal-same-target append support and the `disabled` prop, all with SSR support in `vue-next` master branch - RFC update coming soon.

Regarding the name conflict, we are currently leaning towards `<Teleport>` with a `to` prop:

``` html
<Teleport to="#modal-layer" :disabled="isMobile">
  <div class="modal">
    hello
  </div>
</Teleport>
```

CyberAP:
I had to read about `vue-portal`'s `disabled` prop first to understand what it actually does. For me disabled means the Teleport contents are not displaying or not functioning and is a bit confusing.
What about using `null` as a target to disable teleporting?

 ```
<Teleport :to="null" />
```

**Comment by LinusBorg** [comment]
I just pushed the update!

**Comment by yyx990803** [comment]
@CyberAP neither of your suggestions are as explicit as a `disabled` prop. A falsy target value also should result in a warning, using `null` as a special value is more implicit-ness. Your 2nd use would teardown the inner content and re-mount it, whereas with `disabled` prop the portal preserves the content and simply moves it.

File: active-rfcs/0000-portals.md [hunk 2]
```diff
@@ -0,0 +1,321 @@
...
+};
+```
+
+When using a render function or JSX, component has to imported first, like any other component:
+
+```js
+import { Teleport, h } from "vue";
+export default {
+  render() {
+    return h("div", [h(Teleport, { target: "#endofbody" }, ["Some content"])]);
```
yyx990803:
```suggestion
    return h("div", [h(Teleport, { to: "#endofbody" }, ["Some content"])]);
```

active-rfcs/0000-portals.md
```diff
@@ -0,0 +1,321 @@
...
+  },
+  // or with JSX:
+  render() {
+    <div>
+      <Telport target="#endofbody">Some content</Teleport>
```
yyx990803:
```suggestion
      <Telport to="#endofbody">Some content</Teleport>
```

active-rfcs/0000-portals.md
```diff
@@ -0,0 +1,321 @@
...
+## Miscellaneous
+
+### Naming conflict with native portals
+
+the component introduced by this RFC was named `<portal>` in an earlier version of this RFC. But there's proposal for native portals:
+
+- Spec: https://wicg.github.io/portals/
+- Introduction: https://web.dev/hands-on-portals/
+
+Sinc we don't want to have a naming conflict with a future HTML element that may be called `<teleport>`, especially since it's functionality is about something completely different form what portals in libs like Vue or React mean right now, we chose to rename the component to `<teleport>`
```
yyx990803:
```suggestion
Sinc we don't want to have a naming conflict with a future HTML element that may be called `<portal>`, especially since it's functionality is about something completely different form what portals in libs like Vue or React mean right now, we chose to rename the component to `<teleport>`
```

yyx990803:

leopiccionia:
Something that occurred me: why not name it `<Mount to="...">` (or maybe `<Mount on="...">`, I'm not a native speaker...) instead of `<Teleport to="...">`?

It'd be one less buzzword to teach and learn, except if it's somehow misleading. Is it?

(Sorry to address this so late...)

yyx990803:
@leopiccionia thanks for the suggestion, but the naming has been extensively discussed and the current choice of `<teleport>` has consensus on the core team.

kikyous:
Teleport is great, any vue 2 implement ?

sionzeecz:
Are there any plans on events for teleport?

At my case I need to teleport an element but adjust the coords to old location. 
But I cannot gather any information about the element when was teleported. 

So events like:
beforeEnterTeleport
enterTeleport

are really needed.

PR #117, opened 2020-01-21
Global API updates: more details on mounting behavior + replacement for Vue.prototype

yyx990803:
- More details on the mounting behavior change
- Offer a replacement for attaching properties to `Vue.prototype` in 2.x

### DIFF
--- a/active-rfcs/0009-global-api-change.md
+++ b/active-rfcs/0009-global-api-change.md
@@ -25,6 +25,8 @@ Vue.mixin(/* ... */)
 Vue.component(/* ... */)
 Vue.directive(/* ... */)
 
+Vue.prototype.customProperty = () => {}
+
 new Vue({
   render: h => h(App)
 }).$mount('#app')
@@ -43,6 +45,8 @@ app.use(/* ... */)
 app.mixin(/* ... */)
 app.component(/* ... */)
 app.directive(/* ... */)
+
+app.config.globalProperties.customProperty = () => {}
 
 app.mount(App, '#app')
 ```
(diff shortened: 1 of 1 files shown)


active-rfcs/0009-global-api-change.md
```diff
@@ -160,6 +176,21 @@ app.config.isCustomElement = tag => tag.startsWith('ion-')
...
+
+In Vue 3, since the global `Vue` is no longer a constructor, this is no longer supported. Instead, shared instance properties should be attached to an app instance's `config.globalProperties` instead:
+
+``` js
+// Before
+Vue.prototype.$http = () => {}
+
+// After
+const app = createApp()
+app.config.globalProperties.$http = () => {}
```
posva: how would we access this inside the `setup` function?

active-rfcs/0009-global-api-change.md
```diff
@@ -44,6 +46,8 @@ app.mixin(/* ... */)
 app.component(/* ... */)
 app.directive(/* ... */)
 
+app.config.globalProperties.customProperty = () => {}
```
posva: Is there an interface that can be extended to type new properties?

active-rfcs/0009-global-api-change.md
yyx990803: This is essentially a legacy escape hatch and not for `setup` function. You should just import things directly or use `provide/inject` in setup functions.

active-rfcs/0009-global-api-change.md
yyx990803: I'd say just don't use this if using TS.

posva: Wouldn't this be better for people migrating TS projects using libraries that are not adapted yet (or that may never completely adapt) to the inject/provide?

yyx990803: There's no way to make this work in TS though. If users insist on using this with TS they'd have to shim the types themselves.

posva: > If users insist on using this with TS they'd have to shim the types themselves

Oh yeah, I just thinking of exposing an interface that can be extended by users to add their own typings

active-rfcs/0009-global-api-change.md
CyberAP: This won't work for legacy apps that use multiple instances and apply `Vue.use` globally since you'd have to run `app.use` in every one of them. If that's required only for legacy plugins shouldn't that be included only in the compatibility build with global registration on the prototype as usual? That way you won't have to rewrite that much and stay with global `Vue.use`. Why would you need that hack for anything but legacy plugins? (Assuming that the new strategy for plugins is to utilize `provide`\\`inject`).

yyx990803: The "legacy" term I used was incorrect - it should be for "options/this-based usage". Even with `setup` function this can still be used to expose globally available properties to templates.

With the global API change there is no once-for-all global configs anymore. 2.x multi-root-instance apps will have to be updated to configure each root instance individually. If all root instances share the same config, then it can be extracted into a shared function.

PR #118, opened 2020-01-21
Events API Change

yyx990803:
## Summary

Remove `$on`, `$off` and `$once` instance methods. Vue instances no longer implement the event emitter interface.

[Rendered](https://github.com/vuejs/rfcs/blob/events-api-change/active-rfcs/0000-events-api-change.md)


yyx990803:

bodograumann:
@yyx990803 Please update the link to the rendered rfc.

blake-newman:
@bodograumann I have updated.

muthu32:
Thanks for notifying about [mitt](https://github.com/developit/mitt).

kouts:
Sorry if this is an obvious question, in Vue.js 2 you could extend a component and listen to events that the parent (super) component emitted.
Example:
```
// Popup component
{
  ...
  name: 'Popup', 
  methods: {
    afterOpen () {
      this.$emit('after-open');
    }
  }
  ...
}

// CustomPopup extends Popup
const CustomPopup = {
  extends: Popup,
  ...
  created () {
    this.$on('after-open', function () {
       // do something here...
    });
  }
  ...
};
```
Is there any way to achieve the same now that $on will be removed?
Do component instances now have to depend on an external library that emits these events?

mika76:
@kouts I need to know this too 😊

yyx990803:
@kouts @mika86 yes, use a 3rd party lib. `mitt` is tiny.

randriese:
I'm a bit confused about this RFC... Not by the proposal to remove the instance methods and I have no objections on using Mitt, however, how would this play with [https://github.com/vuejs/rfcs/blob/master/active-rfcs/0030-emits-option.md](url) as that states the usage of $emit?

sinkersan:
Hello.
In Vue 2 we have option to subscribe on lifecycle hook like on events. This gives us ability to write something like this
```
export default {
  mounted() {
    const picker = new Pickaday({
      // ...
    });

    this.$once("hook:beforeDestroy", () => {
      picker.destroy();
    })
  }
};
```

Since in Vue 3 $once and $on are deprecated. Do we have an option to migrate this code and keep it in event based way?

Kocal:
That's a good question, what is the new way to listen lifecycle hooks? Or it is now impossible?
We are stuck for adding Vue 3 support to Sentry Vue integration (https://github.com/getsentry/sentry-javascript/issues/2925).

Friendly ping @posva / @yyx990803 :heart: 

Thanks!

sqal:
@Kocal it's not documented but all lifecycle hooks receive component's instance as the second argument ([source code](https://github.com/vuejs/vue-next/blob/master/packages/runtime-core/src/apiLifecycle.ts#L67)) + there's `getCurrentInstance` api, so with that, you are able to do:


```js
import { onMounted, onBeforeUnmount, getCurrentInstance, ref } from 'vue'

export default {
  name: 'Component',
  setup() {
    const instance = getCurrentInstance()
    const elRef = ref(null)

    onMounted(() => {
       const picker = new Pickaday(elRef.value, {});

      onBeforeUnmount(() => {
        picker.destroy()
      }, instance)
    })
  }
}
```

Kocal:
Alright, thank you!

PR #119, opened 2020-01-21
Scoped Style Changes

yyx990803:
Provide more consistent custom CSS extensions in Single File Component scoped styles.

``` html
<style scoped>
/* deep selectors */
::v-deep(.foo) {}

/* targeting slot content */
::v-slotted(.foo) {}

/* one-off global rule */
::v-global(.foo) {}
</style>
```

[Rendered](https://github.com/vuejs/rfcs/blob/scoped-styles-changes/active-rfcs/0023-scoped-styles-changes.md)


active-rfcs/0000-scoped-styles-changes.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Targeting / Avoiding Slot Content
+
+Currently, slot content passed in from the parent are affected by both the parent's scoped styles AND the child's scoped styles. There is no way to author rules that explicitly target slot content only, or ones that do not affect slot content.
+
+In v3, we intend to make child scoped styles NOT affecting slot content by default. To explicitly target slot content, the `::v-slotted()` pseudo class can be used.
+
+## One-Off Global Rules
+
+Currently to add a global CSS rule we need to use a separate unscoped `<style>` block. We are introducing a new `::v-global()` pseudo class for one-off global rules.
```
posva: does this bring any improvement in performance or anything? I find the separate `style` easier to read and more intuitive because it's just like a regular `style` tag.
Unless `scoped` becomes the default but that would be even harder to migrate

yyx990803: Technically, yes. For webpack, an extra `<style>` block becomes a separate module with a whole pipeline (i.e. all the loaders configured for CSS are invoked for this block again). Putting them in the same block improves bundler build performance.

posva: That makes sense, I think it's worth indicating that as a benefit

ota-meshi:
Hi @yyx990803 

I have a question about how to use `::v-deep()`.

How do I write a nesting selector with deep selector?

In Vue.js 2.x, I could write:

```scss
.foo {
  & ::v-deep .bar {
    /* ... */
  }
}
```

```scss
.foo {
    & ::v-deep .bar {
        /* ... */
        &--baz {
          /* ... */
        }
    }
}
```

yyx990803:
@ota-meshi to convert to the new syntax, just put everything after `::v-deep` into its arguments (wrapped with `()`) instead:

``` css
.foo {
  & ::v-deep(.bar) {
    /* ... */
  }
}
```

ota-meshi:
I'm sorry. I made a mistake in the example code. Then rewrote the question.

The code I want to ask is:

```scss
.foo ::v-deep .bar {
    & .baz {
        / * ... * /
    }
}
```

Justineo:
I'm afraid the function syntax won't play well with nesting as the old "combinator" did. As I understand, the whole selector inside `::v-deep()` cannot be nested.

Justineo:
Oops. Sorry for closing this by mistake.

cawa-93:
The previous syntax looks a bit more flexible. Correct me if I'm wrong.

Let's say how the new syntax can be matched with `:is()` CSS pseudo-class function?
```html
<style scoped>
::v-deep :is(header, footer, p) {
    /* ... */
}
</style>
```
Or nested selectors with some preprocessors
```html
<style scoped lang="scss">
::v-deep {
    header, footer, p {
        /* ... */
    }
}
</style>
```

yyx990803:
@cawa-93 

Your first example can be written as:

``` css
::v-deep(:is(header, footer, p)) {
    /* ... */
}
```

While for nesting, the new syntax is indeed a bit less convenient. Maybe we can keep the special case of `::v-deep` as a combinator so nesting usage?

cawa-93:
Current syntax is ideal for scenarios when you want to mark a single rule (selector + styles) as deep or global.

However, how about scenarios when you need to mark more than one group of rules? I do not know how common this script is in the community, but if it is in demand then why not implement the following syntax:

```html
<style>
/* Global styles (as in Vue 2)*/
</style>

<style scoped>
/* Scoped styles */
</style>

<style scoped deep>
/* Scoped deep styles */
</style>

<style scoped slotted>
/* Scoped slotted styles */
</style>
```

As far as I understand it corresponds to the syntax of Vue 2

active-rfcs/0000-scoped-styles-changes.md
```diff
@@ -0,0 +1,116 @@
...
+
+  Compiles to:
+
+  ``` css
+  .foo[v-data-xxxxxxx-s] {}
+  ```
+
+  Notice the `-s` postfix which makes it target slot content only.
+
+- New pseudo class `::v-global()` can be used to apply global rules inside a `<style scoped>` block:
```
Justineo:
```suggestion
- New pseudo element `::v-global()` can be used to apply global rules inside a `<style scoped>` block:
```

yyx990803:
@Justineo can you make it into a single PR to the `scope-styles-changes` branch?

Justineo:
@yyx990803 Sure

thedamon:
> Current syntax is ideal for scenarios when you want to mark a single rule (selector + styles) as deep or global.
> 
> However, how about scenarios when you need to mark more than one group of rules? I do not know how this scenario is common in the community, but if it is in demand then why not implement the following syntax:
> 
> ```
> <template> ... </template>
> 
> <style>
> /* Global styles (as in Vue 2)*/
> </style>
> 
> <style scoped>
> /* Scoped styles */
> </style>
> 
> <style deep>
> /* Scoped deep styles */
> </style>
> 
> <style slotted>
> /* Scoped slotted styles */
> </style>
> ```
> 
> As far as I understand it corresponds to the syntax of Vue 2

Evan mentioned having one style block is better for webpack performance as each block would be a separate module.

It's also very nice to colocate rules that are conceptually related to each other within the same style block (similar to the readability benefits of the composition API)

yyx990803:

b-strauss:
I think the ability to nest is very important and should not be removed. While everything that can be done with the old `::v-deep` can be done with `::v-deep()` sometimes the result can be rather cumbersome.

In some cases you have third party components that you have no control over, but you want to customise them. So you have to overwrite many styles in the scoped child component. Until now these styles can all be written in one block. With the new system this does not work anymore.

```sass
.slider::v-deep
  .vue-slider-rail
    background-color: transparent

  .vue-slider-dot
    cursor: pointer

  .vue-slider-process
    background-color: transparent
    border-radius: 0

    &::after
      content: ''
      display: block
      position: absolute
      top: 0
      left: 0
      right: 0
      bottom: 0

  .vue-slider-process,
  .vue-slider-dot,
  .vue-slider-mark
    z-index: auto

  .vue-slider-mark-step
    display: none

  .vue-slider-dot-tooltip
    pointer-events: none

  .vue-slider-dot-tooltip-top
    top: 0
```

Instead you have to convert it to this which is not nice:
```sass
.slider::v-deep(.vue-slider-rail)
  background-color: transparent

.slider::v-deep(.vue-slider-dot)
  cursor: pointer

.slider::v-deep(.vue-slider-process)
  background-color: transparent
  border-radius: 0

.slider::v-deep(.vue-slider-process::after)
  content: ''
  display: block
  position: absolute
  top: 0
  left: 0
  right: 0
  bottom: 0

.slider::v-deep(.vue-slider-process),
.slider::v-deep(.vue-slider-dot),
.slider::v-deep(.vue-slider-mark)
  z-index: auto

.slider::v-deep(.vue-slider-mark-step)
  display: none

.slider::v-deep(.vue-slider-dot-tooltip)
  pointer-events: none

.slider::v-deep(.vue-slider-dot-tooltip-top)
  top: 0
```

I don't think v-deep's nesting functionality should be deprecated or removed.

mtskf:
I've found these below don't work...

```
<style deep>
/* Scoped deep styles */
</style>

<style slotted>
/* Scoped slotted styles */
</style>
```

PR #120, opened 2020-01-22
Custom getter */ setter for ref / reactive

wellrus:
[Rendered](https://github.com/wellrus/rfcs/blob/master/active-rfcs/0000-custom-getter-setter-on-ref-and-reactive.md)

### DIFF
--- a/active-rfcs/0000-custom-getter-setter-on-ref-and-reactive.md
+++ b/active-rfcs/0000-custom-getter-setter-on-ref-and-reactive.md
@@ -1,0 +1,111 @@
+- Start Date: 2020-01-22
+- Target Major Version: 3.x
+- Implementation PR: (leave this empty)
+
+# Summary
+
+Currently, it is possible to define getter and setter in computed.
+There is no such possibility for ref / reactive. However, this functionality may be useful.
+
+# Motivation
+
+Implementation will allow you to create more flexible entities without the need for additional calculated values.  
+
+# Detailed design
+
+If the programmer does not specify getter and setter, then the behavior is identical to the current one.
+
+```javascript
+import { ref, reactive } from 'vue'
+
+let primitive = ref(0, {
+    get: function (value) {
+        return value * 2;
+    },
+    set: function (value) {
+        if (value < 0) {
+            value = 0;
+        }
+
+        return value;
+    }
+})
+
+let obj = reactive(
+    {
+        tracked: 0,
+        untracked: 0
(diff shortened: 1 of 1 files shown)


ycmjason:
Getters and setters are still available in `computed` if I am not wrong. So there is no need to make ref and reactive too overly complicated.

jacekkarczmarczyk:
But you need additional data to store your value

LinusBorg:
I don't see an issue with that. or to put it another way, if you want to keep this vallue internal, create a one-line wrapper around `computed`.

posva:
As mentioned this over complicates the api with no gain (memory usage won't be different).
On top of that this can be implemented locally:

```js
function computedRef(value, { get, set }) {
  const refValue = ref(value)
  return computed({
  // some wrapping of get and set
 })
}
```

ycmjason:
@posva

I believe that there isn't a way to have deep getter/setter tho? Would it be useful if computed provides a way to get/set deeply? Currently the getter and setter will only be triggered at `.value` level?

posva:
@ycmjason  what do you mean by having a deep getter/setter? How is that related to this RFC?

ycmjason:
It is related to this RFC since what's mentioned in the RFC could not be fully implemented if computed can only return Ref.

Let me rephrase my previous comment. It wasn't very clear. Perhaps it would be useful to allow computed to return an reactive object where setting/getting of its entries will trigger the getter/setter? Much like how proxy traps works?

ycmjason:
Although conceptually that's the same as an reactive with many computed values, but that wouldn't allow new properties to be added to the reactive?

yyx990803:
Technically this is indeed something that the current API does not support, however what I find not convincing enough is the Motivation section. It is too brief, and "may be useful" simply doesn't justify feature additions like this. I would like to see what **real world use cases** this can solve or make simpler.

donnysim:
The only real use case I could see is having value validator as in:

```js
price: {
  get() {
    return this._price;
  },
  set(value) {
    if (isNaN(Number(value))) {
      this._price = 0;
    } else {
      this._price = Number(value);
    }
  },
},
```
or global state getter/mutator without having to create a bunch of computed properties, but if what @posva said would work the same then I don't really see the need of it other than convenience. Maybe just the thought of computed property having a setter sounds weird, as it's more often associated with just retrieving a value (very few knew they can have a setter on a computed from my encounters/helping others).

PR #124, opened 2020-01-29
API Docs Standardisation

sustained:
[Rendered](https://github.com/sustained/rfcs/blob/api-docs-standardisation/active-rfcs/0000-api-docs-standardisation.md)

This RFC relates to the inconsistencies/lack of cohesion between the different VuePress-based API docs for Vue, Vuex, Vue Router, Vue CLI etc.

### DIFF
--- a/active-rfcs/0000-api-docs-standardisation.md
+++ b/active-rfcs/0000-api-docs-standardisation.md
@@ -1,0 +1,72 @@
+- Start Date: 2020-01-29
+- Target Major Version: 3.x
+- Reference Issues: N/A
+- Implementation PR: (leave this empty)
+
+# Summary
+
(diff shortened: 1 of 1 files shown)


yyx990803:
This is definitely a good area for improvement (especially as we are reworking the core docs for v3).

We will most likely explore generating API references automatically from TS typings (we can already generate `api.json` files using api-extractor for `vue-next`). Once we find that working well we can apply that workflow to sub projects as well.

sustained:
Thanks for the response.

That sounds promising, I will definitely have to take a closer look at this "api-extractor" and the JSON it generates for vue-next. Just to clarify, you are [referring to this Microsoft tool, yes](https://github.com/microsoft/rushstack/tree/master/apps/api-extractor)?

yyx990803:
@sustained yes, `vue-next` uses it to rollup types and extract API metadata.

CyberAP:
@yyx990803 I think people would also appreciate if Vue 3 had examples written both in JS and TS, where you could toggle between those two.

yyx990803:
@CyberAP TS code and JS code will look almost the same as most of the APIs provide decent inference.

PR #125, opened 2020-01-30
Change watch Options API to trigger immediately 

lmiller1990:
[Rendered](https://github.com/lmiller1990/rfcs/blob/master/active-rfcs/0000-change-default-options-watch-api.md)

### DIFF
--- a/active-rfcs/0000-change-default-options-watch-api.md
+++ b/active-rfcs/0000-change-default-options-watch-api.md
@@ -37,9 +37,37 @@ Although this component is valid in both Vue 2 and Vue 3, it behaves differently
 - If this component is mounted in Vue 2, the `message` handler will not be triggered until `message` is changed for the first time. A user can opt to fire the `handler` immediately by passing `immediate: true`. 
 - In Vue 3, the `message` handle will trigger immediately; this is the default behavior in Vue 3 - the opposite of Vue 2. To opt out of this behavior in Vue 3, the user can pass a `{ lazy: true }` option.
 
+It is relevant to note that `watch` in Vue 3 is not fired synchronously (even for the initial call) - they are in fact deferred until the component is mounted. Even with immediate by default, watchers won't fire during SSR unless you explicitly use { flush: 'sync' } in the watcher's options.
+
 # Motivation
 
 It is better to be consistent across the Composition API and the Options API. While they look different, the two APIs Vue provides are just two different ways of accomplishing the same thing.
+
+One common occurrence is the following:
+
+```js
+created() {
+  this.fetchData(this.id)
+},
+watch: {
+  id: fetchData
+},
+methods: {
+  fetchData(id) {
+    // ...
+  }
+}
+```
+
+An example of this is an application with a route like `/users/:id`, where you want to load the user on the initial load, and whenever the `id` param changes. With `immediate: true` as the default, the above can be written like this:
+
+```js
+watch: {
+  id(id) {
+    // fetch data...
+  }
+}
+```
 
 # Detailed design
 
--- a/active-rfcs/0000-change-default-options-watch-api.md
+++ b/active-rfcs/0000-change-default-options-watch-api.md
@@ -7,9 +7,9 @@
 
 Vue 3 introduces the Composition API, and as part of that, a `watch` method. More information and examples can be found [here](https://vue-composition-api-rfc.netlify.com/api.html#watch).
 
-Currently, the code for this feature is shared with the code used by the `watch` options API. However, the default behaviors are slighly different; the new `watch` method will trigger immediately after the component is mounted. This is not the case for Vue 2.X; this behavior is opt in, via the `immediate` option.
+In Vue 3, the code for the `watch` Composition API is shared with the code used by the `watch` Options API - they exhibit the same behavior. This default behavior is slighly different from Vue 2. The new `watch` Composition API will trigger immediately after the component is mounted. This is not the case for Vue 2; this behavior is opt in, via the `immediate` option.
 
-This RFC proposes to keep this change moving forward in Vue 3.x; `watch` handlers will be invoke immediately, both when using the `watch` method and the `watch` options API.
+This RFC proposes to keep this change moving forward in Vue 3 `watch` handlers will be invoke immediately, both when using the `watch` method and the `watch` Options API.
 
 # Basic example
 
@@ -32,33 +32,30 @@ const App = {
 }
 ```
 
-If this component is mounted in Vue 2 the `message` handler will not be triggered until `message` is changed for the first time. A user can opt to fire the `handler` immediately by passing `immediate: true`. 
+Although this component is valid in both Vue 2 and Vue 3, it behaves differently.
 
-Currently in Vue 3, the `message` handle will trigger immediately; this is the default behavior in Vue 3 - the opposite of Vue 2. To opt out of this behavior in Vue 3, the user can pass a `{ lazy: true }` option.
+- If this component is mounted in Vue 2, the `message` handler will not be triggered until `message` is changed for the first time. A user can opt to fire the `handler` immediately by passing `immediate: true`. 
+- In Vue 3, the `message` handle will trigger immediately; this is the default behavior in Vue 3 - the opposite of Vue 2. To opt out of this behavior in Vue 3, the user can pass a `{ lazy: true }` option.
 
 # Motivation
-
-Why are we doing this? What use cases does it support? What is the expected
-outcome?
 
 It is better to be consistent across the Composition API and the Options API. While they look different, the two APIs Vue provides are just two different ways of accomplishing the same thing.
 
 # Detailed design
 
-The `watch` options API should behave the same as the `watch` method provided by the Composition API.
+The `watch` Options API should behave the same as the `watch` method provided by the Composition API.
 
 # Drawbacks
 
 The main cost of keeping this breaking change is the migration cost for existing codebases. Having `watch` trigger immediately when upgraing to Vue 3 may cause unintended behavior. The user will need to add `{ lazy: false }` to each watch API they do not want to trigger immediately.
 
+It is possible many commonly used modules use the old `watch` behavior; this could break, or at least change the behavior of many applications and libraries.
+
 # Alternatives
 
-An alternative would be to implement an `immediate` option that defaults to `false` for the `watch` Options API in Vue 3.
+An alternative would be to implement an `immediate` option that defaults to `false` for the `watch` Options API in Vue 3. This would mean no breaking change, but a larger API surface with discrepancy between the Options API and Composition API.
 
 # Adoption strategy
-
-If we implement this proposal, how will existing Vue developers adopt it? Is
-this a breaking change? Can we write a codemod? Can we provide a runtime adapter library for the original API it replaces? How will this affect other projects in the Vue ecosystem?
 
 - The user will need to add `{ lazy: false }` to each watch API they do not want to trigger immediately in their existing codebases. 
 - A codemod to assist with this migration should be possible in most cases. 
--- a/active-rfcs/0000-change-default-options-watch-api.md
+++ b/active-rfcs/0000-change-default-options-watch-api.md
@@ -1,0 +1,69 @@
+- Start Date: 2020-01-30
+- Target Major Version: 3.x
+- Reference Issues: https://github.com/vuejs/vue-next/issues/671
+- Implementation PR:
+
+# Summary
+
+Vue 3 introduces the Composition API, and as part of that, a `watch` method. More information and examples can be found [here](https://vue-composition-api-rfc.netlify.com/api.html#watch).
+
+Currently, the code for this feature is shared with the code used by the `watch` options API. However, the default behaviors are slighly different; the new `watch` method will trigger immediately after the component is mounted. This is not the case for Vue 2.X; this behavior is opt in, via the `immediate` option.
+
+This RFC proposes to keep this change moving forward in Vue 3.x; `watch` handlers will be invoke immediately, both when using the `watch` method and the `watch` options API.
+
+# Basic example
+
+This is a simple component that works as-is with Vue 2.x and 3.x.
+
+```js
+const App = {
+  data() {
+    return {
+      message: "Hello"
+    };
+  },
+  watch: {
+    message: {
+      handler() {
+        console.log("Immediately Triggered")
+      }
+    }
+  }
+}
+```
+
+If this component is mounted in Vue 2 the `message` handler will not be triggered until `message` is changed for the first time. A user can opt to fire the `handler` immediately by passing `immediate: true`. 
+
+Currently in Vue 3, the `message` handle will trigger immediately; this is the default behavior in Vue 3 - the opposite of Vue 2. To opt out of this behavior in Vue 3, the user can pass a `{ lazy: true }` option.
+
+# Motivation
+
+Why are we doing this? What use cases does it support? What is the expected
+outcome?
+
+It is better to be consistent across the Composition API and the Options API. While they look different, the two APIs Vue provides are just two different ways of accomplishing the same thing.
+
+# Detailed design
+
+The `watch` options API should behave the same as the `watch` method provided by the Composition API.
+
+# Drawbacks
+
+The main cost of keeping this breaking change is the migration cost for existing codebases. Having `watch` trigger immediately when upgraing to Vue 3 may cause unintended behavior. The user will need to add `{ lazy: false }` to each watch API they do not want to trigger immediately.
+
+# Alternatives
+
+An alternative would be to implement an `immediate` option that defaults to `false` for the `watch` Options API in Vue 3.
+
+# Adoption strategy
+
+If we implement this proposal, how will existing Vue developers adopt it? Is
+this a breaking change? Can we write a codemod? Can we provide a runtime adapter library for the original API it replaces? How will this affect other projects in the Vue ecosystem?
+
+- The user will need to add `{ lazy: false }` to each watch API they do not want to trigger immediately in their existing codebases. 
+- A codemod to assist with this migration should be possible in most cases. 
+- It should be possible to write a runtime adapter to maintain the Vue 2 behavior.
+
+# Unresolved questions
+
+N/A


yyx990803:
I think we should include the information in [this comment](https://github.com/vuejs/vue-next/issues/671#issuecomment-579827250) and [this one](https://github.com/vuejs/vue-next/issues/671#issuecomment-580310207)

lmiller1990:
I will update the RFC within 24h to include that information

CyberAP:
The motivation part does not look that convincing to me. There is a single use case provided but watch is used for lots of things apart from that. I have collected a short list of examples from my project where watch is not used immediately:

- Triggering pagination request on parameters change (the list is already present on the first render)
- Debouncing value
- Emitting native events on value change (for example on a hidden input that does not emit event out of the box)
- Autosizing on value change (updating and instantiating are different methods, also instantiating happens on the mounted hook, not on the created one, to get an element reference)
- Syncing with window.history (initial state is always in sync)
- Searching with network requests (initial state is always empty)
- Controlling scroll

Most of these cases have common similarities:

- Initial state is already correct (pagination, history)
- Triggering on `created` would provide undesirable effects (debouncing, autosizing, searching)

I have found 13 `watch` occurrences overall in my project and the ratio between immediate and lazy watchers are 3 to 10. I am yet to see this change being motivated enough to be valuable for me since it certainly doesn't fall into a common pattern for my project.

lmiller1990:
I actually looked at my own personal/work projects and found a similar pattern: around 30% of watchers use `immediate` - the remaining 70% does not, and some of them would have unintended consequences if they were `immediate`.

This is just an RFC to outline the current behavior of Vue 3, and get feedback and comments - I'm leaning towards `watch` as non immediate by default as well.

donnysim:
Same here, I find it extremely rare to add an immediate watcher, I find myself using it once or twice in the whole project compared to normal watchers.

when an immediate watcher is needed, I find it that it's because:
1. is better in this case only;
2. some kind of repettetive task that can be extracted into mixin/use.

Also if the options api runs the watcher async after mounted, I dont really see why I would want to use immediate watcher without manually triggering the function, unless the perceived difference is negligable.

yyx990803:
@CyberAP it seems a lot of your use cases are actually simpler with immediate by default:

- Triggering pagination: you'd likely want to request the initial page on mount, no?

  ``` js
  watch(() => requestPage(props.pageIndex))
  ```

- Autosizing on value change: not sure how different the logic would be - but you can perform conditionals in watchers. Also, v3 `watch` does get fired at `mounted`, not `created` (so technically, it's not equivalent to immediate mode in v2, I'd actually call it "eager" watchers)

- Searching with network requests: you actually should always have a condition to avoid firing a request when the value is empty:

  ``` js
  watch(() => {
    if (search.value) {
      doSearch(search.value)
    }
  })
  ```

- Debouncing value: not sure how exactly you are doing this, could use an example to help me understand it better.

- Emitting native events on value change / syncing with `window.history`: these are valid use case for `{ lazy: true }`.

---

I think there are a lot of existing `watch` usage in 2.x are somewhat guided by the fact that 2.x `watch` is lazy by default. A decent percentage of the use cases can actually be refactored into a simpler form with a "run on mounted by default" version of the API.

But again, I agree with the fact that the verb `watch` simply entails "lazy", and it feels weird to have it run on mount by default. Maybe instead of the lazy/immediate options, we simply separate them into two different APIs:

``` js
// BEFORE
watch(() => { /* eager */ })
watch(() => { /* lazy */ }, { lazy: true })

// AFTER
runAndWatch(() => { /* eager */ })
watch(() => { /* lazy */ })
```

I personally find having these two separate APIs makes the intention of the code and the difference between the two usage types much clearer. This would also keep `watch` behavior exactly the same as 2.x. Obviously the naming of the new method is still up for discussion (I just find `runAndWatch` the most explicit about what it does)...

cawa-93:
Continuing the dialogue.

How about a little "fantastic" sentence: Is it possible to create a wrapper function to combine hooks and watchers?
```js
setup() {
    const someMethod = method(() => {/* some actions */}, {
        onCreated: true, // onMounted, etc.
        watch: true
    })

    return { someMethod }
}
```
Or no wrapper function at all
```js
setup() {
    const someMethod = () => {/* some actions */}
    someMethod.onCreated = true, // onMounted, etc.
    someMethod.watch = true

    return { someMethod }
}
```

I am not sure whether such an implementation is possible and how convenient it will be. What is the community opinion?

LinusBorg:
@cawa-93 this seems only tangentialy related to this RFC and would be better discussed in its own RFC or an issue as a pre-discussion.

kiaking:
@yyx990803 I really like the `runAndWatch` approach. I think it would eliminate all of these discussions once and for all.

CyberAP:
@yyx990803 

- Pagination does not always have to perform a request on `mounted`, for example in a case when data is already present on the initial load. Checking for data presence would solve this problem I guess but then you'll have to deal with checking for `null` or `undefined`.

- Debouncing is a case of derived state, example can be found here: https://github.com/CyberAP/vue-debounced/blob/master/src/component.js
  Checking for prop and local state equality could also work here although it's risky to do this with async operations.

---

As to separating watchers into two different options I think that would bring even more confusion considering we now have so many options to define a watcher:

- lazy `watch` option
- eager `runAndWatch` option
- lazy `$watch` instance method
- eager `$watch` instance method with options
- eager `watch` exported function for `setup`
- lazy `watch` exported function for `setup` with explicit dependencies

I like the idea of having two different functions because they work different from each other in the first place.

`watch` requires explicit dependencies and cannot be run without having at least one.
`runAndWatch` is the opposite of that and does not require any dependencies and performs tracking on the first run (event its name already suggests that).

So if there are going to be two separate watch functions I would like them to be present in all contexts and work the same way:

- lazy `watch` option (as is)
- eager `runAndWatch` option (an array of functions)
- lazy `$watch` instance method (as is)
- eager `$runAndWatch` instance method with no requirement for explicit deps
- lazy `watch` exported function for `setup` that does require explicit deps (that's different from its current behaviour)
- eager `runAndWatch` exported function for `setup`

If the new reactivity system does not require setting `deep: true` for deep watch then all these methods should no longer require any extra options at all.

yyx990803:
If we split them into two functions then yeah, we probably gonna remove `lazy/immediate` options. And I also like the idea of `watch` always requires watching source, while `runAndWatch` directly runs a function.

However:

-  `deep` has to remain an option due to its performance implications (and it would only apply to `watch`).

- There is another option (new in v3) called `flush` which controls the timing for the callback to be called in relation to the DOM updates.

donnysim:
@yyx990803 what do you mean by "requires watching source"? This is probably a bit sidetracked, but with watchers could there a better way to achieve multi-reference watchers, like a variable could not be accessed during the watch execution? for example: in vue 2:
```js
const removeValueUpdateWatcher = this.$watch(() => {
  // Register dependency tracking
  [this.value, this.items, (this.ajaxAdapter && this.ajaxAdapter.sourceItems), this.textField, this.keyField]
    .filter(x => Boolean(x));

  return Date.now();
}, () => {
  this._updateInternalValue(this.value);
}, { immediate: true });
```
where you just loop through a list of variables for Vue to register as watch dependencies.

yyx990803:
@donnysim v3 `watch` already supports watching multiple sources, but that's for composition API.

For v2 style watchers you could always have done `this.$watch(() => [this.foo, this.bar, this.baz], (latestValues) => {})`

michaeldrotar:
I'm in the camp where I usually design such functions to be non-destructive and letting them run when defined saves me the usual headache of having to possibly duplicate code to provide an initial good state.

That said, I wanted to offer an alternate route. Generally in frameworks I've dealt with, watch-like functions are always lazy, but you can simply call the function at the same time you set it up to watch cause, afterall, it's just a normal js function.

```js
function foo(value) {};
watch('something', foo);
foo(this.something);
```

with composition api, this can work in nearly the same way
```js
const foo = () => { /* do stuff */ };
watch(foo)
foo();
```

or for a slimmer syntax, return the handler (or a wrapped version if needed)
```js
const foo = watch(() => { /* do stuff */ });
foo();
```

for object api (is that what it's called?) - I'm not sure if they're exposed currently, but they could be something like this
```
module.exports = {
  watch: {
    foo: function() {
      // do stuff
    }
  },
  mounted: function() {
    this.$watchers.foo();
  }
};
```

Perhaps rather than throwing more methods or syntax at the problem, maybe using more standard js is something people might find intuitive.

yyx990803:
@aztalbot some good points there!

- RE longer name: we may still be able to change to a shorter name before getting to beta. But with Vue 3's typing you'll be able to get auto import/completion even using JS!

- RE adding to the API: if we separate `runAndWatch` and `watch`, we are not really adding to the API, but rather splitting an overloaded `watch` into two more discrete functions. The underlying capabilities/complexity did not change - they were just implicitly hidden behind different function signatures.

Once separated, `runAndWatch` will only accept a single callback and will always be eager, while `watch` will always require a source + a callback (can be eager OR lazy, although defaults to lazy).

---

The point about `async setup()` usage alongside watchers is interesting. `runAndWatch` returning a stop handle was simply being consistent with v2. However, it might be worth it change it slightly so that it returns **a Promise with an additional `stop` method, which resolves to the return value of the first call**. Think of it as an "awaitable handle":

``` js
const handle = runAndWatch(async () => { /* fetch data */ })

// the handle is a Promise that can be awaited (which resolves to the return value of the first run):
await handle

// and the handle also exposes the `stop` function:
handle.stop()
```

So your last example can be written as (omitting the emit part):

``` js
const MyComponent = {
    async setup(props) {
        const user = ref(null)
        await runAndWatch(async () => {
          user.value = await fetchUser(props.id)
        })
        return { user }
    }
}
```

donnysim:
Adding additional functions to a promise is kind of sketch from my experience. It can get lost real fast, e.g.:
```js
const handle = runAndWatch(async () => { /* fetch data */ }).then(result => JSON.parse(result));

handle.stop(); // undefined function, because stop exists on the initial promise and cannot be chained
```

unless it's kind of a fake promise that overrides `then` etc. afaik.

yyx990803:
I don't think it makes sense to do anything on the `runAndWatch` return value other than `await`... e.g. your `JONS.parse` call should be inside the `runAndWatch` callback, not outside of it.

CyberAP:
Not sure why would you need to await an immediate watcher when you can easily do the same thing with an eager watcher?

```js
async setup(props) {
  const user = ref(null);
  const loadUser = async function() {
    user.value = await fetchUser(props.id);
    if (props.id === null) unwatch();
  }
  const unwatch = watch(props.id, loadUser);
  await loadUser();
  return { user };
}
```

IMO, this is an unnecessary complication of an already complex function.

yyx990803:
@CyberAP

Because you might be tempted to start with an eager watcher due to its simplicity:

``` js
setup(props) {
  const user = ref(null)
  runAndWatch(async () => {
    user.value = await api.fetch(`/users/${props.id}`)
  })
}
```

But later on decided that you want to make `setup()` async, and suddenly you have to refactor to use a lazy watcher because there's no way to await an eager watcher's first call:

``` js
async setup(props) {
  const user = ref(null)

  const fetchUser = async (id) => {
    user.value = api.fetch(`/users/${props.id}`)
  }
  watch(() => props.id, fetchUser)
  await fetchUser(props.id)
}
```

See the duplication and changes needed here?

Where as if we make eager watch await-able, the only change needed is adding `async` to setup and `await` to the watch call:

``` js
async setup(props) {
  const user = ref(null)
   await runAndWatch(async () => {
    user.value = await api.fetch(`/users/${props.id}`)
  })
}
```

CyberAP:
I like the idea of `watchAndRun` returning an abstract handler, but I'm not a fan of extending Promises. That brings more confusion than convenience IMO. When you're working with Promises the last thing you want to deal with are some custom properties on the Promise object. Because in that case the behaviour is unpredictable (will it resolve or reject?).

Also `watch` and `watchAndRun` do not imply a meaningful return (only watcher controls are expected if any), they are purely side-effects. I don't think that awaiting on side-effects is a good pattern to have.
I see the immediate benefits of having a simple way of awaiting watcher callback, but I think that works only if you use it exactly like in that case. When you'll have multiple watchers, mixed with computeds, and external composition functions that would be difficult to understand at the first glance. (what does `runAndWatch` return, why we wait for the watcher, how is the watcher result used) So I would consider that code write-only, rather than a simplification.
I can also easily imagine people using async `runAndWatch` as the `setup` return. That would bring even more confusion.

As to the handler, I'd much rather have `watchAndRun` to share similarities from `setTimeout`, where you have a timeout id and cancel it with another function.
So for watchers that could look like this:

```js
setup(props) {
  const handle = watchAndRun(() => {
    console.log(`props changed`, props);
    if (props.once) {
      stopWatch(handle);
    }
  });
}
```

yyx990803:
Instead of patching Promise (which does feel dirty), how about explicitly exposing the return value of the first call on the handle:

``` js
async setup(props) {
  const user = ref(null)
  await runAndWatch(async () => {
    user.value = await api.fetch(`/users/${props.id}`)
  }).result

  // OR
  const { result, stop } = runAndWatch(...)
  stop()
  await result
}
```

CyberAP:
> Instead of patching Promise (which does feel dirty), how about explicitly exposing the return value of the first call on the handle:

This feels a little bit like using a `.value`, which has a kind of a hackish feeling to it. Also my guess is that people will try to use it like this:

```js
async setup(props) {
  return await runAndWatch(async () => {
    const user = await api.fetch(`/users/${props.id}`)
    return { user }
  }).result
}
```

yyx990803:
@aztalbot I realized the problem too (first run timing) when I was experimenting with the change. Now I think it might make sense to always invoke the first run synchronously. A side effect of this is that the first run won't have access to the DOM. Solution though is to simply setup the watcher in `onMounted`:

``` js
watch(() => {
  // no access to the DOM since this will be called synchronously
})

onMounted(() => watch(() => {
  // has access to the DOM
}))
```

CyberAP:
With the synchronous watch you could simply say:

```js
async setup(props) {
  const state = reactive({ user: null, request: null })
  runAndWatch(async () => {
    state.request = api.fetch(`/users/${props.id}`)
    state.user = await state.request
  })
  await state.request
  return state
}
```

yyx990803:
@CyberAP that's still quite some juggling compared to a simple `await`.

michaeldrotar:
I was going to suggest `{ result, stop } = runAndWatch(...)` but `await runAndWatch(...).result` feels really bad... I didn't even notice the tailing `.result` for a bit and wondered _what voodoo is this!!_ -- I could see that causing alot of confusion

---

I think we're saying patching promises is dirty because of the chaining...?

```js
async setup(props) {
  const user = ref(null);
  const task = runAndWatch(async () => {
    return await api.fetch(`/users/${props.id}`);
  }).then((user) => {
    user.loadedAt = new Date();
    return user
  });
  if (props.quickStop) {
    task.stop(); // would this work? I would expect not due to the `then`
  }
  user.value = await task;
  return { user };
}
```

I'm not seeing a cleaner solution though.. and I believe stopping a promise is not part of the spec yet?
So if `runAndWatch` returns a thenable with some extras.. and if you need that original reference.. that seems like maybe a not horrible compromise. How common is it to need to stop a watch manually?

```js
async setup(props) {
  const user = ref(null);
  const task = runAndWatch(async () => {
    return await api.fetch(`/users/${props.id}`);
  });
  const userPromise = task.then((user) => {
    user.loadedAt = new Date();
    return user
  });
  if (props.quickStop) {
    task.stop(); // this should work
  }
  user.value = await userPromise;
  return { user };
}
```

Here's another crazy idea.. what if we mod the function rather than the result?

```js
async setup(props) {
  const user = ref(null);
  const getUser = async () => {
    return await api.fetch(`/users/${props.id}`);
  };
  const task = runAndWatch(getUser).then((user) => {
    user.loadedAt = new Date();
    return user
  });
  if (props.quickStop) {
    getUser.stop(); // this should work
  }
  user.value = await task;
  return { user };
}
```

Or some other crazy departure... building off @CyberAP 's setTimeout/cancelTimeout idea...
```js
const getUser = createWatcher(async () => { ... }); // not watching yet
const user = await runWatcher(getUser); // single run..?
startWatcher(getUser); // how would it know which props to watch on? maybe it would error unless it had been ran once already to figure that out
stopWatcher(getUser);
```

CyberAP:
I dig the watcher constructor idea, but it's a little too verbose right now, how about this:

```js
async setup(props) {
  const user = ref(null)
  const userIdWatcher = watcher(async () => {
    user.value = api.fetch(`/users/${props.id} `)
  })
  await watchWithPromise(userIdWatcher)
  onMounted(() => stopWatch(userIdWatcher))
  return { user }
}

```

yyx990803:
That creates a ton of extra imports and I think we do have to limit the verbosity to a certain limit. Also I think we are getting a bit off-track on the async/await issue.

Back to the original eager/lazy problem: I was thinking maybe we can avoid a name change/split by going with different behavior for the two signatures:

- `watch(source, cb)` is lazy by default. This aligns with 2.x behavior, and also with the assumption that `cb` is only called if `source` has changed. It can be made eager by explicitly passing `{ immediate: true }`.

- `watch(fn)` is always eager. It can only be eager because it must run `fn` to collect dependencies.

- Note that when the `source` is a getter function, e.g. `watch(() => props.id, cb)`, the getter function is also called eagerly (in order to collect deps). Only the `cb` is invoked lazily. When `source` is a ref, Vue also internally creates a getter to access its `value`. So in some way, it can be thought of as "the first argument is always called/evaluated eagerly". Or you can think of the second signature as "a watcher where the side effect is performed directly in the getter".

CyberAP:
Having `watch` to accept a single function doesn't make much sense to me. When we say `watch(source, cb)` it's clear that we watch source to then run a callback. With a single argument it's not clear what are we watching and what for (should our callback return dependencies to track?). Basically `watch` with a single argument states that its callback is reactive, so I'd probably prefer something like this: `reactive(cb)`. Unfortunately `reactive` is already taken and I can't think of a better name for it.

donnysim:
With single callback, will it end up as 
```js
const removeValueUpdateWatcher = this.$watch(() => {
  // Register dependency tracking
  [this.value, this.items, (this.ajaxAdapter && this.ajaxAdapter.sourceItems), this.textField, this.keyField]
    .filter(x => Boolean(x));

  // do stuff
});
```
to watch multiple sources?

CyberAP:
With a single callback you do not explicitly define watch targets, they are tracked automagically (like in computeds). And that's the problem with the naming of the function, since it's not clear what exactly are we watching. It's a reactive callback, watcher is an implementation detail.

michaeldrotar:
Overloaded functions doing different things is pretty common and intuitive in js-land for me (thanks jquery >_<)

I think we could argue clarity vs verbosity for days and not reach a happy medium.. just gotta pick something at some point that meets the functional needs and wait to see where people struggle with it -- they'll always struggle somewhere that can be improved later.

Coming from Ember, I struggle with `source` being a function -- I totally understand why it is, of course.. but it's a weird paradigm, right? Let's call this function.. that reads some variables.. but doesn't do anything with them. Or _shouldn't_ do anything with them... it could and it'd still work. But it's like.. read these variables and generate some side effects to setup watchers (I'm pretty sure that's how it works).

Side effects on _reads_ is generally major code smell.. but it works and it's slim and it's teachable so we like it. And (I think) since you _could_ do something like
```js
const foo= ref(1);
const bar = ref(2);
watch(() => foo.value = foo.value + 1, () => bar.value = bar.value * 2);
```

It doesn't seem a stretch to me to teach this variation
```js
const foo = ref(1);
const bar = ref(2);
watch(() => {
  foo.value = foo.value + 1;
  bar.value = bar.value * 2;
});
```

Actually, back up a minute... in the second.. isn't `bar` then watched as well? So every time bar changes.. it multiples bar by 2? I don't think that's the intention.

Here's another naming idea I was going to toss out...

```js
watch(source, callback, { immediate: false });
immediate(callback);
```

But more importantly.. with `runAndWatch` or any of the derivatives, how are we separating those values that should re-call those functions vs those values that shouldn't? Maybe I missed something.

🤔

yyx990803:
@michaeldrotar technically, "side effects" here do not include state mutations. It usually means API requests, direct DOM mutations. Anything related to state should be expressed as a declarative graph using computed properties, if possible.

As in Vue 2.x, the best practice is avoid mutating state in watchers.

michaeldrotar:
Hm I think I've had a misunderstanding of how watchers work.

When we had examples like `watch(() => a + b, doStuff)` I pictured that just anything that was read was then watched and the return value didn't matter.

But re-reading the [v2 docs](https://vuejs.org/v2/api/#vm-watch) and [composition-api docs](https://vue-composition-api-rfc.netlify.com/api.html#watch) ... I think I totally misunderstood.

So in this example...
```js
const num = ref(5);
watch(() => Math.abs(num.value), doStuff);
num.value = -5;
```
it would never trigger "do stuff", right? Cause the absolute value stays the same.. but setting it to 6 or -20 would trigger it.

-----

So question..

Getting back to the user api example:
```js
const user = ref(null)
runAndWatch(() => {
  user.value = await api.fetch(`/users/${props.id}`)
})
```
This makes sense.. when `props.id` changes, re-execute this function.
But how does it know to execute _only_ when `props.id` changes and not when `user` changes?

And what about in this example...?

```js
const foo = ref(1);
const bar = ref(2);
runAndWatch(() => {
  bar.value = bar.value + foo.value;
});
```

This looks like an infinite loop.. like it should run when either `bar` or `foo` changes. The intention is probably actually this

```js
watch(foo, () => {
  bar.value = bar.value + foo.value;
}, { immediate: true });
```

The `runAndWatch` might end up being less clear... and the 3 arg `watch` is kind of hard to read.
Personally, I think I'd want to do this

```js
const increment = () => {
  bar.value = bar.value + foo.value;
};
watch(foo, increment);
increment();
```

CyberAP:
While reading through the latest changelog I noticed that the `watch` behaviour has been adjusted slightly. You can now call `watch(effect, options?)` to launch an effect. So why not just call the function itself an `effect`? I know it's similar to React's `useEffect`, but they actually do the same thing and that'd be easier for people from React world to migrate to Vue!

```js
watch(source, callback, options?) // for lazy watchers

effect(callback, options?) // for eager watchers
```

No `immediate` nor `lazy` option is required anymore.

For component options that would be:

```js
{
  watch: {
    [source](prev, next) {},
    [source]() {
      handler: function callback(prev, next) {},
      ...options?
    }
  },
  effects: [
    function callback() {},
    {
      handler: function callback() {},
      ...options?
    }
  ]
}
```

-----

I'd also imagine for "watchable" callbacks defined in component's options to contain a stop handler as a function argument to be able to stop it in place.

```js
{
  watch: {
    [source](prev, next) {},
    [source](stop) {
      handler: function callback(prev, next) {},
      ...options?
    }
  },
  // OR
  watch: {
    [source](prev, next, stop) {},
  },
  effects: [
    function callback(stop) {},
    {
      handler: function callback(stop) {},
      ...options?
    }
  ]
}
```

yyx990803:
@CyberAP that's something I've considered too. It would require some substantial change to the Composition API RFC though.

yyx990803:
As of [3.0.0-alpha.6](https://github.com/vuejs/vue-next/blob/master/CHANGELOG.md#300-alpha6-2020-02-22) the single callback API has been separated to `watchEffect` for the following reasons:

- Having different lazy/eager behavior under the same name based on different signatures is less explicit and can be confusing for beginners.

- The name `effect` doesn't properly signify the fact that the effect will be reactively re-run, nor does it imply its connection to `watch`, which 2.x users are already familiar with.

- `watchEffect`, while slightly longer, is more explicit about what exactly it is doing. With typing and IDE support, methods can also be imported automatically so API name length shouldn't be a major concern.

PR #126, opened 2020-02-01
Attr coercion behavior change

Justineo:
This RFC describes the changes we intend to make about attribute coercion behavior.

[Rendered](https://github.com/vuejs/rfcs/blob/attr-coercion/active-rfcs/0000-attribute-coercion-behavior.md)


jacekkarczmarczyk:
How `attr="attr"` should behave? `draggable="attr"` or `draggable=""` or `draggable="true"`?

See:
https://github.com/vuetifyjs/vuetify/blob/3a510e8af57d76da660c6b4a407f7940fefae5cb/packages/vuetify/src/mixins/detachable/index.ts#L126

Justineo:
> How `attr="attr"` should behave? `draggable="attr"` or `draggable=""` or `draggable="true"`?

This RFC only affects attributes defined by `v-bind`. Otherwise the attribute won't be coerced.

yyx990803:

xSorc:
@Justineo link is broken. I suppose this is valid link?
https://github.com/vuejs/rfcs/blob/master/active-rfcs/0024-attribute-coercion-behavior.md

PR #127, opened 2020-02-01
Improve Vue Test Utils async workflow

dobromir-hristov:
Improve the VTU experience after the removal of `sync` mode in `beta-29`

```js
    const wrapper = mount(Component)
    await wrapper.find('.element').trigger('click')
    expect(wrapper.emitted('event')).toBeTruthy()
```

[Rendered](https://github.com/vuejs/rfcs/blob/a872c27baa7bf604148639e8df941fd4aefcedd4/active-rfcs/0000-vtu-improve-async-workflow.md)

### DIFF
--- a/active-rfcs/0000-vtu-improve-async-workflow.md
+++ b/active-rfcs/0000-vtu-improve-async-workflow.md
@@ -1,0 +1,133 @@
+- Start Date: 2020.02.01
+- Target Major Version: VTU beta-0.3x/1.x
+- Reference Issues:
+- Implementation PR:
+
+# Summary
+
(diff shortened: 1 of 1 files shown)


Justineo:
How can we assert certain behavior is performed synchronously? eg. After a mouse click, a custom event should be emitted without an async task.

dobromir-hristov:
Just don't await? You can still use `emitted` to track events.

dobromir-hristov:
This has already been implemented in VTU-next - https://vuejs.github.io/vue-test-utils-next-docs/api/

We will implement it in VTU-beta as soon as possible.

PR #135, opened 2020-02-26
Component's public API

CyberAP:
Introduce a new `expose` option to declare component's public API.

[Rendered](https://github.com/CyberAP/rfcs/blob/components-public-interface/active-rfcs/0000-component's-public-api.md)

### DIFF
--- a/active-rfcs/0000-component's-public-api.md
+++ b/active-rfcs/0000-component's-public-api.md
@@ -1,0 +1,307 @@
+- Start Date: (fill me in with today's date, 2020-02-27)
+- Target Major Version: 3.x
+- Reference Issues: (fill in existing related issues, if any)
+- Implementation PR: (leave this empty)
+
+# Summary
+
+Use `provide` to define component's public API available for parent components via `refs`.
+
+# Basic example
+
+Right now you can access the whole component's context without any limitations when using `refs`.
+
+```html
+<template>
+  <div />
+</template>
+
+<script>
+  export default {
+    name: 'Foo',
+    methods: {
+      someMethod() {}
+    }
+  }
+</script>
+```
+
+```html
+<template>
+  <Foo ref="foo" />
+</template>
+
+<script>
+  export default {
+    name: 'Bar',
+    mounted() {
+      this.$refs.foo.someMethod() // you can access anything on the context
+    }
+  }
+</script>
+```
+
+After the change you'll be required to expose your data and methods explicitly.
+
+```html
+<template>
+  <div />
+</template>
+
+<script>
+  export default {
+    name: 'Foo',
+    provide() {
+      const { someMethod } = this;
+      return {
+        someMethod,
+      }
+    },
+    methods: {
+      someMethod() {},
+      anotherMethod() {},
+    }
+  }
+</script>
+```
+
+```html
+<template>
+  <Foo ref="foo" />
+</template>
+
+<script>
+  export default {
+    name: 'Bar',
+    mounted() {
+      this.$refs.foo.someMethod() // exposed
+      this.$refs.foo.anotherMethod() // error, not exposed
+    }
+  }
+</script>
+```
+
+# Motivation
+
+Right now component's context is free to access by anyone and that brings up a number of issues:
+
+1. You can not guarantee consistent component's behaviour due to external modifications. You can modify components data and won't be able to tell where that change came from.
+2. There's no contract between receiver and provider components. This could lead to refactoring problems when there's an unused method within the component, but it's required by another component and there's no easy way to tell that.
+3. No clear separation of data that is required by the component itself and other components.
+
+To fix these issues components would be required to explicitly declare their public interface.
+
+# Detailed design
+
+Components should not be able to directly access other components context anymore. To declare component's public interface use `provide` and pass any data that should be exposed. The object returned in `provide` would serve as a main access point for the accessor component. To eliminate conflicts and provide better cohesion use `Symbol` for exposed data property names.
+
+A component with such an interface could look like this:
+
+```html
+<template>
+  <input ref="input">
+</template>
+
+<script>
+  export const FOCUS_INPUT = Symbol()
+  
+  export default {
+    name: 'MyInput',
+    provide() {
+      const { focusInput } = this
+      return {
+        [FOCUS_INPUT]: focusInput
+      }
+    },
+    methods: {
+      focusInput() {
+        this.$refs.input.focus()
+      }
+    }
+  }
+</script>
+```
+
+This interface could be utilized as following:
+
+```html
+<template>
+  <MyInput :ref="myInput" />
+</template>
+
+<script>
+  import { FOCUS_INPUT } from 'MyInput.vue'
+  
+  export default {
+    mounted() {
+      this.$refs.myInput[FOCUS_INPUT]()
+    }
+  }
+</script>
+```
+
+To expose refs use `mounted` hook and object as a wrapper to preserve reactivity.
+
+```html
+<template>
+  <input ref="input">
+</template>
+
+<script>
+  export const INPUT_EXPOSED = Symbol()
+  
+  export default {
+    provide() {
+      const { exposed } = this
+      return {
+        [INPUT_EXPOSED]: exposed
+      }
+    },
+    data() {
+      return {
+        exposed: {
+          inputRef: null
+        }
+      }
+    },
+    mounted() {
+      this.exposed.inputRef = this.$refs.input
+    }
+  }
+</script>
+```
+
+Or function refs:
+
+```html
+<template>
+  <input :ref="(input) => exposed.inputRef = input">
+</template>
+
+<script>
+  export const INPUT_EXPOSED = Symbol()
+  
+  export default {
+    provide() {
+      const { exposed } = this
+      return {
+        [INPUT_EXPOSED]: exposed
+      }
+    },
+    data() {
+      return {
+        exposed: {
+          inputRef: null
+        }
+      }
+    },
+  }
+</script>
+```
+
+Example using Composition API:
+
+```html
+<template>
+  <input ref="inputRef">
+</template>
+
+<script>
+  import { ref } from 'vue'
+  export const INPUT_REF = Symbol()
+  
+  export default {
+    name: 'MyInput',
+    setup() {
+      const inputRef = ref(null)
+      provide(INPUT_REF, inputRef)
+
+      return {
+        inputRef
+      }
+    }
+  }
+</script>
+```
+
+You'll then be able to use it as a ref:
+
+```html
+<template>
+  <MyInput ref="myInput" />
+</template>
+
+<script>
+  import { INPUT_REF } from 'MyInput.vue'
+  
+  export default {
+    setup() {
+      const myInput = ref(null)
+      onMounted(() => {
+        console.log(myInput.value[INPUT_REF].value)
+      })
+
+      return {
+        myInput
+      }
+    }
+  }
+</script>
+```
+
+# Drawbacks
+
+* Should be triggered after data init, `refs` then would require a lot of fiddling around to preserve reactivity (wrapping exposed values inside an object at a minimum)
+* Provides value down the render tree as a side-effect, which may not be the desired behaviour
+
+# Alternatives
+
+* Use dedicated hook instead of `provide` but with exactly the same API. Could be called `expose`.
+
+  **Pros**:
+
+  * Does not cause conflicts with other provisions
+  * Clear separation of concerns
+
+  **Cons**:
+
+  * One more option to learn about
+  * Could be confusing to pick one between the `provide` and `expose`
+  * Does not fix the issue with `refs`
+
+* Use events to expose your component's public interface:  
+  ```html
+  <template>
+    <input ref="input">
+  </template>
+
+  <script>    
+    export default {
+      mounted() {
+        const exposed = {
+          focusInput: () => {
+            this.$refs.input.focus()
+          }
+        };
+        this.$emit('ready', exposed);
+      }
+    }
+  </script>
+  ```
+
+  **Pros**:
+
+  * Does not require any API change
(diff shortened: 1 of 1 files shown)


leopiccionia:
Wait, Vue 2 & 3 already has a `provide` option (the complement of `inject`...), with very different semantics.

Anyway, is it good idea to incentive the coupling between parent and child? There are good reasons for [loose coupling](https://en.m.wikipedia.org/wiki/Loose_coupling) being considered an OOP good practice.

About the RFC, I'd like to see a comparison with the use of `interface` (which obviously requires TS), more specifically, child components implementing interfaces and parent components depending on interfaces, not concrete implementations. Is it currently possible? Is it ergonomic to implement and use?

CyberAP:
> Wait, Vue 2 & 3 already has a provide option (the complement of inject...), with very different semantics.
Yes, current proposal is essentially extending `provide` behaviour.

> Anyway, is it good idea to incentive the coupling between parent and child? There are good reasons for loose coupling being considered an OOP good practice.
The primary goal of this proposal is to reduce component coupling and make it as explicit as possible. When you have explicit public interface it can be refactored with ease and all your dependencies are tracked.

> About the RFC, I'd like to see a comparison with the use of interface (which obviously requires TS), more specifically, child components implementing interfaces and parent components depending on interfaces, not concrete implementations. Is it currently possible? Is it ergonomic to implement and use?
With class API not officially supported I doubt it's feasible.

leopiccionia:
> Yes, current proposal is essentially extending `provide` behaviour.

Does it make sense that, to expose some data to my descendants (current behavior), I'd have to also expose it to my parent (proposed extension), and vice-versa? It seems leaky.

Another con of both sharing the same name is that `provide` can't/shouldn't be erased in production mode. I can't see why this sort of protocol/interface mechanism would need to exist beyond development.

I'd imagine, too, that providing data has some (probably rather small) runtime penalty, even if it's not injected by any descendant. I'd like to be proved wrong on this.

CyberAP:
> Does it make sense that, to expose some data to my descendants (current behavior), I'd have to also expose it to my parent (proposed extension), and vice-versa? It seems leaky.

That's listed in the drawbacks section of the proposal. I'd be happy with an alternative `expose` option if using `provide` for that purpose doesn't fit well.

> Another con of both sharing the same name is that provide can't/shouldn't be erased in production mode. I can't see why this sort of protocol/interface mechanism would need to exist beyond development.

This interface should persist in the production mode. You should not be able to access component instance freely anymore.

> I'd imagine, too, that providing data has some (probably rather small) runtime penalty, even if it's not injected by any descendant. I'd like to be proved wrong on this.

That's true if we have any injections, since they're recursive they will cycle through all the provisions of every component in the parent tree starting from the injecting component to the root.
I'd probably should state that in the drawbacks section.

jacekkarczmarczyk:
Not sure if the proposed RFC is the best way to do that but I do like the idea of explicit exposing public component's methodss so I hope it's at least good starting point for discussion. 
I think the `expose` thing mentioned in the last comment may be a better idea than using `provide` (and actually I had this name in mind before I read that comment)

CyberAP:
Thanks, @aztalbot, I'm leaning towards `expose` more and more. I like the idea of simple API for quick exposure using array of strings, the same way `inject` works.

I am thinking that the `expose` API could have 3 versions each serving its own purpose:

* Array of strings — quickly share data with direct bindings
  ```
  expose: ['foo', 'bar']
  ```

* Object syntax — same as array, but with ability to rename exposed data
  ```
  expose: {
    foo: 'bar' // expose foo as bar
  }

* Function syntax — full control over exposed data
  ```
  expose() {
    const { foo } = this
    return { foo }
  }
  ```

smolinari:
Is the ability to dig into a child component's full context really an issue? If someone does it and does something improper or ends up even changing behavior, that's their problem in the end, isn't it? They are misusing the component. 

Put another way, the API contract of a Vue component is implicit and doesn't need to be explicit. For instance, from the craziness I've seen, like some noob using private (as in with underscore prefixed) methods is very, very rare. And even when doing so, almost always things just don't work right or as expected. So, yeah, they may run into the footgun, but they back out of that attempt quickly and they'd usually learn to just compose a new component with the behavior they wanted. 

I feel this is trying to fix a problem that doesn't exist, and the motivational points of the RFC don't really offer real world examples of the concerns either. Granted, this is coming only from my personal experience and perspective and even though I've been working to support and help people with a Vue framework for several years, I don't consider myself an expert.  

As the saying goes, You can't design a system to avoid stupid. Stupid will always win.

Ok. I just made that up. :grin: But, I think you get the point and I feel overcomplicating the component API for a very rare problem is adding an unnecessary cost.
 
Scott

smolinari:
Maybe coming up with some real world examples would be helpful? Explain where the implicit API wasn't enough and show how the suggested explicit API would be better. For sure that exercise will either strengthen your arguments and sell the suggestion better or maybe also change your mind on its overall value?

Scott

jacekkarczmarczyk:
Components can have a lot of methods. Some of them can be considered as public and others as internals. The main gain IMHO would be IDE autocompletion where you could see just the safe (public) ones. so for example [Vuetify VTextField component](https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/components/VTextField/VTextField.ts#L232), it would be nice if after typing `this.$refs.textField.` you would get listed only `focus` and `blur` methods in autocompletion instead of endless list of internals. And hackers should be still able to access other methods, but using something like `this.$refs.textField.$options.methods.genTextFieldSlot()`

CyberAP:
> @CyberAP, migration-wise, were you thinking a script could just find all usage $refs, and then add the necessary exports to that child component?

@aztalbot, I'd better go with a warning in compatibility build, or some way of a `loose` mode, opposed to default `strict` mode.

I have rewritten the RFC to use `exposed` option and now it's open for reviews.

active-rfcs/0000-component's-public-api.md
```diff
@@ -45,37 +45,32 @@ After the change you'll be required to expose your data and methods explicitly.
...
 
 <script>
   export default {
-    name: 'Bar',
+    name: 'Parent',
     mounted() {
-      this.$refs.foo.someMethod() // exposed
-      this.$refs.foo.anotherMethod() // error, not exposed
+      this.$refs.input.focus() // exposed
+      this.$refs.input.blue() // error, not exposed
```
smolinari: spelling of blur

CyberAP: Thanks

privatenumber:
### Feedback
Great idea! I think tighter encapsulation of components is a step in the right direction.

Currently, a parent component can access and manipulate everything about a child component. As a part of a framework's role to **guide engineers towards better practices**, I think it makes sense to prevent things like letting [a parent component directly manipulate a child component's state](https://jsfiddle.net/hirokiosame/9u73mkcf/) and [vice-versa](https://jsfiddle.net/hirokiosame/a57c1m68/).

1. Might be a good idea to update the first example to something that highlights a bigger problem. Currently, it expectedly calls a method on the child component, which doesn't highlight some of the more severe consequences of looser encapsulation (eg. anti-patterns to [one-way data flow](https://vuejs.org/v2/guide/components-props.html#One-Way-Data-Flow)).
2. If the purpose is tighter encapsulation, would properties like `.$el` have to be explicitly exposed too? I wonder what the implications this might have on the convenience of easily being able to access the DOM element to invoke API methods or retrieve data. And how far would we go to protect it when it can still be accessed via event objects?
3. I think the only thing that should be exposed are methods or read-only data for the component to better enforce one-way data-flow and ownership of data.

### Alternative considerations
#### `publicMethods` hash
Where only methods can be exposed and are defined in a separate hash
 ```html
<template>
  <input>
</template>

<script>
export default {
  name: 'MyInput',
  publicMethods: {
    el() { return this.$el },
    focus() { return this.$el.focus() },
    blur() { return this.$el.blur() },
  }
}
</script>
```

#### Public method prefix
 ```html
<template>
  <input>
</template>

<script>
export default {
  name: 'MyInput',
  methods: {
    // Public methods prefixed with $
    $focus() { return this.$el.focus() },
    $blur() { return this.$el.blur() },
  }
}
</script>
```

#### `protected` property
Similar to the `functional` property (I know Vue 3.0 removes it), but a component type would prevent any breaking changes (probably wouldn't be _completely_ encapsulated and will still have to expose core properties like `$el`, `$parent`, `$vnode`, `$options`, etc.).
 ```html
<template>
  <input>
</template>

<script>
export default {
  name: 'MyInput',
  protected: true, // Nothing is exposed unless explicitly defined below
  publicMethods: {
    el() { return this.$el },
    focus() { return this.$el.focus() },
    blur() { return this.$el.blur() },
  }
}
</script>
```

#### `api` tag
Similar to above, would not introduce any breaking changes (unless people are using the `api` tag as a custom block)
 ```html
<template>
  <input>
</template>

<script>
export default {
  name: 'MyInput',
}
</script>

<api>
export default {
  el() { return this.$el },
  focus() { return this.$el.focus() },
  blur() { return this.$el.blur() },
}
</api>
```

CyberAP:
@privatenumber thanks for your feedback! I have also considered the idea of sharing just methods to avoid directly manipulating other component's data. Probably the best way to deal with this would be printing an error in a console when you try to change component's state directly, the same way it works with props mutations. Exposing just methods is too limiting in my opinion, since we can have computeds that could potentially contain important information about the component and to access those you'll need to call a method, which essentially duplicates existing functionality. But yes I believe everything exposed on a component should be read-only. Need more feedback on that issue before I add it to the RFC.

Regarding `$el` and other implicit props — I believe they should behave exactly the same as the rest of the context and required explicit exposure.

active-rfcs/0000-component's-public-api.md
```diff
@@ -0,0 +1,408 @@
...
+    }
+  }
+</script>
+```
+
+# Drawbacks
+
+* `expose` should be executed after data init, `refs` then would require a lot of fiddling around to preserve reactivity (wrapping exposed values with an object at a minimum)
+* Could be difficult to implement
+* Not backwards compatible (but could only warn in compatibility build for example)
```
posva:
```suggestion
* Not backwards compatible (but could only warn in compatibility build for example)
* Increased bundle size
* Decreased performance: a mechanism like a proxy is necessary to prevent specific properties to be accessible
```

It's worth noting this as well

jacekkarczmarczyk:
Maybe it's enough there's an easy built-in way to properly type the component so that only public data/methods are available in the public interface?

CyberAP:
Thanks for yout input @posva.

> By default, specially when using ui libraries, no internal components properties should be used

This is unfortunately not true. You'll have to access component internals in order to focus an element. Doing querySelector or `$el` are way worse than doing `$refs.input.focus()`.

You'll also have to do this when using an uncontrolled component, like a self-closing dropdown. In order to forcefully close it you'll have to call `$refs.dropdown.close()` and there's no better way of doing it. You might use a watcher on a prop, but that's a reactivity hack, rather than a real solution to the problem.

And in case your component has these externally required methods there's no guarantee they will be present after the next refactor. If you do not specifically test for this then you'll even won't be able to tell that it's broken.

posva:
> This is unfortunately not true. You'll have to access component internals in order to focus an element. Doing querySelector or $el are way worse than doing $refs.input.focus().

It is. That's not the component internals (data, methods, etc). It's part of a Vue instance. I literally mentioned `$el` right after in the sentence you cropped

> You'll also have to do this when using an uncontrolled component, like a self-closing dropdown. In order to forcefully close it you'll have to call $refs.dropdown.close()

Yeah, and that is fine as long as it is documented. That's why I said _by default_, as in, if no documentation is provided.

> And in case your component has these externally required methods there's no guarantee they will be present after the next refactor. If you do not specifically test for this then you'll even won't be able to tell that it's broken.

They will be there if they are part of the public api and there are documented. That's pretty much the whole point

leopiccionia:
> Maybe it's enough there's an easy built-in way to properly type the component so that only public data/methods are available in the public interface?

You can type `this.$refs` using TS interfaces. It's not the same of typing a child component directly.

The good parts are that interfaces are reusable and can be re-implemented by various components, and are totally erased during build (and, therefore, has zero runtime cost).

The bad part is that I don't know how to disallow `$el` access, for example.

See demo: https://github.com/leopiccionia/typing-vue-with-interfaces

jacekkarczmarczyk:
> See demo: https://github.com/leopiccionia/typing-vue-with-interfaces

Thanks, but I'd rather see a demo without using vue-class-component

CyberAP:
> The bad part is that I don't know how to disallow $el access, for example.

And even worse you can't disallow any access to the instance. There's nothing stopping you from doing `vm.$refs.component.$emit('input', 'no-srp')`. Or directly manipulating data.
I'm fine with closing down instance in development and completely disabling that in production. At least it will be very hard to do bad things while you're developing and consider finding alternatives or exposing data explicitly.

leopiccionia:
> And even worse you can't disallow any access to the instance.

At this point, it's important to settle which is the **primary** goal of this RFC:

* Strict defensive programming (motivation 1);
* Avoiding reliance on undocumented API (motivations 2 and 3).

Those motivations are quite different and affect distinct users sets, so the earlier we settle this, the more productive will be the RFC discussion from then on.

If we're only interested in avoiding unintended breaking changes and easing code refactoring, type checking + well-written documentation is a good enough combo. Linting rules can also go a long way to educate users about relying on other components' internals.

If strict defensive programming is critical, we should ask if something like `expose` is a productive solution, as `$refs` are not the only way to access component's instance externally AFAIK.

CyberAP:
I wouldn't call that defensive programming but rather a [contract programming](https://en.wikipedia.org/wiki/Design_by_contract).

leopiccionia:
I created a [POC](https://github.com/leopiccionia/vue-public-api) for use in Composition API, in case anybody is interested in userland possibilities.

yyx990803:
Closing in favor of https://github.com/vuejs/rfcs/pull/343

PR #136, opened 2020-02-28
Change active and exact-active behavior for router-link

posva:
[Rendered](https://github.com/posva/rfcs/blob/router/active-matching/active-rfcs/0000-router-active-link.md)

Make active matching more consistent by avoiding existing caveats by design

- Is this breaking change bringing more value than the old behavior?
- Are any of the alternatives listed more desirable?
- Do you need the `exact` prop?

Please, vote with :+1: and :-1: if this is something you want or not


active-rfcs/0000-router-active-link.md
```diff
@@ -0,0 +1,119 @@
...
+      // empty child
+      { path: '' },
+      // child with id
+      { path: 'child/:id' }
+    ]
+  }
+]
+```
+
+If the current route is `/parent/1/child2`, these links will be active:
```
cawa-93: Maybe `/parent/1/child/2` ?

posva: fixed!

posva:
I added a few examples about the breaking change that should make this RFC clearer

posva:
I added a section about aliases to explain how they behave with the active behavior

posva:
Added a section about nested absolute aliases

posva:

PR #140, opened 2020-03-05
Prop modifier for v-model

CyberAP:
Add a `prop` modifier for `v-model` to simplify wrapper components.

[Rendered](https://github.com/CyberAP/rfcs/blob/props-model/active-rfcs/0000-props-model.md)

### DIFF
--- a/active-rfcs/0000-props-model.md
+++ b/active-rfcs/0000-props-model.md
@@ -1,0 +1,142 @@
+- Start Date: 2020-03-05
+- Target Major Version: 3.x
+- Reference Issues: (fill in existing related issues, if any)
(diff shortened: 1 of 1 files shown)


victornpb:
I really like this, wrapping something should be just as easy as "wrapping it"!

yooneskh:
I absolutely support this and this is the best among rfcs about this problem. This really reduces a lot of code when we wrap things.

`v-model` is already (presented as) a syntactic sugar for binding and emitting, so a modifier for `v-model` would modify the syntactic sugar a bit.

CyberAP:
Closing in favour of https://vuejs.org/guide/components/v-model and `defineModel` composable.

PR #143, opened 2020-03-07
Components registration with props

nsznsznjsz:
Provide default props on component registration when define component.
```javascript
const Panel = {
  components: {
    Card: [Card, { shadow: 'hover' }]
  },
  template: `<Card />` // Is equivalent to: <Card shadow="hover">
}
```

[Rendered](https://github.com/nsznsznjsz/rfcs/blob/components-registration-with-props/active-rfcs/0000-components-registration-with-props.md)

### DIFF
--- a/active-rfcs/0000-components-registration-with-props.md
+++ b/active-rfcs/0000-components-registration-with-props.md
@@ -1,0 +1,83 @@
+- Start Date: 2020-3-7
+- Target Major Version: 3.x
+- Reference Issues: (fill in existing related issues, if any)
+- Implementation PR: (leave this empty)
+
+# Summary
+
+Components registration with default props when define component, default props will auto injected. 
+
+# Basic example
+
+Provide default props on component registration.
+```javascript
+const Panel = {
+  components: {
+    Card: [Card, { shadow: 'hover' }]
+  },
+  template: `<Card />` // Is equivalent to: <Card shadow="hover">
+}
+```
+
+# Motivation
+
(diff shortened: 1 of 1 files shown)


posva:
The motivation part is, to me, way too short given the small improvement this brings.

Even though, this can just be a helper function:

```js
function fixedProps(component, propsData) {
  // this example is missing how things should be properly passed, but it doesn't change much
  return () => h(component, {...propData}, children)
}
```

and then use that:

```js
components: {
  Card: fixedPropd(Card, { shadow: 'hover' })
}
```

So I don't think there is a need to add another api for components

jacekkarczmarczyk:
@posva how'd you do that with components that are already registered (i.e. coming from some lib) and you want to set default props globally and not just within one `components` object?

posva:
@jacekkarczmarczyk with `resolveComponent` to retrieve the component, which could be done by the function itself if it gets a string (I updated the example)

jacekkarczmarczyk:
Maybe I wasn't clear enough, I meant that I don't want to use `components` option for components that are registered globally. Imagine for example I want `exaxt` prop of a `router-link` to be always true, without need of adding it to `components` option as it is done in your example

CyberAP:
I'd better go with a component wrapping for this one. It's already an established practice and works perfectly fine. Could you expand on why component wrapping doesn't fit you and how this proposal fixes the issue?

nsznsznjsz:
Seems implement this feature as a lib is the better way.

PR #144, opened 2020-03-12
Merge `meta` fields from parent to child

posva:
[Rendered](https://github.com/posva/rfcs/blob/router/meta-merging/active-rfcs/0000-router-merge-meta.md)


KaelWD:
> I believe this is what Nuxt does

[vue-meta](https://vue-meta.nuxtjs.org/faq/#how-is-metainfo-resolved)?

kiaking:
I like this one. It's going to make meta accessing much easier 🙌 

One question. Why is `nested` key removed in "Basic example" section's example, and not in "Detailed design" section's example?

```js
{
  path: '/parent',
  meta: { nested: { requiresAuth: true, isChild: false } },
  children: [
    { path: 'child', meta: { isChild: true }}
  ]
}

// becomes
{ requiresAuth: true, isChild: true }
```

And 

```js
{
  path: '/parent',
  meta: { nested: { a: true } },
  children: [
    { path: 'child', meta: { nested: { b: true }}}
  ]
}

// becomes
{
  nested: {
    b: true
  }
}
```

Wouldn't the first example become as below?

```js
{
  nested: {
    requiresAuth: true,
    isChild: true
  },
  isChild: false
}
```

Or, is `nested` key means something special?

posva:
Thanks @kiaking That was indeed a mistake. I fixed it

posva:

Rolanddoda:
The rendered link gives 404

ludohenin:
I may jump into that one a bit late but in my case merging meta doesn't sound that great.
I mainly use the meta to hold the document title and use it to build it as a tree.

```js
'/'                // title = App name
  'settings'       // title = Setting - App name
    'users'        // title = Users - Settings - App name
```

I guess merging the `meta.title`

posva:
@ludohenin Since you need all values from `meta` to create your tree, you probably go through the array of `matched` like [it's shown on documentation](https://router.vuejs.org/guide/advanced/meta.html#route-meta-fields). That array is kept the same, only the `to.meta` and `from.meta` properties are the results of merging parent to child `meta` properties, the ones at `to.matched` and `from.matched` are the originals

ludohenin:
@posva ah OK, perfect then ;-)

PR #149, opened 2020-03-25
Custom Elements Interop Improvements

yyx990803:
- **Breaking:** Custom Elements whitelisting is now performed during template compilation, and should be configured via compiler options instead of runtime config.
  - This removes the runtime overhead for custom elements support.
- **Breaking:** Restrict the special `is` prop usage to the reserved `<component>` tag only.
  - This improves support for [customized built-in elements](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements-customized-builtin-example).
- **New**: Introduce a new `v-is` directive to support 2.x use cases where is was used on native elements to work around native HTML parsing restrictions.

[Full Rendered Proposal](https://github.com/vuejs/rfcs/blob/custom-elements-interop/active-rfcs/0027-custom-elements-interop.md)


backbone87:
just some thoughts:
- builtin/custom element: `<my-element>`
- dynamic builtin/custom element: `<[myElementName]>`
- component: `<!my-component>` - registered globally/locally
- dynamic component: `<![MyComponent]>` - returned by setup/instance prop

(the `!` meta char is up to debate)

pros:
- no more `is`
- clear distinction between components and elements directly inside the template
- removes the need to declare locally registered components when using setup functions
- consistent and aligned syntax for dynamic artifacts (square brackets)

yyx990803:
@backbone87 we are not considering any drastic template syntax change at this point.

jods4:
The idea of `<[comp]>` as an additional syntax meant to replace `<component :is=comp>` feels very natural, though, considering the existing dynamic syntaxes. 
Could be an idea for 3.1! 💭

yyx990803:
`<[comp]>` is not valid HTML (it would be interpreted as plain text by native parsers) so that would not work for in-DOM templates. I'm afraid it's not an option.

jods4:
so are self-closing tags? They are really handy, I love them.

yyx990803:
@jods4 yeah, that's an exception for string-only templates (so are CamelCase components). We already have these special cases and it would be wise to avoid adding more, especially when you are not adding new capabilities.

backbone87:
I provided an expanded form, which the browser should be able to parse

leopiccionia:
Would the following pattern be removed ([guide](https://vuejs.org/v2/guide/components.html#DOM-Template-Parsing-Caveats
))?
```html
<table>
  <!-- CustomRow resolves to a function or object, not a string -->
  <tr :is="CustomRow">
    <td>1</td>
  </tr>
</table>
```

As of today, my evergreen browser still parses:

```html
<table>
  <custom-row>
    <td>1</td>
  </custom-row>
</table>
```

Like the following:

```html
<custom-row></custom-row>
<table>
  <tr>
    <td>1</td>
  </tr>
</table>
```

yyx990803:
@leopiccionia that's a good point. I guess we should still this case when `is` binding value is not a string.

active-rfcs/0000-custom-elements-interop.md
```diff
@@ -28,8 +31,8 @@ The default generated render function code is (pseudo code):
 
 ```js
 function render() {
-  const component_plastic_button = resolveComponent('plastic-button')
-  return createVNode(component_plastic_button)
+  const component_plastic_button = resolveComponent("plastic-button");
```
Justineo: Code style seems inconsistent with the later part of the RFC, maybe something wrong with the formatter?

posva: We could add a `.prettierrc` file to make it automatic for people using a plugin with their IDE

yyx990803: done

yyx990803:

PR #150, opened 2020-03-28
Redesigning Navigation failures and global handlers

posva:
[Rendered](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0033-router-navigation-failures.md)

### DIFF
--- a/active-rfcs/0000-router-navigation-failures.md
+++ b/active-rfcs/0000-router-navigation-failures.md
@@ -1,0 +1,173 @@
+- Start Date: 2020-03-26
+- Target Major Version: Vue Router v4
+- Reference Issues: https://github.com/vuejs/vue-router/issues/2833, https://github.com/vuejs/vue-router/pull/3047, https://github.com/vuejs/vue-router/issues/2932, https://github.com/vuejs/vue-router/issues/2881
+- Implementation PR:
+
+# Summary
+
+- Explicitely define what a navigation failure is and how and where we can catch them.
+- Change when the Promised-based `router.push`(and `router.replace` by extension) resolves and rejects.
+- Make `router.push` consistent with `router.afterEach` and `router.onError`.
+
+# Basic example
+
+## `router.push`
+
+If there is an unhandled error or an error is passed to `next`:
+
+```js
+// any other navigation guard
+router.beforeEach((to, from, next) => {
+  next(new Error())
+  // or
+  throw new Error()
+  // or
+  return Promise.reject(new Error())
+})
+```
+
+Then the promise returned by `router.push` rejects:
+
+```js
+router.push('/url').catch(err => {
+  // ...
+})
+```
+
+In **all** other cases, the promise is resolved. We can know if the navigation failed or not by checking the resolved value:
+
+```js
+promise.push('/dashboard').then(failure => {
+  if (failure) {
+    failure instanceof Error // true
+    failure.type // NavigationFailure.canceled
+  }
+})
+```
+
+## `router.afterEach`
+
+It's the global equivalent of `router.push().then()`
+
+## `router.onError`
+
+It's the global equivalent of `router.push().catch()`
+
+# Motivation
+
+The current behavior of Vue Router regarding the promise returned by `push` is inconsistent with `router.afterEach` and `router.onError`. Ideally, we should be able to catch all succeeded and failed navigations globally and locally but we can only do it locally.
+
+- `onError` is only triggered on thrown errors and `next(new Error())`
+- `afterEach` is only called iif there is a navigation
+- `redirect` should behave the same as `next('/url')` in a Navigation guard
+
+The differences between the Promise resolution/rejection vs `router.afterEach` and `router.onError` are inconsistent and confusing.
+
+# Detailed design
+
+One of the main points is to be able to consistently handle failed navigations globally and locally:
+
+- Failed Navigation:
+  - Triggers `router.afterEach`
+  - Resolves the `Promise` returned by `router.push`
+- Uncaught Errors, `next(new Error())`
+  - Triggers `router.onError`
+  - Rejects the `Promise` returned by `router.push`
+
+It's important to note there is no overlap in these two groups: if there is an uhandled error in a Navigation Guard, it will trigger `router.onError` as well as rejecting the `Promise` returned by `router.push` but **will not trigger** `router.afterEach`. Cancelling a navigation with `next(false)` will not trigger `router.onError` but will trigger `router.afterEach`
+
+## Changes to the Promise resolution and rejection
+
+Navigation methods like `push` return a Promise, this promise **resolves** once the navigation succeeds **or** fail. It **rejects** only if there was an unhandled error. If it rejects, it will also trigger `router.onError`.
+
+To differentiate a Navigation that succeeded from one that failed, the `Promise` returned by `push` will resolve to either `undefined` or a `NavigationFailure`:
+
+```js
+import { NavigationFailureType } from 'vue-router'
+
+router.push('/').then(failure => {
+  if (failure) {
+    // Having an Error instance allows us to have a Stacktrace and trace back
+    // where the navigation was cancelled. This will, in many cases, lead to a
+    // Navigation Guard and the corresponding `next()` call that cancelled the
+    // navigation
+    failure instanceof Error // true
+    if (failure.type === NavigationFailureType.canceled) {
+      // ...
+    }
+  }
+})
+
+// using async/await
+let failure = await router.push('/')
+if (failure) {
+  // ...
+}
+```
+
+By not rejecting the `Promise` when the navigation fails, we are avoiding _Uncaught (in promise)_ errors while still keeping the possibility to check if the Navigation failed or not.
+
+### Navigation Failures
+
+There are a few different navigation failures, to be able to react differently in your code
+
+Navigation failures can be differentiated through a `type` property. All possible values are hold by an `enum`, `NavigationFailureType`:
+
+- `cancelled`: `next(false)` inside of a navigation guard or a newer navigation took place while another one was ongoing.
+- `duplicated`: Navigating to the same location as the current one will cancel the navigation and not invoke any Navigation guard.
+
+On top of the `type` property, navigation failures also expose `from` and `to` properties, exactly like `router.afterEach`
+
+### Redirections
+
+Redirecting inside of a navigation guard with `next('/url')` is not a navigation failure by itself, as the navigation still takes place and ends up somewhere. To detect a navigation, specially during SSR, there is a `redirectedFrom` property accessible on the `currentRoute`.
+
+E.g.: imagine a navigation guard that redirects to `/login` when the user isn't authenticated:
+
+```js
+router.beforeEach((to, from, next) => {
+  // redirect to the login page if the target location requires authentication
+  if (to.meta.requiresAuth && !isAuthenticated) next('/login')
+  else next()
+})
+```
+
+When navigating to a location that requires authentication, we can retrieve the original location the user was trying to access via `redirectedFrom`:
+
+```js
+// user is not authenticated
+await router.push('/profile/dashboard')
+
+// `redirectedFrom` is a RouteLocationNormalized, like `currentRoute` but we are omitting
+// most properties to make the example readable
+router.currentRoute // { path: '/login', redirectedFrom: { path: '/profile/dashboard' } }
+```
+
+## Changes to `router.afterEach`
+
+Since `router.afterEach` also triggers when a navigation fails, we need a way to know if the navigation succeeded or failed. To do that, we introduce an extra parameter that contains the same _failure_ we could find in a resolved navigation:
+
+```js
+import { NavigationFailureType } from 'vue-router'
+
+router.afterEach((to, from, failure) => {
+  if (failure) {
+    if (failure.type === NavigationFailureType.canceled) {
+      // ...
+    }
+  }
+})
+```
+
+# Drawbacks
+
+- Breaking change although migration is relatively simple and in many cases will allow the developer to remove existing code
+
+# Alternatives
+
+- Differentiating `next(false)` from navigations that get overriden by more recent navigations by defining another Navigation Failure
+
+# Adoption strategy
+
+- `afterEach` and `onError` are relatively simple to migrate, most of the time they are not used many times either.
+- `router.push
--- a/active-rfcs/0000-router-navigation-failures.md
+++ b/active-rfcs/0000-router-navigation-failures.md
@@ -37,7 +37,7 @@ router.push('/url').catch(err => {
 In **all** other cases, the promise is resolved. We can know if the navigation failed or not by checking the resolved value:
 
 ```js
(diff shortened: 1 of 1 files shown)


active-rfcs/0000-router-navigation-failures.md
```diff
@@ -0,0 +1,173 @@
...
+```js
+router.push('/url').catch(err => {
+  // ...
+})
+```
+
+In **all** other cases, the promise is resolved. We can know if the navigation failed or not by checking the resolved value:
+
+```js
+promise.push('/dashboard').then(failure => {
```
nicolas-t:
```suggestion
router.push('/dashboard').then(failure => {
```

active-rfcs/0000-router-navigation-failures.md
```diff
@@ -0,0 +1,173 @@
...
+## `router.onError`
+
+It's the global equivalent of `router.push().catch()`
+
+# Motivation
+
+The current behavior of Vue Router regarding the promise returned by `push` is inconsistent with `router.afterEach` and `router.onError`. Ideally, we should be able to catch all succeeded and failed navigations globally and locally but we can only do it locally.
+
+- `onError` is only triggered on thrown errors and `next(new Error())`
+- `afterEach` is only called iif there is a navigation
```
trickstival:
```suggestion
- `afterEach` is only called if there is a navigation
```

trickstival (changes requested):

active-rfcs/0000-router-navigation-failures.md
```diff
@@ -0,0 +1,173 @@
...
+- `redirect` should behave the same as `next('/url')` in a Navigation guard
```
bponomarenko: Should it? From current [documentation](https://router.vuejs.org/guide/essentials/redirect-and-alias.html#redirect):
> Note that Navigation Guards are not applied on the route that redirects, only on its target. In the example below, adding a beforeEnter guard to the /a route would not have any effect.

As I understand, `afterEach` (as well as `beforeEach` and route's `beforeEnter') will not be called for redirect routes. Will this change?

posva: You are right, it's not exactly the same and it should be pointed out

ducmerida:
@posva 
Maybe it would better to use a NavigationResult instead of NavigationFailure, with this you can handle all the navigation result cases with the same logic:
- Cancelled
- Duplicated
- Redirected
- Succeeded (or another word like Completed/Performed/Done to make a clear difference with Redirected that is also considered as a success state)

I say it because I think that is not very natural to check if the result of a function is undefined to check if its execution success, I think that is better to handle all with the same logic and always return a result.

posva:
Using a _NavigationResult_ is indeed another possibility. That pattern comes from Node with error
first callbacks. I think it being natural or not varies on backgrounds but it's definitely very present in JS because of Node.

The reason I prefer using this approach is because there is no distinction in the succeeding cases: when navigation succeeds, a new `currentRoute` is set. For Redirections, there is `redirectedFrom` that appears to be able to treat it

StefanFeederle:
Does this mean pushing the current route again will resolve the promise with a NavigationFailureType 'duplicated' instead of rejecting?

The old behaviour encouraged adding `.catch((err) => {})` to all `router.push()` to get rid of `duplicated` rejections. This seems like an antipattern. With the proposed changes, all empty errorHandling can be deleted if no further logic in `route.push().then()` is needed. This is awesome!

I prefer the approach as suggested by @ducmerida to always return a NavigationResult.
This is easy to understand for beginners and handling the different outcomes can be done with the same logic. Is this possible while still providing error stacks on failed attempts?

posva:
> Does this mean pushing the current route again will resolve the promise with a NavigationFailureType 'duplicated' instead of rejecting?

Yes.

> I prefer the approach as suggested by @ducmerida to always return a NavigationResult. This is easy to understand for beginners and handling the different outcomes can be done with the same logic.

That's not really true. A NavigationResult would be redundant as it would contain the previous location and current location:

```js
const from = {...useRoute()} // create a copy of the current route
await router.push('/...')
const currentLocation = useRoute()
```

You still need extra logic to differentiate a failure from a success whereas with only failures, you can simply do:

```js
const failed = await router.push('/...')

if (failed) {
  // handle failure
} else {
  // handle success
}
```

compared to 

```js
const result = await router.push('/...')

if (result instanceof Error) {
  // handle failure
} else {
  // handle success
}
```

It's also important to note that the most common use case is waiting for the navigation to be finished without failure, so making that path easy to check is important. Failures are usually useful for debugging and tracing back what blocked the navigation, that's why them being Error instances is so useful

> Is this possible while still providing error stacks on failed attempts?

That depends on how a _NavigationResult_ would be implemented

trainiac:
A problem I've found is there is no way to achieve certain functionality with vue router when writing a universal web app without throwing an error. However throwing errors now causes unhandled rejections for push and replace.

Here are a couple use cases:
-  **Sending users to a not found page when visiting /users/:userId for a userId that doesn't exist**.  To handle this, in the beforeRouteEnter hook I throw an error.  The server side routing handles the error by responding with response code 404 and the not found page html. The client side router will listen to the router.onError and if the error is a "not found" type the page will reload the url which allows server side to show the not found page.
-  **Redirecting users as a permanent redirect vs temporary redirect**.  To handle this, in the beforeRouteEnter hook I throw an error that contains the information required for the server side to respond appropriately.

My question about all of these is how do I achieve the same thing without throwing an error?
It feels like there should to be a way to communicate more information to your global handlers for  what you want to happen without having to throw an error.

posva:
@trainiac that seems to be a case of redirection and you can handle it by checking the property `redirectedFrom`. You don't need to throw an error. As for how to achieve it, there is an example at https://github.com/posva/rfcs/blob/router/navigation-failures/active-rfcs/0000-router-navigation-failures.md#redirections

trainiac:
@posva Thanks for responding.  Forgive me if I'm missing something but I don't see how this answers my question. How can one tell from `redirectedFrom` whether the desired redirect should issue a 301 or a 302?  This seems like an important piece of a server side navigation result.

posva:
Deciding of its a 301 or 302 is up to you. From the router perspective it’s just a redirection

posva:

active-rfcs/0000-router-navigation-failures.md
```diff
@@ -0,0 +1,173 @@
+- Start Date: 2020-03-26
+- Target Major Version: Vue Router v4
+- Reference Issues: https://github.com/vuejs/vue-router/issues/2833, https://github.com/vuejs/vue-router/pull/3047, https://github.com/vuejs/vue-router/issues/2932, https://github.com/vuejs/vue-router/issues/2881
+- Implementation PR:
+
+# Summary
+
+- Explicitely define what a navigation failure is and how and where we can catch them.
```
snoozbuster, line 8: typo: `Explicitly`

active-rfcs/0000-router-navigation-failures.md
```diff
@@ -0,0 +1,173 @@
...
+
+It's the global equivalent of `router.push().catch()`
+
+# Motivation
+
+The current behavior of Vue Router regarding the promise returned by `push` is inconsistent with `router.afterEach` and `router.onError`. Ideally, we should be able to catch all succeeded and failed navigations globally and locally but we can only do it locally.
+
+- `onError` is only triggered on thrown errors and `next(new Error())`
+- `afterEach` is only called if there is a navigation
+- `redirect` should behave the same as `next('/url')` in a Navigation guard when it comes to the outcome of `router.push` and calls of `router.afterEach`/`router.onError`. The only differenc being that a `redirect` would only trigger leave guards and other before guards for the redirected location but not the original one
```
snoozbuster, line 62: typo: difference

active-rfcs/0000-router-navigation-failures.md
```diff
@@ -0,0 +1,173 @@
...
+One of the main points is to be able to consistently handle failed navigations globally and locally:
+
+- Failed Navigation:
+  - Triggers `router.afterEach`
+  - Resolves the `Promise` returned by `router.push`
+- Uncaught Errors, `next(new Error())`
+  - Triggers `router.onError`
+  - Rejects the `Promise` returned by `router.push`
+
+It's important to note there is no overlap in these two groups: if there is an uhandled error in a Navigation Guard, it will trigger `router.onError` as well as rejecting the `Promise` returned by `router.push` but **will not trigger** `router.afterEach`. Cancelling a navigation with `next(false)` will not trigger `router.onError` but will trigger `router.afterEach`
```
snoozbuster, line 77: typo: unhandled

active-rfcs/0000-router-navigation-failures.md
```diff
@@ -0,0 +1,173 @@
...
+- Breaking change although migration is relatively simple and in many cases will allow the developer to remove existing code
+
+# Alternatives
+
+- Differentiating `next(false)` from navigations that get overriden by more recent navigations by defining another Navigation Failure
+
+# Adoption strategy
+
+- `afterEach` and `onError` are relatively simple to migrate, most of the time they are not used many times either.
+- `router.push
```
snoozbuster, line 173: the rest of this line seems missing. is there some special consideration for migrating `router.push`?

snoozbuster:
Proposal seems good to me (have run into both NavigationDuplicated and "Promise rejected with undefined" errors recently after updating vue-router). I see that this is targeted for v4 - is there a chance this can instead be released with a minor version update of v3 (as such an update introduced the issues with promises in the first place)? would like to adopt this in my company's app ASAP and would prefer not to worry about larger breaking changes in v3 -> v4.

active-rfcs/0000-router-navigation-failures.md
snoozbuster, line 8:
```suggestion
- Explicitly define what a navigation failure is and how and where we can catch them.
```

active-rfcs/0000-router-navigation-failures.md
snoozbuster, line 62:
```suggestion
- `redirect` should behave the same as `next('/url')` in a Navigation guard when it comes to the outcome of `router.push` and calls of `router.afterEach`/`router.onError`. The only difference being that a `redirect` would only trigger leave guards and other before guards for the redirected location but not the original one
```

active-rfcs/0000-router-navigation-failures.md
posva, line 173:
```suggestion
- `router.push` doesn't reject when navigation fails anymore. Any code relying on catching an error should await the promise result instead
```

posva: Thanks for noticing!

posva:
@snoozbuster It's great you are supporting this change but because it is a breaking change, it can only be released as a major bump

snoozbuster:
@posva where was that mentality when promisifying router.push and router.replace was pushed as a minor bump? that was also a breaking change - I find it hard to believe that the solution for a breaking change introduced to consumers of this library by a minor bump is to absorb even more breaking changes caused by a major bump. What is the path to resolution for this issue for consumers which cannot upgrade to v4?

posva:
@snoozbuster It wasn't a breaking change as described at https://github.com/vuejs/vue-router/issues/2881#issuecomment-520554378. There are temporary solutions to prevent Promise rejections for navigation failures and the `NavigationFailureType` can be exposed in vue-router as well.
Introducing a different behavior would have been inconsistent with `onError` and `afterEach` and the existing callbacks. This had to be done with a breaking change to prevent further confusion.
The comments and reports on that regard helped me to create this RFC, which solves the biggest problem of uncaught promise rejections for navigation failures and makes global hooks consistent with `push`.

snoozbuster:
@posva ahh, I had seen that issue but only got as far as the link to this RFC, not all the way up to that comment, so I was expecting this RFC to resolve the issue. 

I don't strictly agree with the statement that "it wasn't a breaking change" because I was neither returning nor awaiting `router.replace()` before the Promise change, but I wasn't getting errors logged until the API changed from under me which caused our global unhandled Promise rejection handler to start logging (a huge volume of) errors. But, I also don't really agree with the idea that a duplicated navigation is a failure, either... in my eyes that's a no-op resulting in success, so I was surprised to see errors where I wasn't expecting any. 

All of our duplicate navigation issues are tied to one terribly janky piece of code anyway (and seeing all these errors made me realize it's even more janky than I thought it was), so I will handle them there for now and wait in anticipation for this new API change. Thanks for explaining - feel free to mark these last couple comments as off-topic.

active-rfcs/0000-router-navigation-failures.md
```diff
@@ -165,9 +165,10 @@ router.afterEach((to, from, failure) => {
 
 # Alternatives
 
-- Differentiating `next(false)` from navigations that get overriden by more recent navigations by defining another Navigation Failure
+- Differentiating `next(false)` from navigations that get overridden by more recent navigations by defining another Navigation Failure
 
 # Adoption strategy
 
+- Expose `NavigationFailureType` in vue-router@3 so that Navigation Failures can be told apart from regular Errors. We could also expose a function `isNavigationFailure` to tell them apart.
```
snoozbuster, line 172: ahh, this will help too. is there a better way to check for navigation failures right now than `e && e.name === 'NavigationDuplicated'`?

posva: There is no documented way to check for duplicated navigations right now (it's private api)

champkeh:
@posva 
Hi posva, after read the rfc and router's doc, I have some confuse
in rfc, example is this:
```js
import { NavigationFailureType, isNavigationFailure } from 'vue-router'

router.push('/').then((failure) => {
  if (failure) {
    // Having an Error instance allows us to have a Stacktrace and trace back
    // where the navigation was cancelled. This will, in many cases, lead to a
    // Navigation Guard and the corresponding `next()` call that cancelled the
    // navigation
    failure instanceof Error // true
    if (isNavigationFailure(failure, NavigationFailureType.canceled)) {
      // ...
    }
  }
})
```
(link: https://github.com/vuejs/rfcs/blob/master/active-rfcs/0033-router-navigation-failures.md#changes-to-the-promise-resolution-and-rejection)


But, in router's doc, example is this:
```js
const { isNavigationFailure, NavigationFailureType } = VueRouter

// trying to access the admin page
router.push('/admin').catch(failure => {
  if (isNavigationFailure(failure, NavigationFailureType.redirected)) {
    // show a small notification to the user
    showToast('Login in order to access the admin panel')
  }
})
```
(link: https://router.vuejs.org/guide/advanced/navigation-failures.html#detecting-navigation-failures)


I confuse that one is **then** and another is **catch** which capture the failure, so which is right place to get failure?

posva:
The catch is for Vue Router 3. I fixed the typo on vue router 4 docs

mariusa:
In Vue 3, how can one navigate to the same route, but with different query parameters? 
None of these answers work: https://stackoverflow.com/questions/59428227/how-to-use-router-push-for-same-path-different-query-parameter-in-vue-js

Thanks

iamandrewluca:
@mariusa I think the navigation is done. Your component data/derived data, should be reactive maybe 🤔 it should change when the query changes 👀

PR #152, opened 2020-04-01
Amendments to RouterLink scoped-slot

posva:
- Clearer statement of `tag` prop removal in v4
- Add `custom` prop to `router-link`

[Rendered](https://github.com/posva/rfcs/blob/router/amend-scoped-slot/active-rfcs/0021-router-link-scoped-slot.md)

### DIFF
--- a/active-rfcs/0021-router-link-scoped-slot.md
+++ b/active-rfcs/0021-router-link-scoped-slot.md
@@ -9,6 +9,7 @@
 - Remove `event` prop
 - Stop automatically assigning click events to inner anchors
 - Add a scoped-slot API
+- Add a `custom` prop to fully customize `router-link`'s rendering
 
 # Basic example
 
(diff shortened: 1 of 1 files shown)


backbone87:
Instead of the custom prop, we could also provide a separate component for full customization. https://martinfowler.com/bliki/FlagArgument.html

posva:
That's true but the size cost of an extra component plus the learning cost is way too bug compared to new prop or a new slot name

backbone87:
I think the learning curve is the same, maybe even clearer with separate components. Definitely better to read templates with separate components. The size aspect is correct, but the components should also be able to share most logic

mika76:
I think I prefer the idea of another component than a custom prop too. <router-link-slot> would be a lot easier to use or something than trying to see why my link's not working properly and remember custom in certain cases. It seems more clear to me an therefore would probably be easier to learn than to remember a custom prop.

posva:
> `<router-link-slot>` would be a lot easier to use or something than trying to see why my link's not working properly and remember custom in certain cases

I think a prop that changes the behavior of an existent component is easier to grasp than a whole new component where you wonder if they behave the same and why is there two components. It would be useful to see those cases because this actually fixes [the problem you were facing](https://github.com/vuejs/vue-router-next/issues/148)

A few points:

- RouterLink always renders its slots
- This amendment is not a breaking change, the only breaking changes are the removal of `event` and `tag` props and the implicit behavior that looked for an `a` element to attach click listeners

posva:

PR #153, opened 2020-04-01
Allow displaying a View different from where the user currently is at

posva:
[Rendered](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0036-router-view-route-prop.md)

Enables Modals that change the URL


nainardev:
so, finally we can have twitter or ig like modals? thanks :blush:

posva:
There is an example here: https://github.com/vuejs/vue-router-next/tree/master/e2e/modal

posva:

serkodev:
> There is an example here: https://github.com/vuejs/vue-router-next/tree/master/e2e/modal

In case someone need the updated location of the `e2e/modal` source code, please visit:
https://github.com/vuejs/router/blob/main/packages/router/e2e/modal/index.ts

PR #160, opened 2020-04-21
Support for KeepAlive and Transition in Vue Router 4

posva:
[Rendered](https://github.com/posva/rfcs/blob/router/router-view-slot/active-rfcs/0000-router-view-keep-alive-transitions.md)


dobromir-hristov:
Looks good. I think that if you want to use a `transition` or `keep-alive` with vue router, you are already going into more advanced territory. 😆  This solution makes pretty good sense, and having it well explained in the docs is good enough. 

If it was possible to detect if `router-view` is a direct child of a `transition`, and warn users that it wont work as they expect, maybe even point to docs, that would be even better, but yeah...

active-rfcs/0000-router-view-keep-alive-transitions.md
```diff
@@ -0,0 +1,101 @@
...
+
+We can achieve this level of control through a slot:
+
+```vue
+<router-view v-slot="{ Component, attrs }">
+  <component :is="Component" v-bind="attrs"></component>
+</router-view>
+```
+
+In the example, `Component` is the _component definition_ that can be passed to the function `h` or to the `is` prop of `component`. `attrs` is a set of event listeners and props that `RouterView` **requires** in order to work properly. Namely, `onVnodeMounted`/`onVnodeUnmounted` and a `ref`. It also contains any _attribute_/_prop_ passed to `RouterView`. Let's say we pass a `class` to it, we will find it in `attrs`:
```
yyx990803, line 32: I don't think we need to pass required internal `attrs` this way - `router-view` can simply get the returned component vnode from the slot and add those `attrs` internally:

``` js
const componentVNode = slots.default
  ? slots.default({ ... })
  : /* non slot handling */
return cloneVNode(componentVNode, attrs)
```

active-rfcs/0000-router-view-keep-alive-transitions.md
```diff
@@ -0,0 +1,101 @@
...
+
+```vue
+<router-view class="view" v-slot="{ attrs }">
```
yyx990803, line 35: I don't see the point of this. It makes sense when using the non-slot syntax, but with a slot the user should simply add the class directly in the slot.

active-rfcs/0000-router-view-keep-alive-transitions.md
```diff
@@ -0,0 +1,101 @@
...
+</router-view>
+```
+
+The same happens with any event listeners directly attached to `RouterView`, they appear on `attrs` **so they can be easily passed down with one singe `v-bind`**.
+
+They also contain any _prop_ passed by using the `props` option in a route record:
+
+```js
+createRouter({
+  routes: [{ path: '/users/:id', component: User, props: true }],
```
yyx990803, line 46: With `attrs` out of the way, I think we should re-consider what is passed there. I was thinking maybe it should pass the route params:

```vue
<router-view v-slot="{ Component, params }">
  <component :is="Component" v-bind="params"></component>
</router-view>
```

But then why not `query`? Why not other information of the current route?

Then I realized you already have access to the entire route object anyway:

```vue
<router-view v-slot="{ Component }">
  <component :is="Component"
    v-bind="$route.params"
    :dark-mode="$route.query.dark">
  </component>
</router-view>
```

But maybe we can still pass in the route object to make it rely less on magical global `$` properties:

```vue
<router-view v-slot="{ Component, route }">
  <component :is="Component"
    v-bind="route.params"
    :dark-mode="route.query.dark">
  </component>
</router-view>
```

This also makes it nicer to use in render functions:

``` js
import { RouterView } from 'vue-router'

return h(RouterView, ({ Component, route }) => h(Component, {
  ...route.params,
  darkMode: route.query.dark
}))
```

active-rfcs/0000-router-view-keep-alive-transitions.md
posva: That's nice, I didn't know about that! In this case, since `slot.default()` returns an array of Vnodes, I think it makes sense to require the user to only use one root element in the slot since we are adding attributes to the vnode

active-rfcs/0000-router-view-keep-alive-transitions.md
posva: It was more of a convenience thing, especially if there are `v-if` statements so the user can pass attributes directly in one single place but if we remove `attrs` then this is no longer necessary

active-rfcs/0000-router-view-keep-alive-transitions.md
posva: I also need to filter out comment nodes from `slots.default()`

posva:
I wished I could keep the same simplicity as in Vue 2 but the best I could come up with is explicitly passing the `route` prop (https://github.com/vuejs/rfcs/pull/153):


```vue
    <transition name="fade">
      <router-view :route="$route" />
    </transition>
```

If I could automatically make that when no `route` prop is based, that would be perfect. @yyx990803 is there a way to do so?
If not, with a warning in dev mode detecting the transition/keep-alive and the absence of the `route` prop, I already think this is easier than the slot version but not as flexible

Adding `DeepTransition`, an alternative to `Transition` doesn't solve the problem of it being complicated and I think it's as confusing as having to use a slot.

I also thought about adding specific `props` like `transition` + `transition-props` but I don't find it very intuitive and only makes `RouterView` compatible with these specific components but not with others added by the user.

posva:
That's a cool concept but I wish it was inside `Transition` directly, maybe we can introduce a similar concept for advanced integrations? @yyx990803

active-rfcs/0000-router-view-keep-alive-transitions.md
posva: added `route` to slot props

posva:

taoorange:
i am confused,so how can use <router-view> now in Vue3.I am using as this:
`import { createRouter, createWebHistory } from 'vue-router'

const routers = [
  { path: '/todomvc', 
    name: 'todo', 
    component: () => import('../components/todomvc.vue')
  }
]

const router = createRouter({
  history: createWebHistory(),
  routes: routers
})

export default router`
.now,i use it :
`const app = createApp(App)
app.mount('#app')
app.use(router)`
but there is a problem,the page is not rendered until I change some code

taoorange:
why i use vue-router-next like this `router.isReady().then(() => app.mount('#app'))`,is useful。but when i use it like this `app.mount('#app')`,It doesn't work

frlinw:
If you can't hardcode the transition name because it depends on the `Component`, you can create a custom component:
```
<script>

export default {
  transitionName: '',

  methods: {
    getTransitionName (vnode) {
      if (vnode) {
        this.$options.transitionName = vnode.type.transitionName || ''
      }

      return this.$options.transitionName
    }
  }
}

</script>

<template>

  <RouterView v-slot="{ Component }">
    <transition :name="getTransitionName(Component)">
      <component :is="Component"></component>
    </transition>
  </RouterView>

</template>
```
Note: you must add a 'transitionName' option to every component rendered by RouterView

PR #161, opened 2020-04-22
VueTestUtils 2 API proposal for Vue 3

dobromir-hristov:
## TLDR

VueTestUtils 2.x, which targets Vue 3, will introduce a few new methods and remove some less used ones. Usage should stay mostly the same.

- **Breaking:** `sync` mode removed. All mutation methods return `nextTick`, you must `await` them.
- **Breaking:** `find` is now split into `find` and `findComponent`.
- **Breaking:** Removal of some wrapper properties and methods, as they induce bad testing habits or are obsolete. 
- **Breaking:** `shallowMount` stubs defaults slots
- **Breaking:** `setProps` only works for the mounted component. 

**Note:** The API for VueTestUtils 1.x will stay the same and will support Vue 2.x. 

[Rendered](https://github.com/dobromir-hristov/rfcs/blob/vtu-2-api/active-rfcs/0000-vtu-api.md)


active-rfcs/0000-vtu-api.md
```diff
@@ -0,0 +1,268 @@
+- Start Date: 19.04.2020
+- Target Major Version: 2.x
+- Reference Issues: (fill in existing related issues, if any)
+- Implementation PR: (leave this empty)
+
+# Summary
+
+VueTestUtils 2.x, which targets Vue 3, will introduce a few new methods and remove some less used ones.
+
+- **Breaking:** `sync` mode removed. All `wrapper` methods return `nextTick`, you must `await` them
```
xanf, line 10: I believe `must` is too restrictive here - there *might* be use cases when you wasn't to perform multiple actions. Maybe we could replace this one with `can`.

Also `await` is also a bit "opinionated" - not all codebase are using `async / await`. Mentioning that it returns Promise might be better

active-rfcs/0000-vtu-api.md
```diff
@@ -0,0 +1,268 @@
...
+- **Breaking** - No longer returns `WrapperArray`.
+- **Breaking** - Accepts query selector only. 
+
+#### findComponent
+
+[Link](https://vuejs.github.io/vue-test-utils-next-docs/api/#findcomponent)
+
+`findComponent` can find component instances, deeply nested in your component tree. This is most useful for edge case assertions, that are not reflected in the DOM or when using `shallowMount` and asserting props on a stub.
+ 
+Users should use `find` for the majority of cases, asserting the DOM properties, not the Vue components.
```
xanf, line 108: Without proper explanation this line definitely sounds confusing and opinionated. What benefits will end user loose (from VTU part, not from "testing philosophy") if he/she chooses to go with `findComponent` approach? 

It would be really great if library specifically pick which end of an egg a person cracks to explain downsides for library consumers

 Is this line an official Vue.js team recommendation? How this aligns with Vue3 being split into "core" and "renderer" parts, so not all Vue components might

active-rfcs/0000-vtu-api.md
dobromir-hristov: Correct! Users can choose to not await these, the dom would not be yet updated. Will reword it :)

posva: Being opinionated here is totally fine.
I think recommending people to `await` those is better than finding yourself not knowing why your test is passing (or not).
The sentence also makes sense given it's noted alongside the removal of `sync` mode, which makes this _awaiting_ necessary.

Recommending `await` against `then` is pretty common in test environments:

```js
it('test', async () => {
  await wrapper.setProps({ one: 'one' })
  expect(wrapper.find('a').classes()).toBe('active')
  await wrapper.setProps({ two: 'two' })
  expect(wrapper.find('a').classes()).not.toBe('active')
})
```

vs

```js
it('test', () => {
  return wrapper.setProps({ one: 'one' }).then(() => {
    expect(wrapper.find('a').classes()).toBe('active')
    return wrapper.setProps({ two: 'two' })
  }).then(() => {
    expect(wrapper.find('a').classes()).not.toBe('active')
  })
})
```

To make the sentence clear, it could say 

> return a promise returned by `nextTick()`

or something similar

active-rfcs/0000-vtu-api.md
dobromir-hristov: Also correct. Chosing how to test is up to the developer. We decided we cannot enforce just one way of querying.

'findComponent' finds only Vue Components and cannot find plain dom elements. We hope this allows test suites running shallowMount to still be able to assert child Components, while making the tests read a bit cleaner, plus the result type be more predictable.

Will reword this a bit, thank you.

posva: The points you are raising are very likely to be addressed in the docs. From my understanding, this sentence is an argument for the sentence right before

active-rfcs/0000-vtu-api.md
```diff
@@ -0,0 +1,268 @@
...
+
+[Link](https://vuejs.github.io/vue-test-utils-next-docs/api/#global-stubs) - Moved to `global.stubs`. 
+
+- **New** - Stubs will no longer render the slots of a component. This can be enabled by a global flag `config.renderStubSlots = true`.
+
+#### mocks
+
+[Link](https://vuejs.github.io/vue-test-utils-next-docs/api/#global-mocks) - Moved to `global.mocks`
+
+#### provide
```
xanf, line 68: While having everything affecting `global` namespace in `global` is perfect, thank you for this, I'm a bit concerned about `provide`. From what I understand We have two things in "global":

* things which are registered globally in real world application (components, directives, mixins, plugins)
* things which are VTU-specific but affect each component in the tree (stubs, mocks)

`provide` is just providing data for `provide/inject`, and for example providing `foo` in `global.provide` does not guarantee that my value will be always injected in component somewhere deep in the tree - it could be easily changed by `provide` in some ancestor

Maybe it's worth to keep this our of `global` object since its behavior is different?

active-rfcs/0000-vtu-api.md
xanf: @posva Hmm, while I definitely support `async / await` as being way more readable, I was confused by assumption, that since Vue3 is targeting (also) IE11, VTU will follow and will not use any syntax that target platform does not support

If this is intended decision to use `async / await` elsewhere it's ok :)

dobromir-hristov:

kspackman:
I am in the process of upgrading to v1.0 and working through the deprecations within our application. Most everything I can use an alternative to achieve the same end, or rewrite to improve the test, which is great.
I have one proposed change that I have concerns about though. We often have parent components that may contain multiple instances of the same component as children, but we want to target a specific instance within the unit test. We often will do something like the following:
wrapper.find('thead').findComponent(MyComponent)
But with the proposed changes, this cannot be done. The only feasible alternative I see to be able to still target a specific component is to add refs to everything. I don't like adding things to production code solely for the sake of unit testing, though, so I really don't want to do that.
Do you have any other alternatives to allow me to still be able to target specific instances of components?

pumano:
I have one question about `find`. Why it deprecated? I want to use that API and query elements by querySelector. Where I can find discussion about that changes?

All test utils provide similar API, and looks like new findComponent API not cover all query strategies

lmiller1990:
@kspackman Sorry no-one replied to your comment. I also missed this.

There is not really a great 1:1 migration path for that particular use case. The idea of separating `findComponent` and `find` are because the returned wrapper `VueWrapper` and `DOMWrapper` have pretty different APIs and usages.

You should still be able to chain `findComponent` of each other - not sure if this helps in your specific use case. Adding a `ref` or something might be the only way forward. You could also just do `findAllComponents` (returns an array) and access the instance you want like that.

lmiller1990:
I think we can merge this - most of these are not only implemented in VTU v2 now, but also back ported to v1 when they could be.

posva:
@lmiller1990 yeah, we can merge this. I couldn't push to the branch so I will merge and rename the file locally

dobromir-hristov:
Thanks!

PR #175, opened 2020-05-15
Props can be provided with a setter.

infinnie:
[Rendered](https://github.com/infinnie/rfcs/blob/patch-1/active-rfcs/0000-props-can-be-provided-with-setter.md)

### DIFF
--- a/active-rfcs/0000-props-can-be-provided-with-setter.md
+++ b/active-rfcs/0000-props-can-be-provided-with-setter.md
@@ -1,0 +1,104 @@
+- Start Date: 2020-5-15
+- Target Major Version: 2.x, 3.x
+- Reference Issues: N/A
+- Implementation PR:
+
+# Summary
+- Make it simpler to pass `v-model` down in (say) wrapper components.
+
+# Basic Example
+```html
+<template>
+  <label>
+    {{label}}
+    <input v-model="value" />
+  </label>
+</template>
+<script>
+  export default {
+    model: {
+      prop: 'value',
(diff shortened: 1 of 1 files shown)


ycmjason:
I'm totally loving this idea.

ycmjason:
On second thought, it will be quite weird if I did `props.x = 3` but it doesn't actually set x to `3`.

phiter:
I like the idea because I see myself doing this quite often:

```
<dialog v-model="isOpen" />

props: {
    open: {
        type: Boolean,
        default: false,
    }
},
computed: {
    isOpen: {
        get() {
            return this.open;
        },
        set(v) {
            this.$emit('update:open', v);
        }
    }
}
```

Adding a setter to the props would definitely reduce this boilerplate that I have to use for some wrapper componnets.

pikax:
if using composition-api you can have a helper to handle that: 

```js
function useVModel(props, name, emit){
return computed({ 
  get(){
    return props[name]
  },
  set(v){
    emit('update:'+name, v);
  }
})
}

defineComponent({
      template: `<div>
      <label :for="child-input">Update value</label>
      <input name="child-input" v-model="value">
      </div>`,
      props: {
        value: String
      },
      setup(props, {emit}) {
        const value = useVModel(props, "value", emit);

        return {
          value
        };
      }
    })
```

[With typescript](https://github.com/pikax/vue-composable/blob/master/packages/vue-composable/src/misc/vmodel.ts)

PR #176, opened 2020-05-31
Redesign the position object returned by `scrollBehavior`

posva:
[Rendered](https://github.com/posva/rfcs/blob/router/scroll-position/active-rfcs/0000-router-scroll-position.md)


JosephSilber:
How does this handle RTL-scrolling elements, where the `x` offset is a `right` offset instead of a `left` offset?

posva:
@JosephSilber since it's using `scrollTo` under the hood, it will handle it the same way. AFAIK it's still `left`. (This behavior doesn't change compared to previous versions, `x` was being used as `left`)

Atinux:
Looking really good! Is the scrollBehavior called after the transition? Because using querySelector May return null if the transition is not done right? 🧐

posva:
> Looking really good! Is the scrollBehavior called after the transition? Because using querySelector may return null if the transition is not done right? 🧐

Like with v3, you need to wait for the transition to be finished inside of scrollBehavior. I didn't mention it but it still supports a Promise, like in v3

posva:

PR #184, opened 2020-07-02
add aborted and isNavigationFailure

posva:
[Rendered](https://github.com/posva/rfcs/blob/router/amend-navigation-failures/active-rfcs/0033-router-navigation-failures.md)

### DIFF
--- a/active-rfcs/0033-router-navigation-failures.md
+++ b/active-rfcs/0033-router-navigation-failures.md
@@ -29,7 +29,7 @@ router.beforeEach((to, from, next) => {
 Then the promise returned by `router.push` rejects:
 
 ```js
-router.push('/url').catch(err => {
+router.push('/url').catch((err) => {
   // ...
 })
 ```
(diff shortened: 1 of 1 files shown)


active-rfcs/0033-router-navigation-failures.md
```diff
@@ -148,27 +149,49 @@ router.currentRoute // { path: '/login', redirectedFrom: { path: '/profile/dashb
...
+  }
+  // Only duplicated navigations
+  if (isNavigationFailure(failure, NavigationFailureType.duplicated)) {
+    // ...
+  }
+  // Aborted or canceled navigations
+  if (
+    isNavigationFailure(
+      failure,
+      NavigationFailureType.aborted | NavigationFailureType.canceled
```
kiaking, line 181: Nit pick but this should be `NavigationFailureType.aborted || NavigationFailureType.canceled` (double pipe)?

posva: No, it's a bitwise OR. People can still do

```js
isNavigationFailure(failure, NavigationFailureType.aborted) || isNavigationFailure(failure, NavigationFailureType.aborted)
```

I think we could show this one first and say in a note we can use the bitwise operator as well. I use the bitwise one inside the codebase to make it smaller

kiaking: Ahhh I see thanks! Good to know 😳

posva: I unresolved the question in case other people want to provide their input too 🙂

posva:

PR #185, opened 2020-07-06
Lazy load for components

rhengles:
**[Full Rendered Proposal](https://github.com/arijs/rfcs/blob/lazy-load-assets/active-rfcs/0000-lazy-load-assets.md)**

### DIFF
--- a/active-rfcs/0000-lazy-load-assets.md
+++ b/active-rfcs/0000-lazy-load-assets.md
@@ -1,0 +1,102 @@
+- Start Date: 2020-07-06
+- Target Major Version: 3.x
+- Reference Issues: https://github.com/vuejs/vue-next/pull/1483
+- Implementation PR: (leave this empty)
+
+# Summary
+
+When you create a Vue component, you pass an object with subcomponents that can be called inside.
+This means that all subcomponents must be defined at creation time of the first component.
+But if we passed a function instead of a object, components could be loaded only when and if
+necessary. This would add lazy loading of components into Vue.
+
+# Basic example
+
+```javascript
+var vm = Vue.createApp({
+	setup: function() {
+		return {};
+	},
+	getComponent: function(spellings, instance) {
+		console.log('App GetComponent', spellings, instance);
+		// example of spellings object:
+		// {
+		// 	"raw": "shift-foo",
+		// 	"hyphenated": "shift-foo",
+		// 	"camelized": "shiftFoo",
+		// 	"PascalCase": "ShiftFoo"
+		// }
+		// In reality, you should return a component corresponding to the name in the object above
+		return {
+			template: '<div><p>She\'s like the {{state.shesLikeThe}}</p></div>',
+			setup: function() {
+				var state = Vue.reactive({
+					shesLikeThe: 'wind'
+				});
+				return {
+					state: state
+				};
+			}
+		}
+	}
+});
+```
+
+# Motivation
+
+In order to have an architecture of an app with hundreds of organized components which are loaded on demand, and this is the crucial part: **without a build step.** I understand that in a high traffic application, you need the optimization opportunities that a build step provides.
+
+However, in my company case, and especially because of the back-end engineers, they want to be able to give maintenance and support to our shipped applications. We're short staffed with front-end engineers, só they leave only the heavy development for us. We have some applications that require constant changes because of business requirements, and they want to do them quickly.
+
+They are used to the old way of doing things, where you had one page with some jquery commands, and they think they could deliver the necessary changes much faster back then, without a build step.
+
+In our case, the performance of our applications is very satisfatory to our clients, but we need to deliver the changes faster, so our bottleneck is in our development team's time. But at the same time, we recognize that to develop a modern and complex application, but in an organized fashion, we need to break It down into components, defining a very clear separation of concerns.
+
+The only way we can develop these components separately and deploy them without a build step is by loading them asynchronally on runtime. We've defined a naming pattern and a structure, a way in which our components must be created. With these rules, we can tell Vue how to load our components without having a pre-defined component name list, which, without a build step, would have to be manually mantained.
+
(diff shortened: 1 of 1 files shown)


CyberAP:
Lazy load for components is already supported in Vue 2 by defining component as a function (`foo: () => import('./Foo.vue')`) and in Vue 3 via [RFC #26](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0026-async-component-api.md).

yyx990803:
This has been in Vue 2 since the very beginning: https://vuejs.org/v2/guide/components-dynamic-async.html#Async-Components

rhengles:
This has left out the part about "no webpack" and "no build step". Could I get an answer about that? I wrote extensively about it on "motivation".

rhengles:
@CyberAP Also, about RFC #26, it requires that you predefine the component names. My proposal doesn't have this limitation. The compiler finds a component name in the template and calls a function to get that component definition. All I'm asking is that my code can provide that function. Why are we so constrained with build tools? I'm sure there's an audience that would prefer to get rid of them if possible.

The rfc template mentions that increase of conplexity is a concern for Vue codebase. Why then the complexity of webpack is shoven down our throats? I'm trying to reduce the complexity of the final application. This is such a simple change, what is the motivation to decline it? I honestly can't understand.

CyberAP:
Sorry I don't understand what exactly are you proposing from reading the rendered proposal.
If you need runtime template compilation it is already possible. You don't have to define a component name in order to create a component in runtime:

```html
<component :is="{ template: `<div>hello world</div>` }" />
```

It also does not require a build step.

yyx990803:
The API doesn't have a hard requirement on the build setup. You can do anything inside the async component factory, including custom AJAX/fetch requests.

rhengles:
@CyberAP what? How would that resemble anything that looks like a organized codebase? You mean I have to replace every single component call into <component is="foo"></component>? Really? Every single one?? That makes for a very ugly template! No one would like to do that, it's like writing render functions by hand! Might as well write my web app in C.

privatenumber:
A build step (eg. Webpack) should only be necessary for compiling the SFC. You might not even need it considering [Codepen seems to offer client-side support](https://blog.codepen.io/2020/04/14/a-custom-editor-just-for-vue-single-file-components-vue-files/) ([vueify](https://github.com/vuejs/vueify)?).

Here's an example using async components without a build: https://jsfiddle.net/hirokiosame/076s9em4/

Here's an example using globally registered async components (not in `components` hash): https://jsfiddle.net/hirokiosame/076s9em4/3/

(I simulate an external file using `URL.createObjectURL(new Blob(...))`, which you can ignore)

Hope that helps

rhengles:
## Thank you @privatenumber for your genuine effort into writing those examples and trying to help me. 🙏

Wait, what? There exists a function called

# `resolveComponent`?

😳😱 That's exactly what I needed!!!!!!

However, there's one small issue - it doesn't fallback to the standard resolving of components if I return `undefined`.

Example:

```
Vue.resolveComponent = function(name) {
	console.log('ResolveComponent', name);
};
// output:
// => ResolveComponent block--header
// => ResolveComponent router-view
// => [Vue warn]: Invalid vnode type when creating vnode: undefined. 
//  at <Anonymous> 
//  at <AsyncComponentWrapper>
```

**How do I resolve `router-view` and every other component created by plugins?**

rhengles:
Ah, I got it:

```javascript
var originalResolveComponent = Vue.resolveComponent;

Vue.resolveComponent = function(name) {
	console.log('ResolveComponent', name);
	return originalResolveComponent.apply(this, arguments);
};
```

# I ❤ U @privatenumber !

rhengles:
@privatenumber @CyberAP 

Now I'm having an issue trying to combine my solution above with `<component>`.

Here's an reproduction link: https://jsbin.com/bejupuyuli/edit?html,js,console,output

Can anyone shine a light on this issue? Is it related to `Vue.resolveDynamicComponent()` ?

rhengles:
@privatenumber @CyberAP @underfin

Oh, I had to extend `resolveDynamicComponent` as well. I don't understand the relation between RC and RDC. Why doesn't RDC consume the output of RC by default?

```javascript
var originalRDC = Vue.resolveDynamicComponent;

Vue.resolveDynamicComponent = function(name) {
	var res = Vue.resolveComponent(name);
	if (!res || 'string' === typeof res) {
		return originalRDC.apply(this, arguments);
	} else {
		return res;
	}
};
```

rhengles:
@privatenumber @CyberAP @underfin

Since today is a two-month "anniversary" of last post, I would be very grateful if the purpose of the `resolveDynamicComponent` function would be explained. The docs don't mention it and I couldn't find the reasoning by search, maybe if you could just point me in the right direction that'd already be enough.

Thanks!

privatenumber:
I'm not apart of the core team or have any involvement with building Vue 3, and I'm still using Vue 2 in my projects so I'm probably not the best person to ask.


Based on the source code ([`resolveComponent`](https://github.com/vuejs/vue-next/blob/89e9ab8a2a/packages/runtime-core/src/helpers/resolveAssets.ts#L19) & [`resolveDynamicComponent`](https://github.com/vuejs/vue-next/blob/89e9ab8a2a/packages/runtime-core/src/helpers/resolveAssets.ts#L30)), it seems like the differences are:

- `resolveComponent` only supports a component name (`name: string`), and warns if it can't resolve it. If it can't resolve, it returns the name as a string.
- `resolveDynamicComponent` accepts any type for `component`. If it's a `string`, it tries to resolve it but won't warn if it can't. If it's not a string, it warns but fall through to creating a `vNode`.

There's probably more if you dig deeper.

---


`vue/rfcs` is a place to propose changes to Vue. When you have questions in the future, try the forums, chats, or Stack Overflow (as recommended at the top of the page [here](https://new-issue.vuejs.org/)). People that are proactively looking to help browse these places so you'll get more traffic there.


Alternatively, if it's a simple question "What's the difference between Vue 3's RDC and RC?", you might have some luck tweeting at some of the [core members](https://github.com/orgs/vuejs/people) -- they're very friendly and helpful.

PS If you want more responses, be less wordy with posts as most people are busy and want to skim. A lot of people won't even read if it's too long.

Good luck

PR #187, opened 2020-07-08
Allow guards to return instead of calling the `next` callback

posva:


active-rfcs/0000-router-return-guards.md
```diff
@@ -0,0 +1,110 @@
...
+
+# Summary
+
+Allow navigation guards to return the value **or a Promise** of it instead of calling `next`:
+
+```js
+// change
+router.beforeEach((to, from, next) => {
+  if (!isAuthenticated) next(false)
+  else next
```
ycmjason, line 14: did you forget to call it? `()`?

ycmjason: Exactly why this RFC is needed.

> - Avoid forgetting to call `next`

posva, line 14:
```suggestion
  else next()
```

michalsnik:
I really like the proposed change. In fact it's actually exactly what I was doing in my project with the help of extra utility I wrote to be able to more easily compose different guards. The individual guards look exactly as you propose here, with the `return` that can either contain a route to be redirected to, or false to cancel any further action.

It looks more or less like this:
```ts
function authRequired(to, from) {
  if (!isAuthenticated) {
    return { name: routes.ROUTE_LOG_IN };
  }
}

function emailVerificationRequired(to, from) {
  if (!isVerified) {
    return { name: routes.ROUTE_VERIFY_EMAIL };
  }
}

async function orgAccessRequired(to, from) {
  try {
    await checkOrgAccess(to.params.orgId);
  } catch (err) {
    // ...
    return { name: routes.ROUTE_403 }
  }
}
```

However, in order to keep things even easier to use and understand just by scanning routes file, we implemented ability to compose guards like so:

```
{
  path: "/dashboard",
  name: routes.ROUTE_DASHBOARD,
  component: () => import("@/pages/Dashboard.vue")
  beforeEnter: composeGuards([
    authRequired,
    emailVerificationRequired,
    orgAccessRequired
  ])
}
```

So if we return in any guard, none of the following guards will be executed, this way we can create guards that are more easily testable and understanding guards logic is super easy.

I think it might be a good idea to add support for this on the router level too.

posva:
Arrays in `beforeEnter` are already supported in v4 🙂

michalsnik:
Nice! Can't wait to start using it and removing the extra logic we have in order to support it :)

ByScripts:
Isn't it already possible to write `router.beforeEach(() => next(isAuthenticated)` ?

posva:
it would become `router.beforeEach(() => isAuthenticated)`

ByScripts:
@posva Yes, I was answering @michalsnik saying he wants to remove extra logic :)

Anyway, your proposal goes further by allowing the return of a Promise 👍

posva:

anthonygore:
> [Rendered](https://github.com/posva/rfcs/blob/router/navigation-guards-return/active-rfcs/0000-router-return-guards.md)

Link should be updated to https://github.com/vuejs/rfcs/blob/master/active-rfcs/0037-router-return-guards.md

PR #194, opened 2020-07-22
Style and Class binding improvements

CyberAP:
[**Rendered**](https://github.com/CyberAP/rfcs/blob/improve-style-and-class-bindings/active-rfcs/0000-style-and-class-binding-improvements.md)

Introduce a couple of improvements to Style and Class bindings:

* Secondary argument for direct prop-value bindings

  ```html
  <input :class:is-active="isActive">
  ```

* Modifiers for secondary style arguments

  ```html
  <input :style:custom-property.prop="customPropertyValue">
  <input :style:width.px="width">
  ```

### DIFF
--- a/active-rfcs/0000-style-and-class-binding-improvements.md
+++ b/active-rfcs/0000-style-and-class-binding-improvements.md
@@ -1,0 +1,212 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- Target Major Version: 3.x
+- Reference Issues: (fill in existing related issues, if any)
+- Implementation PR: (leave this empty)
+
+# Summary
+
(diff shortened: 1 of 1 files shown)


posva:
This was asked for before at https://github.com/vuejs/vue/issues/4956
Same for implicit prop values https://github.com/vuejs/vue/pull/2877

Given the existing syntax for directives (like `v-bind`) this introduces a custom attribute named `class:property` (property being dynamic) and a custom behavior for v-bind, which ironically helps to making things overcomplicated which is your first point in motivation

CyberAP:
It does not introduce a custom attribute and is a syntax sugar for an existing object configuration for v-bind. That's why it said that users who use render functions won't be affected by this change.

PR #203, opened 2020-08-24
[Amendment] add `created` custom element hook

yyx990803:
Adds a new custom directive hook: `created`.

The `created` hook is called before the vnode's props are applied. This is useful in cases where the directive need to attach event listeners before vnode prop or template listeners, or it intentionally want its DOM mutations to take a lower priority than vnode props.

/cc @vuejs/docs


yyx990803:
Relevant commit in core: https://github.com/vuejs/vue-next/commit/11804fe93f66d43320498bfea564af1c9f7a9eb7

PR #206, opened 2020-08-30
Change model update event name

CyberAP:
Change:

* `update:modelValue` to `modelUpdate`

* `update:fooValue` to `modelUpdate:foo`

[**Full rendered RFC**](https://github.com/CyberAP/rfcs/blob/change-model-update-event-name/active-rfcs/0000-change-model-update-event-name.md)


CyberAP:
I was mislead that `v-model:foo` would emit an `update:fooValue` event, while in reality it actually emits an `update:foo` event instead. The way it works right now is fine.

PR #212, opened 2020-09-20
Reactivity's `effectScope` API

antfu:
[Rendered](https://github.com/vuejs/rfcs/blob/reactivity-effect-scope/active-rfcs/0000-reactivity-effect-scope.md)

Draft PR: https://github.com/vuejs/vue-next/pull/2195


CyberAP:
Not strictly against this proposal, but I don't completely understand the motivation point. I have several questions if you don't mind:

1. What are the real word applications of this? How frequently does this happen\might happen?
2. What kind of problems do require that style of authoring reactivity?
3. What are the challenges of collecting stop handlers? (See example below)
4. How wrapping this in a callback scope would solve this? (maybe creating a special higher-order watcher could work in that case)
5. How is it going to affect Composition API? In particular how it would interplay with the existing `setup` cleanup behaviour?

Also the code could be authored in a more simplified way, which might help with the issue:

```js
const disposables = []

const counter = ref(0)
const doubled = computed(() => counter.value * 2)

disposables.push(() => stop(doubled.effect))

disposables.push(
  watchEffect(() => {
    console.log(`counter: ${counter.value}`
  }))
)

disposables.push(watch(doubled, () => console.log(double.value)))

disposables.forEach(f => f())
disposables = []
```

antfu:
1. If you like to make your own a framework. For example, [@vue/lit](https://github.com/yyx990803/vue-lit), it does not handle effects in the component instance lifecycles, which will cause mem leakage if you try to mount and unmount the sub-component multiple times. In order to solve this, you will need to implement `recordInstanceBoundEffect` for your framework, which is not an easy & simple task from the outside. And also they will lose the ability to be reused interchangeably with Vue's ecosystem.

2. Clean up side-effects, preventing mem leak.

3. I don't see your code being "simplified". A huge amount of `disposables.push(` will make the logic hard to read. And if you take the following example, the thing we become much more complex.

```ts
import { useMyLogic1, useMyLogic2 } from './utils'
import { useLogicFromLibrary } from 'xxx-compsoable'

const disposables = []

const { state1, disposables: disposables1 } = useMyLogic1()

disposables.push(...disposables1)

const { state2, disposables: disposables2 } = useMyLogic2()

disposables.push(...disposables2)

const state3 = useLogicFromLibrary()
// it does not collect the disposables, how do you dispose it? (note this is ok in setup)
```

4. Similar to what setup does.

5. This actually more like "extracting the setup functionality, generalized and abstract into @vue/reactivity", and the goal is to be reused for `setup()` as well.


Thanks for the feedback, and hope I answered your concerns.

CyberAP:
> And also they will lose the ability to be reused interchangeably with Vue's ecosystem.

Why is that important?

> This actually more like "extracting the setup functionality, generalized and abstract into @vue/reactivity", and the goal is to be reused in setup() as well.

This resembles a solution to the specific component-based framework problems, but outside of these frameworks what would be the case when you need to do a lot of effect creation and disposal? Cases when we need to do a lot of cleanup are of the most interest to me.

antfu:
> Why is that important?

It's not that important, but why not if you can? Like I might be able to set up an alias and make VueUse work for @vue/lit, so I don't need to reinvent a set of utils that doing the same thing for it or yet another framework. They could even contribute back to Vue's ecosystem.

> what would be the case when you need to do a lot of effect creation and disposal?

Even outside of Vue's instance this could be a problem. Maybe I would like to build a state management system my own, I might need some side-effects and to be cleaned afterward. I need to collect them.

More widely, I can use it in VSCode extension (I already did), I need to watch on config changes, editor changes, file system changes... I would also need to clean them up when I no longer need some. I might use it in a server to do something like HMR updates, that once some file gets deleted, I also need to dispose the watchers.

There are too many possibilities for this great reactivity system. And we just got it for only a while now.

jods4:
I think this is nice because the concept is already in Vue but not cleanly abstracted. 
The fact that Vue itself needs this shows that there is a use case. 
It's already in the framework, so instead of making a special case just for components, why not extract a reusable API?

It won't get much use when `@vue/reactivity` is used in UI though, as Vue already provides the very useful Component == reactive scope, but it could be handy if you're doing something that is not based on components. And it's impossible to do from outside, without heavy hacks.

> In particular how it would interplay with the existing setup cleanup behaviour?

If this is accepted, the existing cleanup should 100% be based on this new API.

This would give us an easy way to perform something that is really hacky today: escape component lifetime (some users have already asked how to do this in issues during the beta).
Opening a nested scope would escape the component scope:
```js
setup() {
  // Stopped when component is unmounted
  watch(...)
  // This one won't be stopped automatically
  effectScope(() => {
    watch(...)
  })
}
```

I'm wondering if this API could have more advanced uses by exposing the captured effects, maybe through an `effects` property on the returned stop function?
To be honest I have no use case for that right now. I was thinking along the lines being able to force a refresh of everything in a scope, or maybe for tooling / debuggers.

privatenumber:
Great proposal! I love to see support for making `@vue/reactivity` work more conveniently as a standalone package, especially since being used independently of Vue is advertised as a pretty big feature of Vue 3 (eg. brought up multiple times in the Vue 3 announcement keynote).


@jods4 

If `@vue/runtime-dom` `setup` is refactored to use `effectScope` behind the scenes, I think any new `effectScope` instantiated inside `setup` will also be cleaned up as per the [Nested Scopes spec](https://github.com/vuejs/rfcs/blob/reactivity-effect-scope/active-rfcs/0000-reactivity-effect-scope.md#nested-scopes).

I think this can still be worked around though, simply by instantiating `effectScope` outside of `setup`, but if users _really_ want to create a scope unbound to the component lifespan, I think they'll always find a way.

jods4:
> If @vue/runtime-dom setup is refactored to use effectScope behind the scenes, I think any new effectScope instantiated inside setup will also be cleaned up as per the Nested Scopes spec.

That's a good point.
I think Components don't work like that today -- they only stop effects that were created inside them, not their children (which get stopped by unmounting the whole tree, but that's not the same mechanism).

This is incentive to change the spec slightly so that nested effectScopes are not stopped by the outer one, so that Components can be implemented on top of that spec.
If there is a good use-case for the opposite behavior, then maybe this could be controlled by an optional flag passed to `effectScope` (if there is no use case today, such a flag can always be added later).

IMHO not implementing components on top of this spec feels bad because:
- it's exactly the kind of use-case for that spec, it'd be weird that the framework itself can't make use of it;
- it's duplicated code for practically the same thing;
- then the interaction between component scope and `effectScope` must be specified clearly, leading to added complexity, possibly weird edge cases and unnecessary coupling between reactivity and components.

>  think this can still be worked around though, simply by instantiating effectScope outside of setup, but if users really want to create a scope unbound to the component lifespan, I think they'll always find a way.

It's been asked already and I know of at least two ways you can achieve this today... but both are hacky. It'd be a nice bonus if this spec allowed that in a clean way through a documented API.

antfu:
Update: `effectScope` now returns the scope instance instead of the stop handler. Providing more flexibility and make the `stop` api more general.

```ts
const scope = effectScope(() => {
  const doubled = computed(() => counter.value * 2)

  watch(doubled, () => console.log(double.value))

  watchEffect(() => console.log('Count: ', double.value))
})

// to dispose all effects in the scope
stop(scope)
```

unbyte:
I really like its concept
however I have doubts about it 🙋‍♂️

1. in most cases computed values would be used as meaningful variables, and closures complicate its use. and it will lead to some bug codes if computed values can be exported from the closure, since stopping `effectScope` breaks reactivity of exported computed values that may still be used without notification.

2. if stopping a parent scope also stops all child scopes, how to create a standalone scope inside a scope (it's useful)? and if can by certain api, how to stop this standalone scope since it's in a nested closure simply?

CyberAP:
What do you think of attaching effects to the scope after scope creation? That might come handy for effects that are initiated asynchronously.

```js
const scope = effectScope()

setTimeout(() => {
  scope(() => {
    watch(...)
  })
})
```

Or attach by pushing:

```js
const scope = effectScope()

setTimeout(() => {
  scope.push(watch(...))
})
```

If the scope has been destroyed the effect should not be created.

antfu:
@unbyte 
> and it will lead to some bug codes if computed values can be exported from the closure

Can't image it, can you provide some example of why to use computed outside the scope? Need some context so maybe I can think about how to change the design to fit the equipment.

> how to create a standalone scope inside a scope

Nested scope should be collected by its parent scope, but can still be disposed explicitly

```ts
const scope1 = effectScope(() => {
  

  const scope2 = effectScope(() => {
    watchEffect(/* ... */)
  ))
  
  stop(scope2)
})
```

@CyberAP 

> What do you think of attaching effects to the scope after scope creation?

With the last change, `effectScope()` returns the scope instance instead of the stop handler. You can do that by 

```ts
const scope = effectScope()

setTimeout(() => {
  scope.effects.push(watch(...))
})
```

unbyte:
@antfu 

> Can't image it, can you provide some example of why to use computed outside the scope? Need some context so maybe I can think about how to change the design to fit the equipment.

I may get you... so you mean all related logical codes are inside the scope right?

> Nested scope should be collected by its parent scope, but can still be disposed explicitly

😂 my mean is how to make the child scope have a longer life than the parent scope and to be stopped outside, for example https://github.com/vuejs/vue-next/issues/1532

mathe42:
> 😂 my mean is how to make the child scope have a longer life than the parent scope and to be stopped outside, for example [vuejs/vue-next#1532](https://github.com/vuejs/vue-next/issues/1532)

Maybe we can have a namespace so 
```js
const scope1 = effectScope('namespaceA', () => {
  
  const scope2 = effectScope('namespaceB', () => {
    watchEffect(/* ... */)
  ))
})
stop(scope1)
```

So scope2 is still running. Also the namespace could be optional and when omitted it is useing a default namespace.

unbyte:
@mathe42 so in this way users should stop a scope by its name string. it looks easy but not so good for coding (personnally). 

and the dependency of each layer in the nested scopes is not always one-way, which may cause some problem when clean up effects under certain namespace.

maybe I have the wrong idea to solve this problem

antfu:
vuejs/vue-next#1532 seems like a reasonable use case, that would be great if we can solve it along with this RFC. I am thinking if we could add an option to it and make it not be collected by its parent. Something like:

```ts
let scope2

const scope1 = effectScope(() => {
  watchEffect(/* ... */)

  scope2 = effectScope(() => {
    watchEffect(/* ... */)
  ), { escaped: true })
})

stop(scope1)
// scope2 will still be alive until you dispose it
```

antfu:
Update: "escaped" options added

A draft PR is filed with the feature itself implemented. https://github.com/vuejs/vue-next/pull/2195

unbyte:
@antfu if reuse it in `runtime-core`,`effectScope` may have to return some values for rendering

maybe it can be like this
```js
const scope = effectScope(()=>{
    return {
        // some value
    }
})
scope.exported // visit exported value
```

antfu:
Actually, I have already made the refactor to runtime-core in the draft PR and all tests are passing ;)

> effectScope may have to return some values for rendering

Did through about that a bit, can find a clean design for it. And TBH, I don't think this feature is required, you can do it like

```ts
let result
const scope = effectScope(()=>{
    result = {
        // some value
    }
})

console.log(result)
```

unbyte:
TBH I don't so like this solution 😂 but it does't matter

I have a suggestion about `escape`, that can it be combined with `callback`?

a callback on stopping scope is useful. sometimes a scope need to know that it's going to be stopped then it can release certain resources or what, just like how callbacks in `watch` do.

so `escape` can be implemented by that call callback before cleanup, if false ( or falsy value ) is returned, terminate clean up

ryansolid:
I will echo this mechanism at the core of my generic reactive renderer DOM Expressions and all my libraries and call them reactive roots. I had to artificially create this in https://github.com/ryansolid/vuerx-jsx. I think there is the question of nested versus isolated. I've opted to isolated in the past. Children should be able to outlive their parents if desired. I make heavy use of this to control reactive context independent of components. Not sure its worth anything, but just wanted to throw in that I think this is a powerful and worthwhile feature.

antfu:
Update Proposal: 

Forward the returned value to the outside of the scope

Mentioned in

https://github.com/vuejs/rfcs/pull/212#issuecomment-696459422
https://github.com/vuejs/rfcs/issues/234#issuecomment-728955621

While we still need to return the scope instance, maybe we can use the scope extending function to do this.

```ts
const scope = effectScope()

const { x, doubled } = scope(() => {
  const { x } = useMouse()
  cosnt doubled = computed(() => x.value * 2)
  
  // this make it a bit similar to `setup()`
  return { x, doubled }
})

console.log(doubled.value)
```

The limitation for this is that when directly calling `effectScope` can't have the same destructing ability as the extending function. One solution might be under the `exported` props as @unbyte mentioned.

```ts
const scope = effectScope(()=>{
  const { x } = useMouse()
  cosnt doubled = computed(() => x.value * 2)
  return { x, doubled }
})
const { x, doulble } = scope.exported
```

Or nested destructuring

```ts
const { stop, exported: { x, doulble } } = effectScope(()=>{
  const { x } = useMouse()
  cosnt doubled = computed(() => x.value * 2)
  return { x, doubled }
})

console.log(doubled.value)
```

What do you think? Feedback wanted. Thanks!

posva:
I like the `effectScope` version, it's like a `setup` without the component. Not sure what a good name would be, I don't find `effectScope` relatable. `scoped()`, `setup()`?

kiaking:
I also thought naming extended scope function `setup` might look really nice.

```js
const setup = effectScope()

const { x, doubled } = setup(() => {
  const { x } = useMouse()
  cosnt doubled = computed(() => x.value * 2)
  
  return { x, doubled }
})
```

It's just a naming convention but this makes people more aware that `setup` and `effectScope` is a different thing, hence wouldn't be surprised they have different return value.

And of course in this case, I'm for `stop`, `exported` object get returned from `effectScope`.

```js
const { stop, exported } = effectScope(() => { ... })
```

posva:
Oh, I was thinking of directly having one function.

```js
const outOfScopeState = setup(() => {
  const { x } = useMouse()
  cosnt doubled = computed(() => x.value * 2)
  
  return { x, doubled }
})

outOfScopeState.x.value // mouse.x
destroySetup(outOfScopeState)
// or
outOfScopeState.$destroy()
```

kiaking:
Ah I see. You mean having totally separate function rather than getting it as return value from `effectScope` that might work too 👀. That `setup` function can just call `effectScope` inside the function call so. Looks nice and I like it.

mathe42:
What happens when the function is a async function and I use await...? Does effectScope lose it's context after the first await?

posva:
It's one of the drawbacks. I would say the function passed to `setup` should not return a promise. In the use cases I have in mind, they are all synchronous

mathe42:
```ts
const  scope = effectScope(async () => {
    console.log(0)
    const res = await wait10Sec() //resolves after 10 seconds
    console.log(1)
  ), { async: true })

const  scope = effectScope(async () => {
    console.log(2)
  ), { async: true })
```

I think it is posible if we add a async option and say async effectScopes don't run in parallel. So the example above would log 0 then waits 10s and log 1 then 2. And if async options is not provided (or false) and a Promise is returned we should throw.

mathe42:
I have some UseCases with webworkers in mind...

matias-capeletto:
@mathe42 @antfu [wrote a comment](https://github.com/vuejs/rfcs/issues/234#issuecomment-728955621) about a possible way to keep the scope in async setup functions using a helper like:
```ts
export function createInstanceScope(instance = getCurrentInstance()) {
  return (fn, options) => effectScope(()=> {
  	const prev = getCurrentInstance()
  	setCurrentInstance(instance)
    fn() // error handling
    setCurrentInstance(prev)
  })
}
```
If I understand correctly, this could be used in composables too:
```ts
async function useAsyncComposable() {
  const scope = createInstanceScope() // uses the current instance
  const data = await fetch() 
  let processed
  scope(() => {
    processed = useXXX(data)
    watch(/*...*/)
  })
  await fetch()
  scope(() => {
     useYYY() // after await but will use the same instance
   })
}
```

antfu:
@mathe42 
> async effectScopes

the instance context relying on the fact that JavaScript is a single thread and by wrapping the function calls like

```js
    setCurrentInstance(instance)
    fn()
    setCurrentInstance(prev)
```

But when the `fn` becomes an async function, `setCurrentInstance(prev)` will be called before the first `await` calls, and the instance context gets lost after the `await`. This is the same limitation as async components' `setup`. See more discussion in https://github.com/vuejs/rfcs/issues/234

@matias-capeletto 
> `useScoped()`

This is already possible with the current design. Every scope is callable to be extended.

@aztalbot 
> is there a cleanup hook available?

That's a good point, I will add them to the RFC.

> just returned from that single scope call

That's exactly what I was thinking. Merging back the exports feels a bit unintuitive.

> add to a scope after initialization

Yes. That would help for async/await like https://github.com/vuejs/rfcs/pull/212#issuecomment-731298352 pointed out. Or the component `setup()` itself: https://github.com/antfu/vue-next/blob/51ee6dcd22475212374086244340edfbac7ae2e4/packages/runtime-core/src/component.ts#L570-L580

antfu:
@posva I am kinda liking your idea, just forward return value could be much simpler and intuitive.

```ts
const { x, doubled, $stop, $scope } = effectScope(() => {
  const { x } = useMouse()
  cosnt doubled = computed(() => x.value * 2)
  return { x, doubled } // force it to be an object so it can be destructurable
})
```

However, AFAIK, the $ prefixed style is no longer been used in the new Vue 3's Composition API, not sure if we could use them here. 

Another idea:

```ts
import { stop, effectScope } from 'vue'

const scope = effectScope(() => {
  const { x } = useMouse()
  cosnt doubled = computed(() => x.value * 2)
  return { x, doubled } // force it to be an object so it can be destructurable
})

const { x, doubled } = scope

// extend scope
effectScope(() => {
	
}, scope) // pass another scope to extend it

stop(scope)
```

posva:
I like the second idea without the `$`

antfu:
**Updates**: Add `onCleanup` API and returns forwarding behavior.

<details>
<summary>Excerpt</summary>

### Forward returning values

Values returned in the scope function will be forward as the return value of `effectScope` and able to be destructured.

```ts
const scope = effectScope(() => {
  const { x } = useMouse()
  const doubled = computed(() => x.value * 2)

  return { x, doubled }
})

const { x, doubled } = scope

console.log(doubled.value)
```

The return value should be an object.

### onCleanup

A `onCleanup` hook is passed to the scope function

```ts
const scope = effectScope((onCleanup) => {
  const doubled = computed(() => counter.value * 2)

  watch(doubled, () => console.log(double.value))

  watchEffect(() => console.log('Count: ', double.value))

  onCleanup(()=> {
    console.log('cleaned!')
  })
})

stop(scope) // logs 'cleaned!'
```

### Extend the Scope

A scope instance is also a callable function, which allows users to extend the scope and combine their effects.

```ts
const myScope = effectScope()

myScope(() => {
  watchEffect(/* ... */)
})

myScope(() => {
  watch(/* ... */)
})

// both watchEffect and watch will be disposed
stop(myScope)
```

The extending function will also forward its return values

```ts
const { foo } = myScope(() => {
  return {
    foo: computed(/* ... */)
  }
})
```

</details>

antfu:
About the way to extend a scope, I am still not sure about the design combining with the return forwarding design. Would like have some feedback about several design proposals, thanks!

## Proposal A

As the original proposed, a scope is callable to extend

```ts
const myScope = effectScope(() => {
  const bar = ref(1)
  
  return { bar }
})

cosnt { bar } = myScope

const { double } = myScope(() => {
  // extend, the effect created in this function will be append to the 
  const double = computed(() => bar.value * 2)
  return { double }
})
```

### Pros
- Easy to use
- Concise code

### Cons
- Runtime overhead
	- Need to proxy the return object onto the function to make it destructurable while being callable
- A bit confusing for reader to understand the code

## Proposal B

Latest updated in RFC. To use `effectScope` to extend the scope

```ts
const myScope = effectScope(() => {
  /* … */
})

const { double } = effectScope(() => {
  /* … */
  return { double }
}, { extend: myScope }) // pass another scope to extend it
```

### Pros
- Less runtime overhead, we can forward the return as-is.

### Cons
- More code to write
- Conflict with other options, e.g. `{ extend: myScope, detached: true }` which we should throw some warning.
	- `detached` behavior should be define at the scope initialization and does not make sense to be reconfigurable on extending

## Proposal C

Introduce a new API entry `extendScope`

```ts
import { effectScope, extendScope } from 'vue'

const myScope = effectScope(() => {
  /* … */
})

const { double } = extendScope(myScope, () => {
  /* … */
  return { double }
})
```

### Pros
- Less runtime overhead compare to Proposal A
- Do not mess up with other options

### Cons
- New API, increase documentation and maintenance surface.

meteorlxy:
To align with vue hooks, what about `onStop` -> `onBeforeStop` or `onStopped` ? (in fact the hook name does not matter at all lol

kiaking:
@antfu **I like Proposal C**. A little verbose, but very easy to understand that you're trying to "extend" the scope. The intention of the code, and the argument order makes perfect sense, looking good.

I honestly love Proposal A, but being able to call `myScope` and also it acts as an object, is I think too confusing in JS land. Proposal B... I think in this case I prefer proposal C.

znck:
I prefer **Proposal C** for its explicitness. It is easier to analyze statically, which can be used to improve IDE experience if needed.

mathe42:
With Proposal C we could do (I think A and B works too but not 100% sure):

```js
import { effectScope, extendScope } from 'vue'

const myScope = effectScope(async () => {
  /* … */
  const data = await fetch()
  
  return extendScope(myScope, () => {
    /* … use Data */
    return { fetchedData }
  })
})
```

And when useing it in a SFC the compiler could add this without the developer needs write this code...

Additionaly we might want to have a function like `getCurrentScope`...?

antfu:
Thanks for the feedback, RFC updated to match Proposal C. Draft PR https://github.com/vuejs/vue-next/pull/2195 is also updated to align with the latest RFC.

antfu:
Thinking about having an extra hook `onEffectScopeStopped` which serves a similar functionality to `onUnmounted` but works for the current scope. This could benefit help composable functions to cleanup their side effects along with its scope. Since `setup()` also creates a scope for the component, it will be equivalent to `onUnmounted` as we are currently using.

Example A:

```js
function useMouse() {
  const x = ref(0)
  const y = ref(0)

  function handler(e) {
    x.value = e.x
    y.value = e.y
  }
  
  window.addEventLisenter('mousemove', handler)

  // we are doing now:
  // onUnmounted(() => {
  //  window.removeEventLisenter('mousemove', handler)
  // })

  // proposed:
  onEffectScopeStopped(() => {
    window.removeEventLisenter('mousemove', handler)
  })

  return { x, y }
}
```

```ts
// lazy init states

let state = undefined

function useState() {
  if (!state)
	state = effectScope(() => useMouse(), { detached: true })
  return state
}

export default {
  setup() {
    return useState()
  }
}
```

If we are using `onUnmounted` for `useMouse` in the example above, although the effects have been collected and detached with the component lifecycle, the `removeEventLisenter` would still be applied to the component's lifecycle. Which means the `x` and `y` would still lose their reactivity after the component gets unmounted. `onEffectScopeStopped` makes it decoupled more from the component model and being more flexible.

Example B:

```ts
let state = undefined

function useState() {
  if (!state)
	state = effectScope(() => useMouse(), { detached: true })
  return state
}

export default {
  setup() {
    const enabled = ref(false)
    const mouseScope = null

    watch(enabled, () => {
      if (enabled.value) {
        mouseScope = effectScope(() => useMouse())
      } else {
        stop(mouseScope)
        mouseScope = null
      }
    }, { immediate: true })
  }
}
```

In the example above, we would create and dispose some scopes on the fly, `onEffectScopeStopped` allow `useMouse` to do the cleanup correctly while `onUnmounted` would never be called during this process.

Please let me know what you think about this proposal, thanks!

posva:
would it make sense to keep `onUnmounted()` and adapt its behavior if it's inside an `effectScope` instead of adding a new `onEffectScopeStopped()` so `useMouse()` can be called inside a component and also with `effectScope()`?

antfu:
@posva that would do as well. But not sure if it would cause some confusion as `onUnmounted` sound more like to bound with the component model 🤔

posva:
It does lack the intention of `onEffectScopeStopped()` but I don't see users thinking about implementing `useMouse()` calling `onEffectScopeStopped()` so it works on both copmonents and `effectScope()`. `onEffectScopeStoppeed()` could also internally be an alias to `onUnmounted()` to keep the intent

jods4:
This is neat.

Drawbacks mention lack of async support but I don't think it's a blocker for a v1. 
It can be added later and the current version of Vue doesn't support effects in `async` setup either.

There's a long discussion that explores the pros and cons of several ideas to support async (in setup, but `effectScope ` is the same) here: https://github.com/vuejs/rfcs/discussions/234

There's one slight difference to keep in mind between `effectScope` and `setup`, though.

`effectScope` is a reactivity primitive and it only cares about the current scope to stop effects.
In addition, `setup` also cares about the current component, for stuff like calls to lifecycle `mounted`, `unmounted`, etc.

It means that the `effectScope` solution will not completely cover the `setup` requirements.

Here's one idea:
if the async solution is based on a function `withVueContext` or similar, maybe we could have 2 different exports similar to what Vue does for `watch` today:

Module `@vue/reactivity` exports `withVueContext` that preserves just the effect scope, for low-level reactivity users, maybe even outside Vue.

Mode `vue` exports a slightly improved `withVueContext` that preserves the current component (for `unmounted` and co) and calls reactivity `onVueContext` to also preserve the effect scope.

SeregPie:
I hope this will be implemented really soon.

My suggestion would be to align effectScope/extendScope with the setup function to return proxyRefs.

```js
let state = effectScope(() => ({
  a: ref(1),
  b: 2,
}));
state.a; // 1
state.b; // 2
```

mcrapts:
When will this get addressed? Currently, Vuex 4.0.0 getters are not being cached, apparently because of this issue. This seems like a very biig issue to me, because at this moment is it not possible to use Vuex 4.0.0 + Vue 3 efficiently. Getters are re-evaluated everytime they are accessed where they should be cached. This is a huge showstopper for me.

basvanmeurs:
> Actually, I have already made the refactor to runtime-core in the draft PR and all tests are passing ;)
> 
> > effectScope may have to return some values for rendering
> 
> Did think about that once, can't find a clean design for it though. And TBH, I don't think this feature is a must-have, you can do it like
> 
> ```ts
> let result
> const scope = effectScope(()=>{
>     result = {
>         // some value
>     }
> })
> 
> console.log(result)
> ```

It should have stopped here. Imho, forwarding return values is the worst thing that happened to this otherwise awesome PR. I think that it's a useless unnecessary complication. Just use normal js variables in the desired scope as antfu suggested.

0x009922:
(forwarding https://github.com/vuejs/vue-next/pull/2195#issuecomment-873359444)

I don't like "Forward returning values" section:

```ts
const scope = effectScope(() => {
  const { x } = useMouse()
  const doubled = computed(() => x.value * 2)

  return { x, doubled }
})

const { x, doubled } = scope

console.log(doubled.value)
```

Here I do not like the fact that (1) what is returned by the function is affected, since it can generate some conflicts and it is generally not clear why it should be affected, and (2) there is a restriction on returning exactly the object, and it is also not clear why is it.

It seems to me that it would be more elegant and simpler to return a tuple with a return value and a scope, or only a scope if the function is `void` (returns `undefined`):

```ts
const [{ x, doubled }, scope] = effectScope(() => {
  const { x } = useMouse()
  const doubled = computed(() => x.value * 2)

  return { x, doubled }
})

const scope2 = effectScope(() => {
  const counter = useInterval(1_000)
  watchEffect(() => console.log(counter.value))
  // no return
})
```

It seems to me that this is simple both in implementation and in typing.

yyx990803:
So I did a full review of the API design while rebasing https://github.com/vuejs/vue-next/pull/2195. Since this is going to be a very low-level API, the API shape needs to be designed with implementation constraints in mind. I decided to greatly simplify the proposed API in order to obtain smaller size increase and better memory efficiency. [details](https://github.com/vuejs/vue-next/pull/2195#issuecomment-874347785)

I will be revising this PR to reflect the updates.

CyberAP:
What do you think of a more fine-grained control over the scope effect execution?

For example if we want to run effects on some changes (or while some flag is set), but to stop on others?

```js
const foo = new ref(0)
const scope = new EffectScope(() => {
  effect(() => { console.log(foo.value) })
})
foo.value = 1 // runs as usual
scope.pause()
animateValue(foo) // doesn't run during animated change
  .then(() => { scope.resume() }) 
```

If we'd like to replicate current `scope.stop()` we could destroy the scope:

```js
scope.destroy() // can not be resumed nor paused
```

yyx990803:
@CyberAP that's not a designed capability of the effect scope, at least in this RFC. The way it is currently designed is only meant for collecting effects so they can be stopped together.

yyx990803:
I revised the RFC to match the latest implementation - it is now much simpler:

```js
const scope = new EffectScope()

const returnValue = scope.run(() => { /* ... */ }

scope.stop()
```

Please read the [full rendered version](https://github.com/vuejs/rfcs/blob/reactivity-effect-scope/active-rfcs/0000-reactivity-effect-scope.md).

In particular, I expanded the usage example to illustrate a new pattern this would unlock: [Shared Composables](https://github.com/vuejs/rfcs/blob/reactivity-effect-scope/active-rfcs/0000-reactivity-effect-scope.md#example-a-shared-composable).

antfu:
Thanks for revising @yyx990803 ! Love the updates and now it's much more concise and elegant!

Just an opinionated thought, would it better align with Vue's other APIs if we use a function to construct the instance instead of `new`? The reason I am thinking this is that in Vue 3 most of the APIs are functions that starts with lowercase letters, `createApp()`, `defineComponent()`, `ref()` while the capitalized exports are more refer to type interfaces, `App`, `Component`, `Ref`. Would it be better if we use something like `effectScope()`, `createEffectScope()` or `createScope()` for this API?

If we do, I think we could have the module implemented like this:

```ts
class EffectScope {
  /* ... */
}

export function effectScope() {
  return new EffectScope()
}

export type { EffectScope }
```

While we could always do it in the user land, just raising this for discussions.

yyx990803:
That makes sense -  added the `effectScope` factory function for API consistency.

yyx990803:

AlexandreBonaventure:
Wow this is truely amazing! 
Funny enough we've built and use in production our own implementation of almost the same API to build shareable compositions for a year already.
https://github.com/AlexandreBonaventure/vue-scoped-composition

This pattern is so powerful and I'm really stoked that Vue is going to provide built-ins for that.
The RFC API is elegant and consistent, great job!

udany:
Looks great, I'd been thinking about this tangentially but hadn't quite put my finger on it before reading this proposal :D

If you guys allow me a last minute suggestion, we could make the syntax shorter for the most common use cases by accepting a function on the constructor/factory that would get automatically run, like this:

```javascript
const scope = effectScope (() => { /* ... */ });
```
Instead of:
```javascript
const scope = effectScope ();
scope.run(() => { /* ... */ })
```

SeregPie:
@udany 
Function `run` returns the value the callback returns and the constructor has an optional `detached` argument.

daniele-orlando:
@antfu @yyx990803 

About the signature of `effectScope`, my comment is that it could be improved

1. from the clearness
2. and future proof

points of view, adopting an options object as first argument, instead of a plain boolean.

### Current Signature
```js
effectScope(detached = false): EffectScope

effectScope(true) // true what?
```

### "Specking" and Future Proof
```js
effectScope(EffectScopeOptions): EffectScope

effectScope({ detached: true })

export interface EffectScopeOptions {
    detached?: boolean
}
```

antfu:
@daniele-orlando Thanks for bringing this up. Just discussed with the team, here is a short summary:

The reason we use a plain boolean for this argument is performance concern. Giving the fact that this API is an advanced low level API, trading a little bit of readability and extendibility allows it to save a few object allocation and restructuring cost (which could be significant on higher level when using the low-level api hundreds of times). You can find the similar reason for the change behind the `effect` api optimization https://github.com/vuejs/vue-next/pull/4001

daniele-orlando:
Hi @antfu and thanks for the answer.

I see the rational behind the choice.
From what I understand, `effectScope` is serving two domains, the public APIs domain and the internal APIs domain.

From the public APIs domain point of view, passing an object should not be a performance problem, given that it is called only during the setup only once or few times per component and only in advanced cases. Similar use case of `watch()` which accepts an options object as third argument.

From the internal APIs domain point of view, passing an object can be a performance penalty and should be avoided.

Given that `effectScope()` is already a wrapper around `new EffectScope()`, would not be the case to provide a user friendly _Public_ `effectScope(options)` API that maps to the _Internal_ and optimised `effectScope(detached: boolean)`?

Just wondering, I'have no knowledge of the Vue internals, so I really don't if it could be too cumbersome.

PR #213, opened 2020-09-24
Cleaner Composition API

johnsoncodehk:
<details>
<summary>Background</summary>

`no-ref` is an idea proposed by @cereschen, and he has a good prototype implementation: https://github.com/cereschen/no-ref
I had some discussions with @cereschen about `no-ref`, and it eventually became this solution.
</details>

When using Composition API, there is no need to write `.value` which will make the code much cleaner.

## Transform Sample

For this code:

```html
<template>
  <button @click="inc">{{ count }}</button>
  <p v-if="done">done!</p>
</template>

<script setup>
import { ref, computed } from 'vue'

export const count = ref(0)
export const done = computed(() => count.value > 100)
export const inc = () => count.value++
console.log(done.value)
</script>
```

When use `no-ref`, the `<script>` can replace to

```html
<script setup noref>
export let count = 0 // @ref
export const done = count > 100 // @computed
export const inc = () => count++
console.log(done)
</script>
```

Rendered: [中文](https://github.com/cereschen/rfcs/blob/no-ref/active-rfcs/0000-sfc-script-no-ref.zh.md) | [English](https://github.com/cereschen/rfcs/blob/no-ref/active-rfcs/0000-sfc-script-no-ref.md)

### DIFF
--- a/active-rfcs/0000-sfc-script-no-ref.md
+++ b/active-rfcs/0000-sfc-script-no-ref.md
@@ -65,7 +65,7 @@ export const baz = computed(() => foo.value + bar)
 ## Multi-line computed
 
 ```html
-<script setup noref>
+<script noref>
 // @computed
 export const baz = (() => {
     let foo = 1
@@ -80,7 +80,7 @@ console.log(baz);
 <summary>Result</summary>
 
 ```html
-<script setup>
+<script>
 import { computed } from 'vue'
 
 export const baz = computed(() => {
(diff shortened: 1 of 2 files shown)


johnsoncodehk:
Close because we are looking for a better way.

social-uni:
Hello, in my opinion, this proposal is very good. Why is it closed? Can you help me with this? Thank you very much.

PR #214, opened 2020-09-27
Script refs

johnsoncodehk:
We understand that most users accept the use of `.value`, but we still hope that we can have some solution for users who do not like `.value`.

Rendered: [中文](https://github.com/cereschen/rfcs/blob/script-refs/active-rfcs/0000-sfc-script-refs.zh.md) | [English](https://github.com/cereschen/rfcs/blob/script-refs/active-rfcs/0000-sfc-script-refs.md)

If you want to try:
`<script refs>` requires additional IDE support, I implemented it in another [Vue Language Service plugin](https://marketplace.visualstudio.com/items?itemName=johnsoncodehk.volar)
For the preview you don't need to compile, but if you want, you can use @cereschen [webpack plugin](https://github.com/cereschen/no-ref)

### DIFF
--- a/active-rfcs/0000-sfc-script-refs.md
+++ b/active-rfcs/0000-sfc-script-refs.md
@@ -47,9 +47,10 @@ export default {
 
 # Motivation
 
-Usually when writing code, we expect that `.value` is not needed to access variables, and the access method of `.value` has no additional benefit to users. The characteristics of `.value` also cause the code to look messy when non-ref variables are used with ref objects.
+Usually when writing code, somebody don't want to write `.value`, and we don't need to access properties other than `value` most time.
+Although it's quite right to write `. value`, it's not elegant.
 
-`.value` is a unified feature of ref objects, so we think we can use the compiler to remove this feature while retaining the response function.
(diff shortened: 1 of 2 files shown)


jods4:
How does that work with imported functions?
How does the compiler know if the function wants a raw value or a ref?
```js
import { square } from "./math.js";
ref x = 3
// What is passed to square? A ref or a number? 
// Both options are common, how can we have the opposite behavior?
square(x)
```

johnsoncodehk:
> How does that work with imported functions?
> How does the compiler know if the function wants a raw value or a ref?
> 
> ```js
> import { square } from "./math.js";
> ref x = 3
> // What is passed to square? A ref or a number? 
> // Both options are common, how can we have the opposite behavior?
> square(x)
> ```
You can't: https://github.com/cereschen/rfcs/blob/script-refs/active-rfcs/0000-sfc-script-refs.md#dont-transform
Before has this feature: https://github.com/cereschen/rfcs/blob/956e74288552b699dafd83f0706430f86c11e0fb/active-rfcs/0000-sfc-script-no-ref.md#dont-transform
In order to reduce the magic, it was removed, welcome any good idea. :)

PR #221, opened 2020-10-23
add vue template syntax supported jsx

caikan:
[Rendered](https://github.com/caikan/rfcs/blob/vue-template-syntax-supported-jsx/active-rfcs/0000-vue-template-syntax-supported-jsx.md)

### DIFF
--- a/active-rfcs/0000-vue-template-syntax-supported-jsx.md
+++ b/active-rfcs/0000-vue-template-syntax-supported-jsx.md
@@ -19,7 +19,7 @@ const MyComponent = {
     return () => (
       <!vt>
         <div> {{ count }} </div>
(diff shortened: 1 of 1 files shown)


antfu:
How could we incorporate with TypeScript support?

caikan:
> How could we incorporate with TypeScript support?

Sorry, I'm not familiar with TS yet. Do you feel that this RFC will cause some problems for TS support?

yyx990803:
Thanks for the proposal. We are not closing this because:

- The proposal can technically be implemented via a userland Babel plugin already.
- For proper IDE support, it will require non-trivial work since it cannot rely on TSX inference.
- With `<script setup>` you already get the benefit of "template in the same scope with js".

PR #225, opened 2020-11-07
Compiler optimization for `with` statements.

caikan:
[Rendered](https://github.com/caikan/rfcs/blob/compiler-optimization-for-with-statements/active-rfcs/0000-compiler-optimization-for-with-statements.md)


yyx990803:
Thanks for the proposal - it's an interesting idea, but the verbosity reduction comes with other readability costs.

Most importantly, we really  shouldn't rely on `with` in userland code, as it won't run in ES modules.

PR #182, opened 2020-06-29
SFC Improvements

yyx990803:
**Note: the Component Import Sugar RFC has been dropped for now due to various unresolved concerns. However, `<script setup>` now supports [using exported values as components](https://github.com/vuejs/rfcs/blob/sfc-improvements/active-rfcs/0000-sfc-script-setup.md#exposing-components).**

This PR includes 2 RFCs related to improving the authoring experience of Vue SFCs (Single File Components):

## `<script setup>` for Composition API in SFCs

**Important: this version of `<script setup>` is being replaced by a new one at https://github.com/vuejs/rfcs/pull/222**

[[Outdated] Rendered](https://github.com/vuejs/rfcs/blob/sfc-improvements/active-rfcs/0000-sfc-script-setup.md)

## State-driven CSS Variable Injection in `<style>`

```vue
<template>
  <div class="text">hello</div>
</template>

<script>
export default {
  data() {
    return {
      color: 'red'
    }
  }
}
</script>

<style vars="{ color }">
.text {
  color: var(--color);
}
</style>
```

[Rendered](https://github.com/vuejs/rfcs/blob/sfc-improvements/active-rfcs/0000-sfc-style-variables.md)

PR #226, opened 2020-11-09
SFC state driven CSS variables (<style vars>)

yyx990803:
This is previously part of #182 - but moved into a separate PR so that it can advance on its own.

```html
<template>
  <div class="text">hello</div>
</template>

<script>
export default {
  data() {
    return {
      color: 'red'
    }
  }
}
</script>

<style vars="{ color }">
.text {
  color: var(--color);
}
</style>
```

[Rendered](https://github.com/vuejs/rfcs/blob/style-vars/active-rfcs/0000-sfc-style-variables.md)


yyx990803:
This RFC is now in final comments stage (prior discussion and feedback in #182).

An RFC in final comments stage means that:

Demivan:
> Note that when scoped and vars are both present, all CSS variables are considered to be local. In order to reference a global CSS variable here, use the global: prefix:
```html
<style scoped vars="{ color }">
h1 {
  color: var(--color);
  font-size: var(--global:fontSize);
}
</style>
```
What is the reason to prefix all variables if there is list of variables to bind? Why not prefix only ones that are in `vars` declaration?

privatenumber:
@Demivan 

Good question. Consider the case where a non-var'd CSS variable is _declared_ within the `<style>`. If it's not scoped, it will leak out into the global scope and affect external styles using the same var name:
```html
<style scoped vars="{ color }">
h1 {
  --fontSize: 12px;   /* ⬅ If this was not scoped, it will leak out to the global scope */

  color: var(--color);
  font-size: var(--fontSize);
}
</style>
```

Similarly, unused classnames in a SFC are also scoped because it might match against something outside of the SFC and cause a collision.

asv1:
```
<template>
  <div class="text">hello</div>
</template>

<script>
export default {
  data() {
    return {
      color: 'red'
    }
  }
}
</script>

<style> // vars are deduced from usage
.text {
  color: var(--color); // auto import color if it used in style
}
</style>
```

Why not auto import variables by their usage?

Demivan:
I don't think scoping declared variables should even be a part of this RFC. And in this example variable will not leak out into global scope because `h1` itself will be scoped. It could leak into child elements though.

My issue with current approach is that adding `scoped` attributed to existing `<style>` block requires adding `v-deep` only in rare cases. But adding `vars` with one variable will force to add `global:` prefix to all currently used global variables.
When developing site that already uses variables (for example for theming)  it will be annoying to use.

You can use global classes in `template` when using `scoped` style. I think it will be intuitive that you can use global variables in `style` with `vars` too.

My suggestion is to use `:global` prefix only for variables with conflicting names to allow access to global variable. And maybe have different `<style>` attribute for scoping all variables.

yyx990803:
@asv1 unlike JavaScript variables, CSS variables may be referencing something defined outside of the component. Without explicit variable passing the scoping will become ambiguous.

@Demivan the `var(--color)` to `var(--hash-color)` conversion is done at compile time. If we want to apply this only to passed variables, we will need to know what variables are passed at compile time. However, the `vars` expression can be any valid JavaScript expression: it may be a dynamic function call which isn't statically analyzable.

So the trade-off is either limit `vars` to a static identifier list (which makes it static analyzable), or requiring the global prefix.

Demivan:
@yyx990803 Thanks for explanation.
Another solution could be to prefix injected variables instead of global. I would rather prefix them and have my global ones same in vue files and in scss (css). And it will be easier to find injected css variables in a component.
```html
<style scoped vars="themeVars">
h1 {
  color: var(--bind:color); // or --var:color
  font-size: var(--fontSize);
}
</style>
```
But what is preferable probably depends on a project. If project does not use global variables automatic prefixing would be more convenient. Maybe option to switch this behavior could be added as `vue-loader` config?

asv1:
@yyx990803 
Understand. And one question, vars passed to CSS is reactive?

privatenumber:
@yyx990803 

If I'm understanding the problem correctly, I think hashing can be moved to run-time and eliminate the need for the `--global:` prefix if the styles gets compiled into a function where all CSS vars are left as placeholders.

Given the SFC:
```html
<style vars="{ textColor }">
.class {
    color: var(--text-color);
}
</style>
```

vue-loader compiles this to:
```js
function compileStyles(vars) {
    return `
    .class{
        color: var(--${vars.textColor}); // could be '--text-color' or '--HASH-text-color'
    }
    `;
}
```

There's probably some overhead I'm not thinking of though. I'm not sure if the theme-injection code-snippet reacts to every `theme-variables` change, but with this implementation, it will definitely have to and trigger a re-render (not that I can think of a use-case where the var names would be constantly changing).

Interesting use-case by the way. I didn't think `vars="themeVars"` would be possible. I thought the `vars="{ color }"` syntax was doing an object-spread on the view model. Or did you mean `vars="{ themeVars }"`?

If the following syntax can be supported `vars="somePropertyOnVM"`, it could also be an alternative to prefixing and marking which CSS vars are local:

```html
<style vars="somePropertyOnVM">
.class {
    color: var(--somePropertyOnVM.textColor);
    font-size: var(--someGlobalVar);
}
</style>
```

Supporting object paths could be beneficial as well since a theme objects can be nested for organizational purposes:
```json5
{
    colors: {
        primary: "red"
    }
}
````

```html
<style vars="themeVars">
.class {
    color: var(--themeVars.colors.primary);
}
</style>
```

yyx990803:
Thanks for the feedback - this actually made me rethink the proposal and identified a number of issues that can be improved.

A new version of this feature is proposed in #231.

PR #230, opened 2020-11-15
Synchronous time slicing

yisar:
This is a time slicing implement way suitable for Vue, which can effectively alleviate Jank.

It's very intrusive, the same as the current behavior, and it's coarse-grained and doesn't sacrifice the time of the microtask.

Thank you~

[rerendered](https://github.com/yisar/rfcs/blob/master/active-rfcs/0038-synchronous-time-slicing.md)

### DIFF
--- a/active-rfcs/0038-synchronous-time-slicing.md
+++ b/active-rfcs/0038-synchronous-time-slicing.md
@@ -54,7 +54,7 @@ const flush = function () {
     deferQueue.push(queue)
     queue.length = 0
   }
-  if (readQueue.length + queue.length + deferQueue.length > 0) {
+  if (queue.length + deferQueue.length > 0) {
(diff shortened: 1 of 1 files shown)


yyx990803:
I don't think your understanding of rAF is correct.

```js
Promise.resolve().then(() => {
  console.log('micro')
  requestAnimationFrame(() => {
    console.log('raf')
  })
  setTimeout(() => {
    console.log('timeout')
  }, 0)
}).then(() => {
  console.log('chained micro')
})
```

The output will be

- micro
- chained micro
- timeout
- raf

There's just no way rAF can be used for "synchronous" time slicing.

yyx990803:
Closing since I don't think this is actually feasible unless a prototype can prove otherwise.

PR #231, opened 2020-11-16
SFC style CSS variable injection (new edition)

yyx990803:
This is an improved alternative of the previous version (`<style vars>` as proposed in #226)

Thanks to feedback by @Demivan and @privatenumber, #226 has a few notable issues that can be improved:

1. Requires manual `vars` declaration for exposing variables that can be used.
2. No obvious visual hint that a variable is injected and reactive
3. Different behavior in scoped/non-scoped mode.
4. In non-scoped mode, the CSS variables would leak into child components.
5. In scoped mode, the `global:` prefix is required to use normal CSS variables declared outside of the component. It would be preferable that normal CSS variable usage stays the same in/out of the component.

This proposal addresses all of the above with the following usage:

```html
<template>
  <div class="text">hello</div>
</template>

<script>
  export default {
    data() {
      return {
        color: 'red',
        font: {
          size: '2em'
        }
      }
    }
</script>

<style>
  .text {
    color: v-bind(color);

    /* expressions (wrap in quotes) */
    font-size: v-bind('font.size');
  }
</style>
```

Summary:

- No need to explicit declare what properties are injected as CSS variables (inferred from usage of `v-bind()` in CSS)
- Reactive variables are more visibly different
- Same behavior in scoped/non-scoped mode
- No leakage into child components
- Usage of plain CSS variables are unaffected

[Rendered](https://github.com/vuejs/rfcs/blob/style-vars-2/active-rfcs/0000-sfc-style-variables.md)


CyberAP:
What do you think about writing those bindings without `--` prefix, so we can distinguish them better from the native CSS Custom Properties?

```html
<script>
export default {
  data() {
    return { baz: 'red' }
  }
}
</script>

<style>
.foo {
  color: var(:bar); // binding
  font-size: var(--global-font-size); // native property
}
</style>
```

We don't reference those transformed bindings in the code so they don't really require `--` prefix.

yyx990803:
@CyberAP I think it's better to keep the leading `--` because it's a smaller spec mismatch than without it:

1. `--:foo` can still be considered a custom property identifier that contains an invalid char
2. `:foo` is not a custom property identifier (which is defined as an identifier with leading `--`) - in fact, CSS identifier tokens cannot start with `:`.

I think (2) has a higher risk of tooling incompatibility as it's more likely for a parser to fail to tokenize `:foo` as an identifier due to the invalid leading char.

privatenumber:
- Could there be an obfuscate/mangle/minify CSS vars option to enable on production? I'm interested in the following benefits:
  - **Minification** The CSS distribution size can be reduced if hashed var names are mangled, especially for long & semantic variable names. eg. if `var(--:theme.color.secondary)` can be compiled to `var(--6b537420)` instead of `var(--6b53742-theme_color_secondary)`.
  - **Obfuscation** Like JS minification, mangled CSS property names can deter reverse engineers. I can't remember which CSS-in-JS library Instagram uses, but this is really nice:

    Screen Shot 2020-11-16 at 6 50 25 PM


- Wondering if there's a scalability issue. If there's a page that renders 100 different themeable components that all reference an injected theme property, I'm thinking that would create 100 different CSS vars that all map to the same state. If all components that read from the same state can share one CSS variable, the performance can improve dramatically. Have a few ideas to address this but would like to validate whether this is a real concern first.

```html
// Button.vue - example of a themable component reading from an injected state
<template>
	<button class="button">
		<slot />
	</button>
</template>

<script>
export default {
	inject: ['theme'],
};
</script>

<style>
.button {
	background-color: var(--:theme.color.primary);
}
</style>
```

```html
// Input.vue - another example of a themable component reading from an injected state
<template>
	<input
        class="input"
        type="text"
    >
</template>

<script>
export default {
	inject: ['theme'],
};
</script>

<style>
.input {
	color: var(--:theme.color.primary);
}
</style>
```

```html
// Layout.vue - injects theme state, and also has UI that rapidly updates theme properties
<template>
	<div class="layout">
		<label>
			Primary color:

			<!-- Every change here will update a different CSS var for each component that reads from this  -->
			<input
				type="color"
				v-model="theme.color.primary"
			>
		</label>

		<slot />
	</div>
</template>

<script>
export default {
	data() {
		return {
			theme: {
				color: {
					primary: 'red',
				},
			},
		};
	},

	provide() {
		return {
			theme: this.theme,
		};
	},
};
</script>
```

xstxhjh:
font-size: var(--:font.size);
`:` `.`  
vscode problems: ) expectedcss(css-rparentexpected)

Invalid CSS Var
```css
:root {
  --font.size: 22px;
}

.text {
  font-size: var(--font.size);
}
```

valid CSS Var
```css
:root {
  --fontSize: 22px;
}

.text {
  font-size: var(--fontSize);
}
```

web2033:
What's the reason of having both long and shorthand notation? Just pick one...

Kocal:
@web2033 We have `v-bind:attr` and `:attr` in templates, so it makes sense

CyberAP:
From my personal experience `v-bind` is never used in projects in a long notation and having two types of notation has never been great because people have to choose one when framework should actually solve that problem for them.

web2033:
How about
```css
.text {
    color: var(--v-bind-color);
    font-size: var(--v-bind-font-size);
}
```
since a finger is already on `-` anyways. 
Valid CSS. No tooling tweaking needed.

web2033:
... or "vue-branded" vars 😁
```html
<template>
  <div class="text">hello</div>
</template>

<script>
  export default {
    data() {
      return {
        color: 'red',
        font: {
          size: '2em'
        }
      }
    }
</script>

<style>
.text {
    color: var(--vue-color);
    font-size: var(--vue-font-size);
}
</style>
```

yyx990803:
@privatenumber the theme injection is an example, and yes it can potentially lead to inefficiency when used everywhere. Maybe for global themes it is still better to provide it as plain CSS variables at root and let is cascade:

```html
<!-- App.vue -->
<script>
const theme = ref({
  color: 'red',
  // ...
})
</script>

<template>
  <div class="theme-provider>
    <input type="color" v-model="theme.color">
  </div>
</template>

<style>
.theme-provider {
  --color-primary: var(--:theme.color);
}
</style>
```

yyx990803:
@xstxhjh ok so I tested this within an SFC with an experimental extension that did not have strict CSS linting... turns out having `:` and `.` inside a variable name does create issues.

So maybe instead we can just use a custom CSS function instead of `var()`:

```css
.text {
  font-size: v-bind(color);
}
```

In order to use expressions that contain invalid CSS indentifier chars inside, wrap it with quotes:

```css
.text {
  font-size: v-bind('theme.color.primary');
}
```

yyx990803:
Revised to use `v-bind()` function instead - this should be valid CSS syntax that is compatible with any tooling.

CyberAP:
This is much better now!

replygirl:
When I was using the first version of `<style scoped vars>` I saw it as a powerful way to extend a Stylus theming system I've been using across most of my projects that uses classes to apply a bunch of CSS variables at different scopes. I wasn't able to do this in reality, because of the scoping applied to the variable names themselves (or maybe I could have using a second, non-scoped stylesheet?)

This proposal makes variable injection perfect for a partially CSS-in-JS type of theming approach [like you showed](https://github.com/vuejs/rfcs/pull/231#issuecomment-728975015), but the function approach would make directly overriding CSS variables on a case-by-case basis a bit unwieldy with `v-bind` acting as a middleman. Maybe this is better, though, because it encourages you to commit to a consistent paradigm for the most part

Recycling `v-bind` feels weird to me, idk. @web2033's [prefixing solution](https://github.com/vuejs/rfcs/pull/231#issuecomment-728877771) makes a little more sense to me—There's semantic parity with how `cli` handles environment variables, and the downside of the function approach I mentioned above goes away

1607611087:
it seems that i could‘t compress all css file？it seems that `<style vars>` generate css by js because i see it use variables from `data` return object in the example

yyx990803:
@1607611087 the generated CSS file is static and compatible for compression. It generates rules using hashed CSS variables which are injected at runtime onto DOM nodes as inline styles by Vue.

1607611087:
@yyx990803 inline style? emmm... it looks like complex for developer if there are many many many inline style on a node.

why not has two way like `<style vars>` and `<style>` just like `<script setup>` and `<script>`?

`<style>` is same as before
`<style vars>` can use variables from `data` or `setup` return object, and generate a CSS `class` with a independent attribute (like `data-v-xxxx`)  and a original attribute, and then dynamically insert it wrapped `<style>` to DOM

```html
<style>
.text {
  /* static style */
}
</style>

<style vars>
.text {
  color: v-bind(color);
  font-size: v-bind('font.size');
}
</style>
```

after generate

```html
<style>
/* other static css style*/

.text[data-v-xxxxabcd]{
  /* static style */
}
</style>

<style>
/* dynamically generate */
.text[data-v-xxxxabcd][data-v-xxxxefgh]{
  color: red;
  font-size: 14px;
}
</style>
```

Justineo:
@1607611087 Dynamically generating a `<style>` element will be much more expensive than precisely changing values of related CSS custom properties. See the benchmark [here](https://jsbench.me/9zkhq0a6ju/1).

1607611087:
sorry,i dont consider this problem😂. Dynamically generating a `<style>` element exactly has a low performance.

yyx990803:

b-strauss:
Will this work with css custom properties?

```css
.foo {
  --main-color: v-bind(color);
}
```

yyx990803:
@b-strauss yes it works.

The syntax is now usable with `vue@^3.0.3` + `vite@^1.0.0-rc.13` or `vue-loader@next`.

tavrez:
Can this be used in this form?
```css
p:nth-child(v-bind(child)) {
  background: red;
}
```

leopiccionia:
@tavrez I'm not sure if this is a syntactically-valid placement for a CSS custom property (a.k.a. variable), so I'd guess not, at least for now.

talmand:
I may be late to the party here, but I have to say as a UI dev I was more interested in the previous version of this concept. I would suggest that this version goes against the nature of CSS variables, removes some of their usefulness, and prevents using the fallback nature of the second parameter of the var() function in CSS.

CSS variables are supposed to "leak" into the child elements (or components as you will) as they follow the cascading part of CSS. Attempting to prevent this feature actually makes using CSS vars more complicated. If I wanted to keep the cascading feature I would now have to pass the data along to children components instead of relying on the cascade, which is far more efficient. Even then, the CSS variables of the parent component are inherited by the child component (you can see them in the devtools) but they can't be used. If the same data is passed down for the child component to create its own CSS variables there will be multiple CSS variables with the same value available to the child component. That would mean that grandchildren or even great-grandchildren will have who knows how many references to CSS variables being inherited that cannot be used; which mucks up the devtools display.

In the previous version of this concept I had several tests built trying different things. I was using props, data object variables, and setup function returning variables. Each of these worked very well with the vars attribute as I was just feeding it an object that I built however I felt. Then just referenced the variables inside the var() CSS function. It could be how I have my first tests of this version working right now, but I can only seem to get it to work with data object variables in the Options API.

Props usage spits out this error:

`Uncaught (in promise) ReferenceError: __props is not defined`

Trying to use the Composition API I get this error:

`Uncaught (in promise) ReferenceError: _unref is not defined`

Again, this could be my fault somehow as I just started tinkering with this version. I'm just trying to pass along the same object in each of those cases. I would try to create an example on CodePen but for some reason they still have the compiler that expects the vars attribute version.

In the current version the v-bind has taken away the var() function altogether and to get it back requires creating yet more CSS variables that take in the reactive-based CSS variables to provide expected functionality such as the fallback:

```
--myColor: v-bind('myReactiveColor');
color: var(--myColor, black);
```

As someone who is currently working on a design system that has heavy usage of CSS variables, I can't see how this feature helps me in that endeavor. The previous version easily allowed for global CSS variables to be overridden at the component level with just props if needed, a component reacting to changes to use existing CSS variables to update styles, and have those changes cascade downward; all with very little setup or overhead. I can see this ability being very useful, the current version is far more restrictive in my opinion.

talmand:
After pondering over this I have two suggestions that could alleviate my concerns above.

How about an optional property for the component much like `inheritAttrs`? This property, named something like `prefixCssVars`, would default to true and the CSS variables are created with their prefixes as currently working. On the other hand, if the property is set to false in the component then the CSS variables are created without prefixes allowing for the cascading feature. I can understand the desire to scope CSS variables just to the component, but I would think the option to leverage a basic feature of CSS variables would also be desirable.

Maybe as a parameter of the v-bind so the prefix can be controlled on a per CSS variable basis?

Since v-bind is a function inside the style block would it be possible to allow for another parameter, that expects a string, that is placed as the fallback in the created CSS var function? For instance, `color: v-bind('myColor', 'black');` would be outputted as `color: var(--myColor, black);`. This would even potentially allow for something like `color: v-bind('myColor', 'var(--myDefaultColor)');`, which would allow for using CSS variables as the fallback. Granted this feature's usefulness depends on the ability to cancel out the variable in the v-bind so that the fallback would be used. In the vars attribute version I was doing this by assigning `null` as the value to be passed along which prevented the CSS variable from being created.

yyx990803:
@talmand 

First of all: **this feature is not about enhancing CSS variables**. It's about injecting state-driven values into your CSS. CSS variables just happen to be the underlying implementation detail.

With that in mind, it doesn't make sense for the bindings to "leak". If you want them to leak, just bind them to an actual CSS variable that "leaks":

```css
.foo {
  --myColor: v-bind(myReactiveColor);
}
```

And if you want a fallback, note that `v-bind` can actually accept JavaScript expressions:

```css
.foo {
  --myColor: v-bind('myReactiveColor || "black"');
}
```

CyberAP:
I'd really wish for CSS and JS code to be separated from each other, i.e. for v-bind to be able to accept only a context reference without proper support for full js expressions.

```css
.foo {
  color: v-bind(bar); /* valid */
  color: v-bind('baz.qux'); /* valid */
  color: v-bind('1+1'); /* invalid */
}
```

It's not a pleasant experience if your logic is split up from the main component's code (where you actually can place exactly the same expression in a form of computed).

talmand:
I didn't really see this as a means to enhance CSS variables, but exactly as you say; to leverage a useful feature of CSS that allows for dynamic changes to the styles without requiring a compiling of files. This is an excellent use of the CSS feature to enhance features of Vue. This kind of thing was already being done via injecting style attributes on elements in the component; this feature takes away the overhead of doing that.

The only real disagreement I have is to permanently prevent an established behavior and feature of CSS variables. Cascading is part of the spec and I would wonder why there shouldn't even be an option to allow it if the dev wishes the cascade to happen. That is, an option that doesn't require overhead in the style block to handle it.

In the end, we'll learn to leverage the feature as we need in ways we are unable to foresee at this time.

risha700:
> This is an improved alternative of the previous version (`<style vars>` as proposed in #226)
> 
> Thanks to feedback by @Demivan and @privatenumber, #226 has a few notable issues that can be improved:
> 
> 1. Requires manual `vars` declaration for exposing variables that can be used.
> 2. No obvious visual hint that a variable is injected and reactive
> 3. Different behavior in scoped/non-scoped mode.
> 4. In non-scoped mode, the CSS variables would leak into child components.
> 5. In scoped mode, the `global:` prefix is required to use normal CSS variables declared outside of the component. It would be preferable that normal CSS variable usage stays the same in/out of the component.
> 
> This proposal addresses all of the above with the following usage:
> 
> ```
> <template>
>   <div class="text">hello</div>
> </template>
> 
> <script>
>   export default {
>     data() {
>       return {
>         color: 'red',
>         font: {
>           size: '2em'
>         }
>       }
>     }
> </script>
> 
> <style>
>   .text {
>     color: v-bind(color);
> 
>     /* expressions (wrap in quotes) */
>     font-size: v-bind('font.size');
>   }
> </style>
> ```
> 
> Summary:
> 
> * No need to explicit declare what properties are injected as CSS variables (inferred from usage of `v-bind()` in CSS)
> * Reactive variables are more visibly different
> * Same behavior in scoped/non-scoped mode
> * No leakage into child components
> * Usage of plain CSS variables are unaffected
> 
> [Rendered](https://github.com/vuejs/rfcs/blob/style-vars-2/active-rfcs/0000-sfc-style-variables.md)

Never worked for me ! vue@3.0.3
Uncaught (in promise) ReferenceError: _unref is not defined

AlexeyZelenko:
> ```
> <div class="theme-provider">
> ```

AlexeyZelenko:
```
<template>
<h5   class="card-title">
	Roman Kravchuk
#</h5>
<div class="theme-provider">
    	<input type="color" v-model="theme.color">
</div>
<template/>
<script>
import {ref, defineComponent, computed } from 'vue'
    name: "About",
    setup() {
        const theme = ref({
            color: 'red'
        })
			const myColor = ref( '#2c790d')
			const colorBind = computed(() => myColor.value)
			return {
					myColor,
          colorBind,
          theme
			}
    },
    data() {
        return {
            color: 'black',
            font: {
                size: '1rem'
            }
        }
    }
})
</script>

<style lang="scss">
	.card-title {
		color: v-bind('theme.color');
	}
	.card-text {
		color: v-bind(color);

		/* expressions (wrap in quotes) */
		font-size: v-bind('font.size');
	}
	.theme-provider {
		--color-primary: var(--:theme.color);
	}

</style>
```

Matt-Deacalion:
Does this work with `ref`? I haven't been able to make it work in Vue `3.0.0`, but I can get it working if I use `reactive`.
```vue
<template>
    <h1>Green text!</h1>
</template>

<script>
import { reactive } from 'vue';

export default {
    name: 'App',

    setup() {
        // …this works
        return { ...reactive({ colour: 'green' }) };

        // …but this does not
        // return { colour: ref('green') };
    },
}
</script>

<style>
h1 {
    color: v-bind('colour');
}
</style>
```

AndrewBogdanovTSS:
@Matt-Deacalion works with refs for me, using it with Vue 3.0.5. I also tried it with setup attribute and seems like it also works fine:
``` js
<script setup>
  const testColor = 'blue'
</script>
<style lang="postcss">
  .test {
    color: v-bind(testColor);
  }
</style>
```

Matt-Deacalion:
@AndrewBogdanovTSS Still no luck with `Vue 3.0.5`. Interestingly it does work if I use the new `<script setup>` syntax.

cxdev:
The syntax is cool ! Can I use it with `teleport` ?
 it don't work: 
```vue
<script lang="ts">
import { defineComponent, inject } from 'vue'
import { Viewer } from 'cesium'
  name: 'ViewerPopup',
  inheritAttrs: false,
  props: {
    top: {
      type: String,
      required: false
    },
    left: {
      type: String,
      required: false
    },
    show: {
      type: Boolean,
      default: true
    }
  },
  setup () {
    const viewer = inject<Viewer>('viewer')
    const queryWidget = viewer.cesiumWidget.container // dom容器

    return {
      queryWidget
    }
  }
})
</script>

<template>
  <teleport :to="queryWidget">
    <div
      v-show="show"
      class="vum-popup-x"
      v-bind="$attrs"
    >
      <slot></slot>
    </div>
  </teleport>
</template>
<style>
.vum-popup-x {
  position: absolute;
  z-index: 1;
  top: v-bind(top);
  left: v-bind(left);
}
</style>
```

cxdev:
The syntax is cool ! Can I use it with `teleport` ?
 it don't work: 
```vue
<script lang="ts">
import { defineComponent, inject } from 'vue'
import { Viewer } from 'cesium'
  name: 'ViewerPopup',
  inheritAttrs: false,
  props: {
    top: {
      type: String,
      required: false
    },
    left: {
      type: String,
      required: false
    },
    show: {
      type: Boolean,
      default: true
    }
  },
  setup () {
    const viewer = inject<Viewer>('viewer')
    const queryWidget = viewer.cesiumWidget.container // dom容器

    return {
      queryWidget
    }
  }
})
</script>

<template>
  <teleport :to="queryWidget">
    <div
      v-show="show"
      class="vum-popup-x"
      v-bind="$attrs"
    >
      <slot></slot>
    </div>
  </teleport>
</template>
<style>
.vum-popup-x {
  position: absolute;
  z-index: 1;
  top: v-bind(top);
  left: v-bind(left);
}
</style>
```

AndrewBogdanovTSS:
@cxdev why should it work with teleport feature? If you investigate how this feature works under the hood you will see that it is placing a specific identifier on the parent node and applies css of the child nodes via it. Teleport moves nodes under completely different parent - so it won't match a selector anymore. If you want to use teleport and still have access to this feature you should create a wrapper component specifically for teleporting

cxdev:
@Andrew
BogdanovTSS I see, Thanks!   I just want to move the componet in a more appropriate (like a `button` in `widget-cotainer`) container which preset by a third party. Is there a better way to do this?
 in addition, I tried `rollup` a  `.vue` file which use the syntax in style, `rollup.config`:
```js
import { nodeResolve } from '@rollup/plugin-node-resolve'
import path from 'path'
import vuePlugin from 'rollup-plugin-vue'
import css from 'rollup-plugin-css-only'
const { entry } = require('./common')

export default [
  {
    input: path.resolve(__dirname, `../packages/${entry}/index.ts`),
    output: {
      format: 'es',
      file: 'lib/index.esm.js'
    },
    plugins: [
      nodeResolve(),
      css({
        output: 'index.esm.css'
      }),
      vuePlugin({
        css: false
      })
      // ....
    ],
    external: ['vue']
  }
]
```
output index.esm.css
```css
```

lanthaler:
> [...]
> And if you want a fallback, note that `v-bind` can actually accept JavaScript expressions:
> 
> ```css
> .foo {
>   --myColor: v-bind('myReactiveColor || "black"');
> }
> ```

@yyx990803 I wonder whether it would be worthwhile to add an optional second parameter to `v-bind` to simplify specifying the unit without having to resort to string concatenation if the reactive variable is a number:

```css
.foo {
  --mySize: v-bind(myReactiveSize, "px");
}
```

instead of

```css
.foo {
  --mySize: v-bind('myReactiveSize + "px"');
}
```

brian-sea:
v-bind seems to ignore props.  For example, in `<board :rows=8 :cols=8>`, rows and cols are completely ignored. My workaround at the moment is to copy them into (differently named) computed data.

Is this expected behavior or am I doing something wrong?

--- SCRIPT ---
```
export default {
    name: 'CheckerBoard',
    props: {
        rows: Number,
        cols: Number
    },
    data() {
    },
    computed: {
        cRows(){
            return this.rows;
        },
        cCols() {
            return this.cols;
        }
    }
}
```
--- STYLE ----
```
div.board {
        --num-rows: v-bind(cRows);
        --num-cols: v-bind(cCols);
}
```

petterw03:
Is there a reason for this functionality not working with props? That would be my main use case - sending for example color props down to a component and then use the color to dynamically change the style.

brian-sea:
Sometimes, I can access props... but it's not consistent.

daiwanxing:
Does anyone knows why writing like this doesn't work ? 
I get the error : **_unref is not defined**
Thanks..
```
<script>
export default {
    setup () {
        return {
            color: "red"
        }
    }
}
</script>
```
```
<style scoped>
h1 {
    color: v-bind(color);
}
</style>
```

mattersj:
Well, as I understand it works only with `data` property, buuut.. let's be honest, who cares about old-fashioned Options API in Vue 3? Why wouldn't use `props` and `setup` as well?
Personally, I've used `data` in Vue 3 about 0 times and something tells me that it's not just me.
I'd really like to see some obvious improvements in the current implementation.

AndrewBogdanovTSS:
> Well, as I understand it works only with `data` property, buuut.. let's be honest, who cares about old-fashioned Options API in Vue 3? Why wouldn't use `props` and `setup` as well?
> Personally, I've used `data` in Vue 3 about 0 times and something tells me that it's not just me.
> I'd really like to see some obvious improvements in the current implementation.

using data property is not "old-fashioned" in any way. If you don't use it - it doesn't mean no one else doesn't. I don't understand why you thought it's working only with data in the first place. I used it with "new" setup syntax and it worked perfectly fine

mattersj:
>using data property is not "old-fashioned" in any way. If you don't use it - it doesn't mean no one else doesn't.

I didn't say no one uses Options API in Vue 3, but I'm pretty sure that most people prefer to use the Composition API if they already picked the Vue 3.

>I used it with "new" setup syntax and it worked perfectly fine

Could you please provide the reproduction example? Here is mine: https://github.com/mattersj/vue-css-vars
Neither `props` nor `setup` does work, as I said before.

In my example **HW** component (with props) throws the following error: `Uncaught (in promise) ReferenceError: __props is not defined`
Whereas the second **HW2** component (with setup) throws this one: `Uncaught (in promise) ReferenceError: _unref is not defined`.

nesteiner:
在style中绑定变量后，给这个变量设置`setInterval`实时更新，但是样式没有变化诶  
比如这里修改 `opacity` 属性
```vue
<template>
  <div class="shining">
    <h1> Vue </h1>
  </div>
</template>

<script lang="ts">
 import {Vue} from 'vue-property-decorator'
 
 export default class Shining extends Vue {
   private opacity: number = 0

   mounted(): void {
     setInterval(() => {
       this.opacity >= 1 && (this.opacity = 0)
       this.opacity += 0.2
     }, 300)
   }
 }
</script>

<style scoped>
 h1 {
  /* color: rgb(65, 284, 131); */
   color: black;
   opacity: v-bind(opacity);
 }
</style>
```

chenxinan:
how i can use `v-bind` in `background-image: url()`
```vue
<template>
  <div class="bg" />
</template>

<script setup>
const img = "https://mdn.mozillademos.org/files/6457/mdn_logo_only_color.png";
</script>

<style scoped>
 .bg {
   --img: url(v-bind(img));
   --img2: url("https://mdn.mozillademos.org/files/6457/mdn_logo_only_color.png");
   width: 100vw;
   height: 100vh;
   background-image: url(v-bind(img)); /* no work */
   background-image: var(--img); /* no work */
   background-image: var(--img2); /* work */
 }
</style>
```

daiwanxing:
> how i can use `v-bind` in `background-image: url()`
> 
> ```
> <template>
>   <div class="bg" />
> </template>
> 
> <script setup>
> const img = "https://mdn.mozillademos.org/files/6457/mdn_logo_only_color.png";
> </script>
> 
> <style scoped>
>  .bg {
>    --img: url(v-bind(img));
>    --img2: url("https://mdn.mozillademos.org/files/6457/mdn_logo_only_color.png");
>    width: 100vw;
>    height: 100vh;
>    background-image: url(v-bind(img)); /* no work */
>    background-image: var(--img); /* no work */
>    background-image: var(--img2); /* work */
>  }
> </style>
> ```

Just write like this,  It's wrting similar to the css var function.

<script setup>
const img = "url(https://mdn.mozillademos.org/files/6457/mdn_logo_only_color.png)";
</script>


<style scoped>
 .bg {
   background-image: v-bind(img)
 }
</style>

nesteiner:
@daiwanxing 试试这个
```vue
<script>
import img from 'path'
export default {
    data() {
        img,
    }
}

</script>
```

0x1af2aec8f957:
> 在style中绑定变量后，给这个变量设置`setInterval`实时更新，但是样式没有变化诶
> 比如这里修改 `opacity` 属性
> 
> ```
> <template>
>   <div class="shining">
>     <h1> Vue </h1>
>   </div>
> </template>
> 
> <script lang="ts">
>  import {Vue} from 'vue-property-decorator'
>  
>  export default class Shining extends Vue {
>    private opacity: number = 0
> 
>    mounted(): void {
>      setInterval(() => {
>        this.opacity >= 1 && (this.opacity = 0)
>        this.opacity += 0.2
>      }, 300)
>    }
>  }
> </script>
> 
> <style scoped>
>  h1 {
>   /* color: rgb(65, 284, 131); */
>    color: black;
>    opacity: v-bind(opacity);
>  }
> </style>
> ```
> 
> 但是如果使用js写的话，没有这个问题
> 
> ```
>  export default {
>    data() {
>      return {
>        opacity: 0
>      }
>    },
> 
>    mounted() {
>      setInterval(() => {
>        this.opacity >= 1 && (this.opacity = 0)
>        this.opacity += 0.2
>      }, 300)
>    }
>  }
> ```

@nesteiner doc: https://github.com/vuejs/rfcs/blob/sfc-improvements/active-rfcs/0000-sfc-style-variables.md#motivation

edisdev:
Hi 🖐 I did make a small example this feature ☺️ 👇

https://gist.github.com/edisdev/c53d0959db41f94e40e84ee3ae42a928

edisdev:
> @privatenumber the theme injection is an example, and yes it can potentially lead to inefficiency when used everywhere. Maybe for global themes it is still better to provide it as plain CSS variables at root and let it cascade:
> 
> ```
> 
> <script>
> const theme = ref({
>   color: 'red',
>   // ...
> })
> </script>
> 
> <template>
>   <div class="theme-provider>
>     <input type="color" v-model="theme.color">
>   </div>
> </template>
> 
> <style>
> .theme-provider {
>   --color-primary: var(--:theme.color);
> }
> </style>
> ```

@yyx990803  Hi 🖐 i did make a small example for latest version this feature. (with v-bind) <a href="https://gist.github.com/edisdev/c53d0959db41f94e40e84ee3ae42a928" target="_blank">👉 this is gist file link<a/>


```vue
<template>
  <div class="Example">
    <div class="area">
      <div class="form">
        <label>Select Color</label>
        <input type="color" v-model="customTheme.bgColor" />
      </div>
      <div class="preview">
        <span>{{ customTheme.bgColor }}</span>
        <div class="customColor"></div>
      </div>
    </div>
  </div>
</template>

<script>
import { ref } from "vue";
export default {
   setup() {
    const customTheme = ref({
      bgColor: "",
    });
    return {
      customTheme
    };
  },
};
</script>

<style>
.Example {
  --custom-color: v-bind("customTheme.bgColor");
}
.customColor {
  background: var(--custom-color);
  width: 100px;
  height: 50px;
}
</style>
```

edisdev:
> @privatenumber the theme injection is an example, and yes it can potentially lead to inefficiency when used everywhere. Maybe for global themes it is still better to provide it as plain CSS variables at root and let it cascade:
> 
> ```
> 
> <script>
> const theme = ref({
>   color: 'red',
>   // ...
> })
> </script>
> 
> <template>
>   <div class="theme-provider>
>     <input type="color" v-model="theme.color">
>   </div>
> </template>
> 
> <style>
> .theme-provider {
>   --color-primary: var(--:theme.color);
> }
> </style>
> ```

 Hi  @yyx990803 🖐 i did make a small example for latest version this feature. (with v-bind) 👉 [The gist link is also here.](https://gist.github.com/edisdev/c53d0959db41f94e40e84ee3ae42a928)

I hope that will be useful. 🙏

```vue
<template>
  <div class="Example">
    <div class="area">
      <div class="form">
        <label>Select Color</label>
        <input type="color" v-model="customTheme.bgColor" />
      </div>
      <div class="preview">
        <span>{{ customTheme.bgColor }}</span>
        <div class="customColor"></div>
      </div>
    </div>
  </div>
</template>

<script>
import { ref } from "vue";
export default {
   setup() {
    const customTheme = ref({
      bgColor: "",
    });
    return {
      customTheme
    };
  },
};
</script>

<style>
.Example {
  --custom-color: v-bind("customTheme.bgColor");
}
.customColor {
  background: var(--custom-color);
  width: 100px;
  height: 50px;
}
</style>

```

m4heshd:
Amazing. Is SCSS compatible with this feature?

ITenthusiasm:
@m4heshd I think so. When I tried to test out the experimental feature while working on a Vue 3 project, it seemed to work fine while using `scss` for the `style` tag.

edisdev:
> Amazing. Is SCSS compatible with this feature?

@m4heshd I tried this. This feature worked with SCSS. But I couldn't get this feature to work in production mode. So when I run vue cli server build it didn't work. What do you think about that? @yyx990803 

https://user-images.githubusercontent.com/21293903/124126368-49e13900-da83-11eb-9dc2-94ec2a69e33e.mov


```vue
<template>
  <div class="Example">
    <div class="area">
      <div class="form">
        <label>Select Color (SCSS)</label>
        <input type="color" v-model="themeColors.bgColor" />
      </div>
      <div class="preview">
        <span>{{ themeColors.bgColor }}</span>
        <div class="customColor"></div>
      </div>
    </div>
  </div>
</template>

<script>
import { ref } from "vue";

export default {
   setup() {
    const themeColors = ref({
      bgColor: ''
    })

    return {
      themeColors
    };
  },
};
</script>

<style lang="scss">

.Example {
  $customColor: v-bind('themeColors.bgColor');

  .customColor {
    background: $customColor;
    width: 100px;
    height: 50px;
  }
}

</style>
```

m4heshd:
@ITenthusiasm good to know that.

@edisdev Ouch. 😬 That's what I was afraid of. I really don't wanna be hit with any "gotchas" when going into production. Did you use `dart-sass` or `node-sass`?

edisdev:
@m4heshd Yes, project dependencies has node-sass and sass-loader. Actually this feature is not working also with pure css in production mode. While reviewing this, I found that the style variable name that binds to the element is incorrect in production mode.

For Example : 
Screen Shot 2021-07-01 at 17 39 56


 `--7ba5bd90-themeColors_bgColor`  is not correct. I guess that's the variable name prepared for development mode.

edisdev:
@m4heshd  I hope it will be fixed soon. I opened issue about this.

GuoChen-thlg:
When I tried to use it in a development environment, it worked fine

```
:root{
     --47536436-theLen: 200px;
    --47536436-angle: 45deg;
    --47536436-offset: 100px;
}
.style{
    position: relative;
    list-style-type: none;
    width: var(--47536436-theLen);
    height: var(--47536436-theLen);
    transform: rotate3d(0.7, 0.5, 0.5, var(--47536436-angle));
    transform-style: preserve-3d;
}
````

But when I packed it, it didn't work

```
:root{
   --47536436-theLen: 200px;
   --47536436-angle: 45deg;
   --47536436-offset: 100px;
}
.style{
    position: relative;
    list-style-type: none;
    width: var(--946a2296);
    height: var(--946a2296);
    transform: rotate3d(.7,.5,.5,var(--8afc983c));
    transform-style: preserve-3d;
}
```

**Comment by GulnazKhasanova** [comment]
Tell me how to dynamically change a property  background-image in ::before?

**Comment by yyx990803** [comment]
Looks like most of the issues raised after the final comments period were about implementation bugs - most of which have been addressed in the latest version (3.1.5).

- One of the issues where the feature does not work in production mode for `vue-cli` is due to an issue of `thread-loader` and is tracked in https://github.com/vuejs/vue-next/issues/3921

- If there are other cases where it's not working as specified in the RFC, please open an issue at the `vue-next` repo with reproduction instead of commenting here.

Since these are all implementation issues, the RFC itself is considered finalized. We will try to resolve all implementation bugs in the upcoming 3.2 and this feature will be out of experimental status when 3.2 stable is released.

**Comment by LiuQixuan** [comment]
How to prevent the rendered css variable with hash prefixed? The same component will render multiple css files, which is obviously unreasonable. I am not afraid of same css variable name.In my mind,the css variable was written in the HTML element and only works for the css var( ) syntax under the HTML element. Why do you need to add a hash prefix? Just to avoid duplication?

**Comment by iDerekLi** [comment]
# BUG: in SCSS

## System Info
 ```
  OS: Windows 11
    CPU: (8) x64 AMD Ryzen 5 3500U with Radeon Vega Mobile Gfx
    Memory: 535.72 MB / 6.94 GB
  Binaries:
    Node: 12.15.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.5 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.13.4 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22000.1.0), Chromium (91.0.864.67)
    Internet Explorer: 11.0.22000.1
  npmPackages:
    "sass": "^1.26.5",
    "sass-loader": "^8.0.2"
  vueCLI:
     @vue/cli 4.5.13
```

## CSS
css parsed variables are normal.

```vue
<script setup>
const width = 123;
</script>
<style scoped>
.canvas {
  position: relative;
  width: v-bind(width + "px");
  box-shadow: 0 0 3px 1px rgba(0, 0, 0, 0.2);
  background: #ffffff;
}
</style>
```
image

## SCSS
scss parsing variables are not equal.

```vue
<script setup>
const width = 123;
</script>

<style lang="scss" scoped>
.canvas {
  position: relative;
  width: v-bind(width + "px");
  box-shadow: 0 0 3px 1px rgba(0, 0, 0, 0.2);
  background: #ffffff;
}
</style>
```

image

berzi:
`v-bind()` doesn't seem to work inside CSS rules with the `::v-global()` selector.

Is this the right place to report this? I got here through a warning in the compiler.

eyun-tv:
v-bind() doesn't seem to work 
i use pug & coffeescript and stylus

image

image

privatenumber:
@iDerekLi @eyun-tv 

In both of your examples, you're using preprocessors that manipulate the source code before passing it to Vue. That means Stylus or SCSS is likely changing the JS expression you pass inside `v-bind`. In the RFC, it says expressions should be wrapped in quotes, and that will probably prevent the preprocessors from interfering too: `v-bind('vol + "%"')`

FYI, I don't think this is the right place to make bug reports or seek help, so ideally you submit an issue instead of responding here.

eyun-tv:
image
I don't think that's the problem you're talking about
I made the changes as you said, but it still doesn't work
I see that the variable is generated in the generated css, but there is no place to initialize this variable @privatenumber

andrei-gheorghiu:
@eyun-tv, according to [this post  above](#issuecomment-881651576), you should not report implementation bugs on this thread. Please create a separate issue on the vue repo. 

Also note, starting tomorrow, vue will [default to version 3](https://blog.vuejs.org/posts/vue-3-as-the-new-default.html). You might want to wait until the move is completed and post on the appropriate renamed repo.

L1yp:
> @eyun-tv, according to [this post above](#issuecomment-881651576), you should not report implementation bugs on this thread. Please create a separate issue on the vue repo.
> 
> Also note, starting tomorrow, vue will [default to version 3](https://blog.vuejs.org/posts/vue-3-as-the-new-default.html). You might want to wait until the move is completed and post on the appropriate renamed repo.

same problem, I guess it was caused by teleported.

L1yp:
> 

[repo url](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtYnV0dG9uIEBjbGljaz1cImRpYWxvZ1Zpc2libGUgPSB0cnVlXCI+T3BlbjwvZWwtYnV0dG9uPlxuXG4gIDxlbC1kaWFsb2dcbiAgICB2LW1vZGVsPVwiZGlhbG9nVmlzaWJsZVwiXG4gICAgdGl0bGU9XCJUaXBzXCJcbiAgICB3aWR0aD1cIjMwJVwiXG4gICAgOmJlZm9yZS1jbG9zZT1cImhhbmRsZUNsb3NlXCJcbiAgPlxuICAgIDxzcGFuIGNsYXNzPVwibWVzc2FnZVwiPlRoaXMgaXMgYSBtZXNzYWdlPC9zcGFuPlxuICA8L2VsLWRpYWxvZz5cbjwvdGVtcGxhdGU+XG5cbjxzY3JpcHQgbGFuZz1cInRzXCIgc2V0dXA+XG5pbXBvcnQgeyByZWYgfSBmcm9tICd2dWUnXG5cbmNvbnN0IGRpYWxvZ1Zpc2libGUgPSByZWYoZmFsc2UpXG5cbiAgY29uc3QgYmdDb2xvciA9ICdibHVlJ1xuPC9zY3JpcHQ+XG48c3R5bGUgc2NvcGVkPlxuICAubWVzc2FnZSB7XG4gICAgYmFja2dyb3VuZC1jb2xvcjogdi1iaW5kKGJnQ29sb3IpXG4gIH1cbjwvc3R5bGU+XG4iLCJpbXBvcnRfbWFwLmpzb24iOiJ7XG4gIFwiaW1wb3J0c1wiOiB7fVxufSIsIl9vIjp7fX0=)

deepthan:
贴个方法，找了好久没找到试出来了，希望可以帮助到你。
css 的 background: src(xxx) 使用变量，在 js 中拼接好 url

```
<script setup lang="ts">
const imgUrl = ref(
  'url(https://p3-passport.byteimg.com/img/mosaic-legacy/3795/3047680722~180x180.awebp)'
)
</script>

<style lang="less" scoped>
.compass {
  background: v-bind(imgUrl);
}
</style>
```

deepthan:
How to bind a variable of @font-face src ? The following does not work.
```
<script setup lang="ts">
const fontFamily = ref(
  'url(http://www.xxx.comxxx.ttf)'
)
</script>

<style lang="less">
.@font-face {
  font-family: 'YouSheBiaoTiHei';
  src: v-bind(fontFamily);
}
 #</style>

```

PR #235, opened 2020-11-18
sfc-template-name-prop

xxbld:
[rendered](https://github.com/xxbld/rfcs/blob/sfc-template-name-prop/active-rfcs/0000-sfc-template-name-prop.md)


posva:
Follow the template properly, thank you.

PR #236, opened 2020-11-18
add context to shouldPrefetch and shouldPreload

gibkigonzo:


mayuxian:
Should we refer to webpack and integrate it into vite ?

PR #243, opened 2020-12-03
docs: change defineEmits to defineEmit

Scrum:


Scrum:
@yyx990803 ping

active-rfcs/0000-script-setup.md
```diff
@@ -75,7 +75,7 @@ Introduce a new script type in Single File Components: `<script setup>`, which e
     foo: String,
   })
   // expects emits options
-  const emit = defineEmits(['update', 'delete'])
+  const emit = defineEmit(['update', 'delete'])
```
posva, line 78: I think the code should be updated to `defineEmits` to match `defineProps`. I don't know if it would be a breaking change since this feature is experimental.

PR #254, opened 2021-01-15
Vue3 inject function should support self-inject

huanggm:
## Summary

Vue3 inject function should support self-inject.

## Links

- [Full Rendered Proposal](https://github.com/huanggm/rfcs/blob/master/active-rfcs/0038-inject-api-support-self-inject.md)

### DIFF
--- a/active-rfcs/0038-inject-api-support-self-inject.md
+++ b/active-rfcs/0038-inject-api-support-self-inject.md
@@ -1,0 +1,51 @@
+- Start Date: 2021-01-15
+- Target Major Version: 3.x
+- Reference Issues: https://github.com/vuejs/vue-next/pull/3022#issue-555432690
+- Implementation PR: https://github.com/vuejs/vue-next/pull/3022
+
+# Summary
+
+Vue3 inject function should support self-inject.
+
+# Basic example
+
+```js
+inject(key, defaultValue, treatDefaultAsFactory, selfInject = false);
+```
+
+Since parameter selfInject's default value is false. This change is an compatible update.
+
+# Motivation
+
+I just wrote a Vue3 plugin that depends on self-inject feature. The plugin could make Vue support `DI` like Angular.
+
+In angular, we can define component scope providers. It injects from self by default. And we can use `@Skip` to inject from parent.
+
+I wrote the plugin depends on Vue@3.0.1. It works perfectly. But when I update to Vue@3.0.2. It can not work.
+
+Without self-inject feature, every container component must be splited into two components.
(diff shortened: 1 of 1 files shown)


CyberAP:
The alternative here is to work with the order of your `provide` and `inject`.

```js
inject('foo') // ancestor-inject

provide('foo', 'bar')
```

```js
provide('foo', 'bar')

inject('foo') // self-inject
```

Or just use a reference?

```js
const provision = 'bar'

provide('foo', provision)

// work with the `provision` reference
```

That of course is not supported for Options API and I think it would benefit the most from this change (though it is not covered in the proposal).

huanggm:
@CyberAP Thank you for your reply.

In fact. With Vue@3.0.1 we can get different inject type by adjust the order of `provide` and `inject`.

Like what you said.

```js
inject('foo') // ancestor-inject
provide('foo', 'bar')

provide('foo', 'bar')
inject('foo') // self-inject
```

But we cannot get the same result from Vue@3.0.2. The real result is this.
```js
inject('foo') // ancestor-inject
provide('foo', 'bar')

provide('foo', 'bar')
inject('foo') // ancestor-inject
```

Of course use a reference can solve this problem. Though I think it is a little verbose.

yyx990803:
Thanks but this doesn't seem to warrant an additional flag given that within Composition API you should already have a reference to the value provided by the same component.

newcat:
> within Composition API you should already have a reference to the value provided by the same component.

That is true, however, there are use-cases where this is difficult to use / increases complexity quite a bit.

Suppose we have a parent component:
```js
setup() {
  const foo = ref("foo");
  provide("foo", foo);
}
```

and a use function:
```js
function useFooBar() {
  const foo = inject("foo");
  return { bar: computed(() => foo.value + "bar") };
}
```

In child components, I can now use `useFooBar` as expected. But it won't work in the parent component:
```js 
setup() {
  const foo = ref("foo");
  provide("foo", foo);
  // doesn't work without self-injecting
  const { bar } = useFooBar();
}
```

Of course I can provide foo as a parameter to the `useFooBar` function, but that would mean I need to write the `inject("foo")` statement into every child component - which is IMHO essentially the opposite of what the Composition was trying to do: decrease complexity by creating small reusable `use` functions.

iamandrewluca:
https://github.com/logaretm/vee-validate also has such a use case

```ts
const form = useForm()
const field = useField()
```

It uses a [custom inject](https://github.com/logaretm/vee-validate/blob/19286de0b71cb7f8eaad860d0d6a4ad5b6cbc645/packages/vee-validate/src/utils/common.ts#L136-L142) so useField can inject context of useForm from same component

PR #257, opened 2021-01-24
vue-router Data Store

hezedu:
## Summary
Maintain a store by history state key. Can be used to save something.

## Links

- [Full Rendered Proposal](https://hezedu.github.io/SomethingBoring/vue-router-positionstore-memory-leak/pr-demo.html#/)

- [Related issue](https://github.com/vuejs/vue-router/issues/3452)


posva:
Thank you but as I already said, this is not something the router needs, even less expose it as a public api

PR #271, opened 2021-03-02
[Vuex] Vuex 5

kiaking:
## Summary

Introducing a brand new Vuex for Vue 3. It's designed to improve various architecture and API of Vuex from what we've learned from past years. Here are the key differentiators from Vuex 3 & 4.

- 2 syntax support for the store creation, options api, and composition api.
- No mutations. Only state, getters, and actions.
- No nested modules. Only stores. Compose them instead.
- Complete TypeScript support.
- Transparent, automated code splitting.

## Links

- [Full Rendered Proposal](https://github.com/kiaking/rfcs/blob/vuex-5/active-rfcs/0000-vuex-5.md)
- [Discussion Thread](https://github.com/vuejs/rfcs/discussions/270)

---

### DIFF
--- a/active-rfcs/0000-vuex-5.md
+++ b/active-rfcs/0000-vuex-5.md
@@ -637,14 +637,14 @@ const useCounter = defineStore({
   getters: {
     // May require annotation as same as Vue Component.
     double(): number {
-      return this.state.count * 2
+      return this.count * 2
     }
   },
 
   actions: {
     // May require annotation as same as Vue Component.
     increment(): void {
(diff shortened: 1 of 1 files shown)


kiaking:
We're closing this RFC due to [Pinia](https://pinia.vuejs.org/) becoming the next official state management lib for Vue.

Pinia has almost the exact same or enhanced API as Vuex 5, described in this RFC. You could simply consider Pinia as Vuex 5 with a different name.

PR #284, opened 2021-03-16
Custom cache strategy and matching rules for KeepAlive

HcySunYang:
## Summary

- Support `matchBy`, allow to specify the match target, there are two available values:
  - `'name'`: use component name to match.
  - `'key'`: use the `key` of the component to match.

- Cache management, provide a new prop called `cache`, users can provide **custom caching strategy**.

- Add two new lifecycle hooks: `onBeforeActivate` and `onBeforeDeactivate`, and the corresponding options API `beforeActivate` and `beforeDeactivate`.

## Links

- [Full Rendered Proposal](https://github.com/HcySunYang/rfcs/blob/keepalive-cache-management-and-match-rules/active-rfcs/0000-keepalive-cache-management-and-match-rules.md)

- [Discussion Thread](https://github.com/vuejs/rfcs/discussions/283)

---


shuxiaoman:
I need this so much!!!!!

PR #293, opened 2021-03-30
`props` object should contain all `props` defined

pikax:
## Summary

Odd behaviour when using `toRefs(props)` if the prop has not been passed

```ts
defineComponent({
  props: {
    foo: String,
  },
  setup(props) {
    const { foo } = toRefs(props) // props is `{}` making `toRefs` To return `{}`

    // correct
    const foo = toRef(props, 'foo')
  },
})
```

## Links

- [Full Rendered Proposal](https://github.com/pikax/rfcs/blob/vue-next/props_torefs/active-rfcs/0000-define-unpassed-props.md)

- [Discussion Thread](https://github.com/vuejs/rfcs/discussions/292)

---


pikax:
fixed by https://github.com/vuejs/vue-next/commit/4fe4de0a49ffc2461b0394e74674af38ff5e2a20

PR #294, opened 2021-04-02
Proposal for dropping ie11 support in Vue 3

yyx990803:
## Summary

- Drop IE11 support plan for Vue 3.
- Focus on backport compatible features back to Vue 2.7.

## Links

- [Full Rendered Proposal](https://github.com/vuejs/rfcs/blob/ie11/active-rfcs/0000-vue3-ie11-support.md)
- [Discussion thead](https://github.com/vuejs/rfcs/discussions/296)
---


yyx990803:

PR #368, opened 2021-08-04
Ref sugar (take 2)

yyx990803:
## Summary

Introduce a compiler-based syntax sugar for using refs without `.value`:

```html
<script setup>
  // declaring a variable that compiles to a ref
  let count = $ref(1)

  console.log(count) // 1

  function inc() {
    // the variable can be used like a plain value
    count++
  }
</script>

<template>
  <button @click="inc">{{ count }}</button>
</template>
```

- [Discussion Thread](https://github.com/vuejs/rfcs/discussions/369)

### DIFF
--- a/active-rfcs/0000-ref-sugar.md
+++ b/active-rfcs/0000-ref-sugar.md
@@ -125,17 +125,17 @@ console.log(x, y)
 <summary>Compiled Output</summary>
 
 ```js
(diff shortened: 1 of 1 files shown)


yyx990803:
The original proposal has been updated and merged with props destructure. The discussion is still in https://github.com/vuejs/rfcs/discussions/369

Issue #392, opened 2021-09-24
How can I use `setup="props, { emit }"` to define `props`


zhenhappy:
![CleanShot 2021-09-24 at 11 50 04](https://user-images.githubusercontent.com/9161946/134615335-5013e134-12c7-47d7-88b9-c4f46bc9820e.png)

[...]

PR #393, opened 2021-09-24
chore: configure issue wizard

iamandrewluca:
Docs: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser

Example: https://github.com/code4moldova/catalog-profesii/issues/new/choose

I see that @posva struggles to close issues like this https://github.com/vuejs/rfcs/issues/392
Maybe this can help!


posva:
Ah yeah, we should use this and add a few links to encourage people to use the discussions rather than the issues

PR #432, opened 2022-02-25
`defineOptions` macro in `<script setup>`

sxzz:
## Summary

Introduce a macro in script setup, `defineOptions`, to use Options API in `script setup`, specifically to be able to set `name`, `inheritAttrs` and `render` in `script setup`.

## Links

- [Full Rendered Proposal](https://github.com/sxzz/vue-rfcs/blob/define-options/active-rfcs/0000-define-options.md)

- [Discussion Thread](https://github.com/vuejs/rfcs/discussions/430)

---


yyx990803:
Relevant: https://github.com/vuejs/core/pull/4997 is merged and component using `<script setup>` now automatically infers `name` option from filename.

PR #646, opened 2024-02-19
Add MIT license to repository

hallundbaek:
Add MIT license to repository allowing other projects to adapt and adopt the vuejs/rfcs process.


posva:
Thanks but this is not Software. Feel free to create your own RFC process that is similar to this one or [others](https://github.com/vuejs/rfcs#reviewing-rfcs)

PR #824, opened 2026-03-06
RFC: Inline Local Components

ubugeeei:
WIP

### DIFF
--- a/active-rfcs/0047-ilc.md
+++ b/active-rfcs/0047-ilc.md
@@ -1,0 +1,309 @@
+- Start Date: 2026-03-06
+- Target Major Version: 3.x
+- Reference Issues:
+- Implementation PR:
+
+# Summary
+
+ILC (Inline Local Components)
+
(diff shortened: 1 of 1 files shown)


naitokosuke:
I like this style! 

I found this tweet in my bookmark.
Maybe related or similar idea, I'll tag it.
https://x.com/sanxiaozhizi/status/1582185668335529985

https://github.com/vuejs/rfcs/discussions/517

jods4:
I really like the simple concept of having named `<template>` tags in SFC for lightweight local components, like the X post linked by naitokosuke shows.

Having full blown `<component>` that includes script/template/style (... and then why not other `<components>` recursively?) feels too heavyweight to me, at this point if I need more than a template I would create a new SFC file.

Two points that would need a dedicated syntax:

1. How could a lightweight local template declare props? Maybe we can have `props` attributes with a tiny DSL or just a TS type? 
`<template name='local-component' props="{ name: string, age: number }">`

2. It would be nice to be able to refer to local components not only from template but also from script.
A use case I would have is declaring a renderer for a data-grid columns definition object.

